PR #41, opened 2015-02-06
initial take on radosgw-agent task

alfredodeza:
This will need some iteration over (commented out lines courtesy of Jenkins copy/pasted shell forms)

Once this hits master it will not overwrite the current radosgw-agent job (this uses a different name). Only when this is finally fully functional we can get it do be the defacto one.

### DIFF
--- a/radosgw-agent/build/build
+++ b/radosgw-agent/build/build
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -ex
+
+date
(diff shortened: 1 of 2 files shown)


alfredodeza:
test this please

ktdreyer:
LGTM. My preference would be to get this in the tree as-is, and then iterate in future PRs

PR #43, opened 2015-03-30
build_rpms.sh: terminate on unpackaged files

ktdreyer:
As Ceph changes and new files get installed by default, we need to ensure that the commits which create these files also update the packaging files lists as well.

One way to do this is to tell Gitbuilder to stop skipping unpackaged files in the RPMs. This will cause Gitbuilder builds to indicate a failure when new files get added without packaging updates.


ktdreyer:
I've submitted a similar PR to autobuild-ceph, at https://github.com/ceph/autobuild-ceph/pull/40

PR #44, opened 2015-04-13
import ceph task

ktdreyer:
This imports the current settings for "ceph" as YAML.

This is an import of the job previously defined in Jenkins' own data store (ie `/home2/jenkins/jobs/ceph/config.xml`)

### DIFF
--- a/ceph-setup/config/definitions/ceph-setup.yml
+++ b/ceph-setup/config/definitions/ceph-setup.yml
@@ -28,7 +28,6 @@
             - $BRANCH
 
     builders:
-      - shell: "pwd"
       - shell:
           !include-raw ../../build/build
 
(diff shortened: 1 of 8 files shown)


ceph-setup/config/definitions/ceph-setup.yml
```diff
@@ -0,0 +1,39 @@
...
+          description: "The git branch (or tag) to build"
+
+    scm:
+      - git:
+          url: git://jenkins.front.sepia.ceph.com/ceph/ceph.git
+          branches:
+            - $BRANCH
+
+    builders:
+      - shell: "pwd"
```
alfredodeza: this is probably debug right? can we remove it?

ktdreyer: Agreed. I pushed a new commit to remove it.

PR #45, opened 2015-04-13
ceph-build-pull-requests: quiet output

ktdreyer:
Running the pull requests script with `-x` is very noisy. The script already prints a lot of information with "echo", so just silence the debugging outupt.

### DIFF
--- a/ceph-build-pull-requests/build/build
+++ b/ceph-build-pull-requests/build/build
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ex
+set -e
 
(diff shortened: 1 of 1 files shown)


alfredodeza:
could we make it configurable with a checkbox? When the checkbox is on all scripts get `-x` 

But I guess that would require changes to all the scripts huh

ktdreyer:
We could certainly adjust it to do that, but since this is the "build all pull requests" script, it's unlikely that our team will enter parameters, right?

PR #58, opened 2015-05-28
Nix dcompute

alfredodeza:
These are now mapped to a mira host.


trhoden:
I like where this is going...  :)

PR #59, opened 2015-06-01
port ceph-deploy scripts to jjb build script

alfredodeza:
Had to concatenate the three scripts together because I am not sure we can do an 'if/else' depending on rpm/deb to execute them, as it was the case before.

From here we can just tinker to get it right. This as it is should still not work with the new hosts added.

### DIFF
--- a/ceph-deploy/build/build
+++ b/ceph-deploy/build/build
@@ -1,0 +1,181 @@
+#!/bin/sh
+
+# This is the script that runs inside Jenkins.
+# http://jenkins.ceph.com/job/ceph-deploy/
+
+set -x
+set -e
+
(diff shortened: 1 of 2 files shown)


ktdreyer:
To clarify what we mean about "still not work with the new hosts added", so we know what that means when we look back a year from now... Our Jenkins slaves used to run in DreamCompute, but we've had some network problems there with our OpenVPN connection, so we're migrating our Jenkins slaves to a set of "new" hosts that are local to the Sepia lab in Irvine.

The changes in this PR itself looks good to me. Merging

alfredodeza:
I should've probably added that to the bottom of the commit :/ sorry about that

ktdreyer:
No prob, that's why I put it here :)

PR #60, opened 2015-06-02
fix/improve ceph-deploy build script

alfredodeza:
It will now use `$DIST` although in a not very flexible way. 

It fixes a EOF error as well and removes some hard coded variables.


dmick:
Hurray

PR #63, opened 2015-06-03
Debian setup

alfredodeza:
Removes the need for `/srv/ceph-build/` and its scripts, porting them to the ceph-build/build/ structure, ensuring that required things for pbuilders exist.

Initial structure for `build_rpm` created as well. 

Added comments and notes for fixing/investigating items at a later point. At the moment we can't build anything Ceph so this wouldn't really break anything and allows us to get in the right direction.


ktdreyer:
LGTM, merging

PR #65, opened 2015-06-03
remove raring from pbuilder setup

alfredodeza:
We aren't even building for it and the mirror doesn't have anything to support it either:

    W: Failed to fetch http://us.archive.ubuntu.com/ubuntu/dists/raring/main/binary-amd64/Packages  404  Not Found [IP: 91.189.91.13 80]


ktdreyer:
LGTM, merging

PR #66, opened 2015-06-03
use $DIST properly

alfredodeza:
This fixes a few issues:

* a variable that was not set properly because of backticks
* use of $DIST, replacing the hard-coded `$dists`
* removes hardcoded DEBEMAIL


ktdreyer:
Looks fine to me. (gitbuilder-cdep-deb-trusty-amd64-basic is offline at the moment so Jenkins can't test this PR. It wouldn't test this change anyways since it's not touching the YAML.)

alfredodeza:
I made the changes manually to the job and it is building

PR #67, opened 2015-06-03
use the 'trusty' label

alfredodeza:
Fixes the problem with all the other jobs that used a rather hardcoded name.

Using 'trusty' is more flexible because we can have more 'trusty' nodes around


ktdreyer:
LGTM, merging

PR #68, opened 2015-06-04
make setup_pbuilder use DIST

alfredodeza:
So that it avoids building every distro on every server on
every build


ktdreyer:
looks good to me, merging

PR #71, opened 2015-06-04
rpm script runs only in rpm systems

alfredodeza:


ktdreyer:
If we care about SUSE, they won't have this file, right?

alfredodeza:
hrmn

PR #80, opened 2015-06-05
ceph-build: better SUSE detection

ktdreyer:
As was done in 18890d5a0351ffe4fc495e0d47e3a86e0a9cb7ad for radosgw-agent, update ceph-build's SUSE detection so we check for the "zypper" file directly on-disk.

### DIFF
--- a/ceph-build/build/build_rpm
+++ b/ceph-build/build/build_rpm
@@ -1,16 +1,7 @@
 #!/bin/bash
 set -ex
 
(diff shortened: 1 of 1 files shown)


alfredodeza:
Is it possible that `/usr/bin/zypper` is a symlink or that it lives in some other place? Those were my concerns at some point.

ktdreyer:
I don't think so - at least, I've never seen that before

PR #81, opened 2015-06-08
cd to $WORKSPACE

alfredodeza:
Otherwise we will fail to be in the right directory to copy the
artifacts


alfredodeza:
I did the change on Jenkins already and got it working:

http://jenkins.ceph.com/job/ceph-build/Arch=x86_64,Distro=rhel7/359/console

PR #83, opened 2015-06-08
Remove 'noreplace' option on installed config file, /etc/yum.repos.d/ceph.repo...

jcalcote:
...in ceph-release package so the package version can replace the user-created one.

When the 'noreplace' option is used on a %config file in rpm.spec, the system will not replace an existing config file which has been modified since the last package was laid down, or which was created by a user if no previous package was installed. Instead it lays down the new config file renamed to .rpmnew and leaves the existing file alone. I think the intent of the author was likely that the package config file replace the one created by the user, as it contains more entries that are required by ceph-deploy, and the quick start preflight guide only indicates a single repo - the ceph-noarch repo.

### DIFF
--- a/ceph-release-rpm/build/build
+++ b/ceph-release-rpm/build/build
@@ -103,9 +103,9 @@ install -pm 644 %{SOURCE0} \
 %defattr(-,root,root,-)
 #%doc GPL
 %if 0%{defined suse_version}
-%config(noreplace) /etc/zypp/repos.d/*
+%config /etc/zypp/repos.d/*
 %else
-%config(noreplace) /etc/yum.repos.d/*
+%config /etc/yum.repos.d/*
 %endif
 #/etc/pki/rpm-gpg/*
 


ktdreyer:
Hi @jcalcote , it's pretty typical to have `.repo` files as `%noreplace` . Here are some examples where the `.repo` files are specified with `%noreplace`:

http://pkgs.fedoraproject.org/cgit/fedora-repos.git/tree/fedora-repos.spec
http://pkgs.fedoraproject.org/cgit/epel-release.git/tree/epel-release.spec?h=epel7
https://github.com/puppetlabs/puppetlabs-release/blob/master/templates/redhat/puppetlabs-release.spec.erb

Can you elaborate on the problem that you're seeing with ceph-deploy, with the exact steps you followed? Maybe there are some other changes we can make to address the issue you describe.

(The separate "noarch" repo convention is annoying and something I hope to drop one day. We should really be cross-shipping all noarch packages into each architecture's directories.)

jcalcote:
Sure, the basic problem is this: The doc (quick start preflight) indicates you should create this file (/etc/yum.repos.d/ceph.repo) with a single entry [ceph-noarch]. However, to properly deploy ceph, multiple entries are required, ceph-noarch, ceph-source, and ceph. The ceph-release package actually  installs a version of ceph.repo with the proper entries in it. If you use 'noreplace' in this case, the user's temporary version will not be replaced by the package version and ceph-deploy install fails because critical repo descriptions are missing from this file.

I believe the original intent of the quick start preflight doc was to hack a short version of ceph.repo into place that would allow ceph-deploy to get far enough to install the proper one, overwriting the original one.

ktdreyer:
It sounds to me that the quickstart guide should say "yum install http://url/to/ceph-release.rpm" instead of "here's how to set up a .repo file by hand". What do you think?

jcalcote:
I'm kinda new to ceph, actually - that's why I'm going through the
preflight and quick start. I'll defer to you on how to actually fix the
problem, if you don't mind. I was just trying to provide a solution for
something I ran into that was reasonable correct without trying to modify
the framework (as that usually take a bit more insight than your average
noob has). :)

ktdreyer:
Oh, definitely :) CC'ing @trhoden here, as I think he was looking into this area of the installation process recently.

alfredodeza:
The scripts and configuration files in this repo are not meant to be used outside of the CI/build environment. 

The quickstart guide mentions the ceph.repo file *to install ceph-deploy* because that package exists in that repository. Ceph repositories and repo files are handled by ceph-deploy afterwards (if following that same guide).

http://docs.ceph.com/docs/master/start/quick-start-preflight/#red-hat-package-manager-rpm

PR #85, opened 2015-06-22
Ports the teuthology-pull-requests job to JJB

andrewschoen:
Let me know if we're missing anybody in the white-list or admin-list.  I'm also unsure what the process is to deploy this after merge, so I'll need some guidance there.

### DIFF
--- a/teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
+++ b/teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
@@ -1,0 +1,72 @@
+- job:
+    name: teuthology-pull-requests 
+    node: trusty
+    project-type: freestyle
+    defaults: global
+    disabled: false
+    display-name: 'Teuthology: Pull Requests'
+    concurrent: false
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - github:
+          url: https://github.com/ceph/teuthology/
+    logrotate:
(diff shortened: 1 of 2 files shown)


teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
```diff
@@ -0,0 +1,72 @@
...
+            - jcsp
+            - gregsfortytwo
+            - GregMeno
+            - dzafman
+            - dillaman
+            - dachary
+            - liewegas
+            - idryomov
+            - vasukulkarni
+          trigger-phrase: ''
```
andrewschoen: Do I need anything here?  Is this where the phrase like "ok to test" would go?

ktdreyer: Yeah, that seems a little odd to leave that blank. Is it blank in the job's `config.xml`?

alfredodeza: I don't think this is needed, it is configured globally in the plugin

andrewschoen: Gotcha, I noticed it was blank in the other *-pull-requests jobs as well.

ktdreyer:
>I'm also unsure what the process is to deploy this after merge, so I'll need some guidance there.

The process is just "merge the PR". Assuming the job has the same name as the previous job, JJB will overwrite the old one.

PR #87, opened 2015-06-24
Requirement support

alfredodeza:
For ceph-deploy-pull-requests so that we have a way to consistently declare what a job needs


ceph-deploy-pull-requests/setup/playbooks/tasks/ubuntu.yml
```diff
@@ -0,0 +1,18 @@
+---
+  - name: add deadsnakes ppa
+    action: apt_repository repo=ppa:fkrull/deadsnakes state=present
+
+  - name: "udpate apt repo"
```
trhoden: s/udpate/update

trhoden:
This seems like a very sane approach to me.

ceph-deploy-pull-requests/build/setup
```diff
@@ -0,0 +1,35 @@
...
+# Define and ensure a GIT Ansible cache
+REPO_CACHE="$HOME/.cache/git/ceph-build"
+mkdir -p $REPO_CACHE
+
+# Clone the ceph-build repo or pull if it exists
+if [ -d "$REPO_CACHE/.git" ]; then
+    cd "$REPO_CACHE" && git pull
+    cd -
+else
+    git clone https://github.com/ceph/ceph-build.git "$REPO_CACHE"
```
ktdreyer: using git operations like `clone` here seems a bit weird, but unfortunately I can't think of a more elegant solution off the top of my head.

ktdreyer: I think Jenkins' git plugin supports multiple remotes - could we add ceph-build.git as a second remote?

alfredodeza: good catch

alfredodeza: wait, I don't think that using ceph-build as a second remote would make sense for setting up. This should only run when the job runs, so it properly sets it up. If you set up ceph-build as a second remote this setup will run but it may run on any host not necessarily where/when we need it

ktdreyer: From what I understand, the Github Pull Requests plugin can be set up to only watch ceph-deploy.git for changes, and then Jenkins will clone both the first and second repositories on the slave.

Let's take this PR as-is for now and we can work to optimize this later.

PR #95, opened 2015-07-06
rm top-level "jenkins" directory

ktdreyer:
Prior to the introduction of Jenkins Job Builder for configuring our jobs, we stored each of Ceph's build instructions in a "jenkins" directory. This was a first crude step towards versioning our build instructions.

These "jenkins/*" scripts are no longer necessary because the JJB job definitions contain the same steps (and these scripts are no longer used). Clean them up.


ktdreyer:
:shower:

PR #98, opened 2015-07-08
ansible: rm green balls plugin

ktdreyer:
This plugin is not color-blind friendly. It is possible to turn it off on a per-user setting, but that requires logging into Jenkins. Let's just turn it off entirely so we're fully accessible for people who
aren't logged in.


alfredodeza:
I personally apologize for this, I had no idea that this was a poor choice for color-blind peeps and had always thought it was better than just blue.

PR #99, opened 2015-07-11
try to get multiple repos for ceph-deploy-pull-requests

alfredodeza:
A first initial pass to tackle this one, if all goes well we can make this a common throughout so we can call playbooks and other bits and pieces we might need from ceph-build on certain jobs.


alfredodeza:
6zpatsz

ktdreyer:
:smile:

PR #100, opened 2015-07-14
fix some issues with `$WORKSPACE` being overwritten

alfredodeza:
This PR uses `basedir` to allow cloning to a sub-directory relative to `$WORKSPACE`

### DIFF
--- a/ceph-deploy-pull-requests/build/setup
+++ b/ceph-deploy-pull-requests/build/setup
@@ -10,17 +10,8 @@ virtualenv venv
 PIP_SDIST_INDEX="$HOME/.cache/pip"
 mkdir -p $PIP_SDIST_INDEX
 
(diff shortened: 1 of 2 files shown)


trhoden:
Looks good!  Thanks @alfredodeza

ktdreyer:
There's valid reasons for wiping the workspace before each build - sorry I'm missing the bigger context of what problem we were trying to solve, here?

alfredodeza:
@ktdreyer I think we need to start cutting tickets for these things. Basically, Jenkins would clone repo1 and then overwrite everything in $WORKSPACE because it would clone repo2 in the same path.

Good catch in that I thought it was a 'do not wipe the workspace' thing, but it was really that we needed to point to a sub directory to avoid $WORKSPACE getting overwritten

PR #108, opened 2015-07-24
infer the BPTAG from the value of DIST

alfredodeza:


ktdreyer:
I'm fine with this as-is. Eventually we could switch to using modern stuff like case statements (and deleting EOL distros :)

alfredodeza:
I am not sure why this was never merged

ktdreyer:
too much goofing off around here :smile:  :tada:

PR #109, opened 2015-07-28
check if we have valid tars for pbuilder

alfredodeza:


ktdreyer:
works for me

You resisted the urge to rewrite the whole script in python! :D

PR #113, opened 2015-08-27
import ice-setup

ktdreyer:
Migrate the ice-setup job from our internal jenkins-calamari server to jenkins.ceph.com.

### DIFF
--- a/ice-setup/config/definitions/ice-setup.yml
+++ b/ice-setup/config/definitions/ice-setup.yml
@@ -1,0 +1,55 @@
+- job:
+    axes:
+    - axis:
+        name: Arch
+        type: label-expression
+        values:
+        - x86_64
+    - axis:
+        name: Dist
+        type: label-expression
+        values:
+        - trusty
+        - precise
+        - centos
+    block-downstream: false
+    block-upstream: false
+    builders:
+    - shell:
+        !include-raw ../../build/build
+    concurrent: false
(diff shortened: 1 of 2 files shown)


ice-setup/config/definitions/ice-setup.yml
```diff
@@ -0,0 +1,55 @@
...
+    name: ice-setup
+    parameters:
+    - string:
+        default: master
+        description: ''
+        name: BRANCH
+    project-type: matrix
+    properties:
+    - github:
+        url: http://github.com/ceph/ice-tools/
```
alfredodeza: is this the right URL?

ice-setup/config/definitions/ice-setup.yml
```diff
@@ -0,0 +1,55 @@
...
+        only-if-success: false
+    - description-setter:
+        regexp: '^BRANCH: (\S+)'
+        regexp-for-failed: null
+        set-for-matrix: false
+    scm:
+    - git:
+        basedir: ice-setup
+        branches: $BRANCH
+        url: https://github.com/ceph/ice-setup
```
alfredodeza: oh this one is the right url, I wonder whats up with the other one?

ice-setup/config/definitions/ice-setup.yml
ktdreyer: great catch. it was wrong in the original job on jenkins-calamari. I'll fix it here.

PR #117, opened 2015-08-31
Add mariner-installer job

trhoden:
Start building upstream packages for
https://github.com/ceph/mariner-installer

### DIFF
--- a/mariner-installer/build/build
+++ b/mariner-installer/build/build
@@ -1,0 +1,43 @@
+#!/bin/bash -ex
+
+echo "BRANCH: $BRANCH"
+cd $WORKSPACE/mariner-installer
+
+# FIXME A very naive way to just list the RPM $DIST that we currently support.
+# We should be a bit more lenient to allow any rhel/centos/sles/suse
+rpm_dists="rhel centos6.5 centos7 rhel6.5 rhel7 centos sles11sp2 opensuse12.2"
+deb_dists="precise wheezy squeeze trusty"
+
+# A helper to match an item in a list of items, like python's `if item in list`
+listcontains() {
+  for word in $2; do
+    [[ $word = $1 ]] && return 0
+  done
+  return 1
+}
+
+if listcontains $DIST "$deb_dists" ; then
+    DEBIAN=1
+    dpkg-buildpackage -us -uc
+elif listcontains $DIST "$rpm_dists" ; then
+    DEBIAN=0
+    tito build --rpm
+else
+    echo "Can't determine build host type"
+    exit 1
+fi
+
+if [ -n "$BRANCH" ]; then
+    # A user explicitly told Jenkins to build a branch.
+    # (ie, this is not an automatic GitHub pull request build.)
+    # Move the artifacts into position for archiving.
+    cd $WORKSPACE
+    rm -rf $BRANCH
+    mkdir $BRANCH
+
+    if [ $DEBIAN = 1 ] ; then
+        mv *.deb *.dsc *.changes *.tar.gz $BRANCH
+    else
+        find rpmbuild -name '*.rpm' -o -name '*.srpm' | xargs -i mv {} $BRANCH
+    fi
+fi
--- a/mariner-installer/config/definitions/mariner-installer.yml
+++ b/mariner-installer/config/definitions/mariner-installer.yml
@@ -1,0 +1,53 @@
+- job:
+    name: mariner-installer
+    description: Build and package mariner-installer
+    project-type: matrix
+    axes:
+    - axis:
+        name: Arch
+        type: label-expression
+        values:
+        - x86_64
+    - axis:
+        name: Dist
+        type: label-expression
+        values:
+        - trusty
+    block-downstream: false
+    block-upstream: false
+
+    properties:
+    - github:
+        url: https://github.com/ceph/mariner-installer/
+
+    scm:
+    - git:
+        skip-tag: true
+        basedir: mariner-installer
+        url: https://github.com/ceph/mariner-installer
+        branches:
+         - $BRANCH
+
+    execution-strategy:
+      run-sequentially: false
+
+    builders:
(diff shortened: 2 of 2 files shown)


mariner-installer/build/build
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -ex
+
+echo "BRANCH: $BRANCH"
+cd $WORKSPACE/mariner-installer
+
+. /etc/os-release
```
alfredodeza: do we do this anywhere else? (not sure why are we sourcing here).

Jenkins should slap a bunch of environment variables for you, so not sure I understand why the need to source this. Are the env vars supplied not sufficient?

mariner-installer/build/build
```diff
@@ -0,0 +1,33 @@
...
+
+if [ -n "$BRANCH" ]; then
+    # A user explicitly told Jenkins to build a branch.
+    # (ie, this is not an automatic GitHub pull request build.)
+    # Move the artifacts into position for archiving.
+    cd $WORKSPACE
+    rm -rf $BRANCH
+    mkdir $BRANCH
+
+    if [ $DEBIAN = 1 ] ; then
```
alfredodeza: so if this is Debian Jessie, we wouldn't do this? (otherwise why call this variable `DEBIAN`)

mariner-installer/build/build
trhoden: No clue.  Straight up copy from what Ken did with takora.  He add the `os-release` source in https://github.com/ceph/ceph-build/commit/3bd7177ad9e6568c43a67eee023aa5d34ee5c6a0

trhoden: I should probably actually tag @ktdreyer so he sees this.  :)

mariner-installer/build/build
trhoden: TL;DR - I don't think these packages are destined for ceph.com

I think this comes from the fact that we are only building this for Trusty?  That's the only dist in the job config.

I think this job config is mostly for a smoke test.  A separate job will come in for downstream packaging (for RHCS on Ubuntu).

For community packages, I intend to host Deb packages in a PPA, just like the RPMs are in a COPR repo.  

That's my story.  Hopefully @ktdreyer agrees.

mariner-installer/build/build
alfredodeza: HALO THAR @ktdreyer 
http://i.imgur.com/jLBIIZg.gif

ktdreyer: The reason for sourcing `/etc/os-release` is so that `"$ID" == 'ubuntu'`

We've got a couple different OS detection mechanisms lying around in this tree, and we should probably standardize at some point

ktdreyer:
an issue with doing "official" builds in a PPA or Copr is that only you can access the infrastructure to do that :/ Unless we share that, I guess?

trhoden:
@ktdreyer completely agreed.  I do not want to host official packages for anything -- those of course belong on ceph.com.  I just am not sure *when* we would intend to build/host those.  We don't build/host calamari, for example.

To your specific point -- the Copr repo is easy to share with you or anyone with a FAS account.  Not sure about PPA/launchpad yet...

Was it a fair comment to say that we aren't worried about Debian Jessie because we are only building for Trusty?

FWIW, mariner-installer is only intended to be "supported" on CentOS 7, RHEL 7, and Ubuntu 14.04 for now (Fedora will probably work, though).

mariner-installer/build/build
alfredodeza: My opinion on why this particular way of detecting the host info shouldn't be used:

    [vagrant@localhost ~]$ cat /etc/os-release
    cat: /etc/os-release: No such file or directory
    [vagrant@localhost ~]$ cat /etc/redhat-release
    CentOS release 6.5 (Final)

I understand this is for Ubuntu/Debian only, but there are other ways that are cross-platform and would help normalize what we use across ceph-build already

trhoden: Okay, so let's figure out what we want to do, and I'll submit a separate PR to fixup takora to do that, then apply that same change here.

`ceph-deploy` checks if `$DIST` is in a static list: https://github.com/ceph/ceph-build/blob/master/ceph-deploy/build/build#L35-L42

`ceph-build` simply calls two scripts, and has the irrelevant script exit immediately:
https://github.com/ceph/ceph-build/blob/master/ceph-build/build/build_deb#L5-L7
https://github.com/ceph/ceph-build/blob/master/ceph-build/build/build_rpm#L4-L6

The "original" way before 3bd7177 was as it is in `ice-setup`, checking $DIST as well: https://github.com/ceph/ceph-build/blob/master/ice-setup/build/build#L8-L12

`radosgw-agent` uses a different permutation of all this: https://github.com/ceph/ceph-build/blob/master/radosgw-agent/build/build#L25

Voting is now open.  :smiley:

mariner-installer/build/build
ktdreyer: @alfredodeza your point is valid about the "`DEBIAN`" variable only being set on Ubuntu. If we really want to clean that up, I propose that we first do it to the takora job, so that this one matches.

mariner-installer/build/build
ktdreyer: My bad! I thought that CentOS 6 had `/etc/os-release`, and I see now that [it does not] (https://bugs.centos.org/view.php?id=8894). Let's just go with ceph-deploy's `listcontains()`

trhoden:
@ktdreyer @alfredodeza okay, take a look.  Made the logic more like ceph-deploy's.  If it looks good, I can add the same to takora.

mariner-installer/build/build
trhoden: @alfredodeza This should be taken care of now -- `DEBIAN` would be set to 1 for all debain distros, not just Ubuntu.

alfredodeza:
this looks good to me

PR #119, opened 2015-09-01
Jessie support

alfredodeza:
Adds Jessie support on the build scripts for Ceph, radosgw-agent, and ceph-deploy

Removes older distros that we do not build for anymore


ktdreyer:
so much clean up ... thank you

PR #121, opened 2015-09-01
Use $DIST for determining distro

trhoden:
Makes takora job similar to ceph-deploy and mariner-installer


ktdreyer:
Thanks!

PR #123, opened 2015-09-02
normalizing to `DIST`

alfredodeza:
This allows us to finally get rid of the dependency of setting `DIST` and `BPTAG` on actual slaves, giving us flexibility to just consume whatever is set by Jenkins.

Since the configuration matrix will use `DIST` and not `Dist` it means that it will be available as an environment variable throughout.

It does some other minor cleanups as well.

### DIFF
--- a/ceph-package/config/definitions/ceph-package.yml
+++ b/ceph-package/config/definitions/ceph-package.yml
@@ -23,7 +23,6 @@
           type: label-expression
           name: DIST
           values:
-            - squeeze
             - wheezy
             - precise
             - trusty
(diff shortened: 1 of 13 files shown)


ktdreyer:
LGTM! @dmick, care to review?

ceph-package/config/definitions/ceph-package.yml
```diff
@@ -13,24 +13,20 @@
...
       - axis:
           type: label-expression
-          name: Distro
+          name: DIST
           values:
-            - squeeze-pbuild
-            - wheezy-pbuild
-            - precise-pbuild
-            - trusty-pbuild
+            - squeeze
```
trhoden: do we need to keep squeeze here even though it is gone in ceph-build?

trhoden:
One nit/question, otherwise LGTM.

PR #125, opened 2015-09-08
rhcs-installer needs ceph-puppet-modules

trhoden:
not takora


ktdreyer:
For a moment I thought that we could make the ceph-puppet-modules package Provide "takora"... but I think that would lead to further subtle "gotcha" problems down the road. I like the explicitness of this approach, here.

trhoden:
Yeah...  I think we both understand that approach better on the RPM side, as well.  maybe we can revisit in the future.

PR #128, opened 2015-09-22
[DNM] use new GPG key for releases

ktdreyer:
See
http://ceph.com/releases/important-security-notice-regarding-signing-key-and-binary-downloads-of-ceph/

17ED316D is the old key

9DCEEEAD is the new key

(This key change probably needs more than simply swapping the key-id, but this PR is a start)

### DIFF
--- a/ceph-build/build/build_deb
+++ b/ceph-build/build/build_deb
@@ -29,7 +29,7 @@ get_bptag() {
 BPTAG=`get_bptag $DIST`
 
 export GNUPGHOME=/home/jenkins-build/build/gnupg.ceph-release/
-export KEYID=17ED316D
+export KEYID=9DCEEEAD
 HOST=$(hostname --short)
 echo "Building on $(hostname) Date: $(date)"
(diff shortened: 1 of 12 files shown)


alfredodeza:
I was going to merge (this looks OK) but you have a DNM on the subject. Let me know when this is good to go

ktdreyer:
The reason I put "[DNM]" is that I don't see the new key in the usual location on jenkins.front... so if we switch the key ID right now, GPG isn't going to know what to do

ktdreyer:
Maybe it would be more secure to simply remove the GPG stuff altogether from all the Jenkins jobs, and run that by hand instead?

alfredodeza:
I AGREE

dmick:
Pretty sure that's not the new key.  That looks like the old 'testing' key

ktdreyer:
You're right Dan, my bad! 460F3994 is the new key.

I think we're agreed that I should close this and open a new PR to remove all the GPG signing bits entirely from JJB.

PR #130, opened 2015-09-22
remove -e flag because that 'execs' an ssh command and there is none

alfredodeza:
Manually changed this in Jenkins and the job was a success.

http://jenkins.ceph.com/job/teuthology-docs/608/


andrewschoen:
Awesome, thanks!

PR #134, opened 2015-09-23
Ceph gpg

alfredodeza:
This PR removes GPG signing from everything. We no longer sign anything when building in Jenkins.

We will sign *once* as a last (manual) step before publishing repos.

Added comments where pertinent 

Removed completely the kmod directory - not used anymore


ktdreyer:
yessssss

trhoden:
Woohoo!  :beers:

PR #140, opened 2015-10-20
duplicate the ceph-build job as ceph-build-next

andrewschoen:
We're going to use ceph-build-next as the new build job that utilizes chacra.ceph.com.

This doesn't do much yet except setup the basic structure for the new job.


ceph-build-next/build/setup
```diff
@@ -0,0 +1,85 @@
...
+
+# Create the virtualenv
+virtualenv venv
+. venv/bin/activate
+
+# Define and ensure the PIP cache
+PIP_SDIST_INDEX="$HOME/.cache/pip"
+mkdir -p $PIP_SDIST_INDEX
+
+# The setup for this job ensures that a copy of ceph-build will be available
```
alfredodeza: this is not true I think, can we remove?

PR #144, opened 2015-10-22
ansible: use Ubuntu UID for nginx.conf

ktdreyer:
On CentOS, the nginx package from EPEL defaults to using the "nginx" UID.

On Ubuntu, the nginx package defaults to using the "www-data" UID.

Ideally we'd conditionalize this with Ansible variables, but I'll just hack it in for now.


ktdreyer:
vroom :speedboat:

PR #146, opened 2015-10-22
remove rhel and dotted releases from the build jobs

alfredodeza:


ktdreyer:
looks like there's some more "`rhel`" references to remove in "`rpm_dists`" but this definitely solves the main issue today! thanks

PR #149, opened 2015-10-23
programatically use a jenkins job config ini file

alfredodeza:
That sets the password directly from the Jenkins master. This allows us to not need any config file in any slave


ktdreyer:
LGTM

PR #154, opened 2015-10-26
programatically make the .chacractl file

andrewschoen:
This uses variables created by the masks-passwords jenkins plugin.


alfredodeza:
are these already set in jenkins?

andrewschoen:
@alfredodeza nope, and I'm not even sure how to do that. Does this look like the correct way to do this though?

alfredodeza:
yes this would be right, is there a need for the user and url to be masked? the key needs to be for sure

ktdreyer:
To do it:

1. visit https://jenkins.ceph.com/configure
2. under "Global Properties", set the `CHACRACTL_USER` environment variable
3. scroll down and under "Global Passwords", set the `CHACRACTL_KEY` value

andrewschoen:
@alfredodeza I don't plan to mask them, but doesn't hurt to make them configurable.

PR #155, opened 2015-10-26
rm ceph-kmod-rpm job

ktdreyer:
RHEL 7.1 includes kRBD support in the main kernel package, and we no longer maintain any out-of-tree modules for RHEL.

This Jenkins job is now obsolete. Remove it.

CephFS users should not use the out-of-tree cephfs kmod packages. Use Linus' mainline kernel packages from elrepo instead.


ktdreyer:
thanks for the merge

There is nothing in JJB that deletes jobs once they're removed from GitHub, so I've removed the job at https://jenkins.ceph.com/job/ceph-kmod-rpm/ by hand

PR #157, opened 2015-10-26
add the ability to do ceph test builds

andrewschoen:
If TEST_BUILD is checked, it'll push binaries to chacra.ceph.com under the ``test`` ref. This way we can make sure to separate testing binaries from release ones.


alfredodeza:
we tend to call builds like this from irc passing the options on the message, do you see a problem with using `TEST` instead of `TEST_BUILD` ? The call to kraken would look like:

    !ci build ceph BRANCH=v0.94.1 TEST=1

andrewschoen:
@alfredodeza not at all, I'll make that change

PR #158, opened 2015-10-26
use publish-over-ssh plugin for docs jobs

ktdreyer:
We will move docs.ceph.com to DreamCompute and publish docs directly to it.

### DIFF
--- a/teuthology-docs/config/definitions/teuthology-docs.yml
+++ b/teuthology-docs/config/definitions/teuthology-docs.yml
@@ -34,3 +34,11 @@
     builders:
       - shell:
           !include-raw ../../build/build
+
+    # publish docs to docs.ceph.com/teuthology/docs
+    publishers:
+      - ssh:
+          site: 'docs.ceph.com'
+          target: 'teuthology/docs/'
+          source: '.tox/docs/tmp/html/**'
+          remove-prefix: '.tox/docs/tmp/html/'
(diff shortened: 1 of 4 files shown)


teuthology-docs/config/definitions/teuthology-docs.yml
```diff
@@ -34,3 +34,12 @@
...
           !include-raw ../../build/build
+
+    # publish docs to docs.ceph.com/teuthology/docs
+    publishers:
+      - ssh:
+          site: 'docs.ceph.com'
+          target: 'teuthology/docs/'
+          source: '.tox/docs/tmp/html/**'
+          remove-prefix: '.tox/docs/tmp/html/'
+          clean-remote: true
```
alfredodeza: I would rather not do this, because if the transfer gets cut in the middle for whatever reason it will leave us with a broken docs site

ktdreyer: Yeah, I'd added `clean-remote` to match the `--delete` command that we were passing to rsync, but it really does not behave the same way.

I'll just remove `clean-remote` for now.

A more robust solution would be to copy all the files to an individual time-stamped directory, and then exec a command that quickly moves the old directory out of the way and moves the new one into the old one's place. Or maybe use https://wiki.jenkins-ci.org/display/JENKINS/SSH+Agent+Plugin

ktdreyer:
I tested https://wiki.jenkins-ci.org/display/JENKINS/SSH+Agent+Plugin with rsync , and it's two orders of magnitude faster when publishing the ceph.com docs. Let's go with that instead of the publish-over-ssh plugin.

PR #159, opened 2015-10-27
allow some jobs to use globally configured variables

andrewschoen:
Thes variables are defined on the jenkins master using the web UI.

http://jenkins.ceph.com/configure


andrewschoen:
I couldn't find if there was a more "global" place we could put this so that all jobs would get these variables by default. That's probably the better solution if it exists.

ktdreyer:
LGTM

It is possible to use templates with JJB's YAML, but we're not doing that at the moment. I think it's ok to just copy & paste for now.

PR #161, opened 2015-10-27
check chacra to see if a binary exists before rebuilding it

andrewschoen:


ceph-build-next/build/build_deb
```diff
@@ -6,6 +6,19 @@ if test -f /etc/redhat-release ; then
...
 
+[ "$TEST" = true ] && chacra_ref="$vers" || chacra_ref="test"
+vers=`cat ./dist/version`
+distro=`python -c "exec 'import platform; print platform.linux_distribution()[0].lower()'"`
+chacra_endpoint="ceph/${chacra_ref}/${distro}/{$dist}/${ARCH}"
+
+binary_exists=`chacractl binary exists ${chacra_endpoint}`
+
+# if the binary already exists in chacra, do not rebuild
+if [ "$binary_exists" = true ] ; then
```
alfredodeza: the command returns `0` for an existing url and `1` for a non existing one, I don't think that just checking the output would be what we want here. We should trap the exit code with `$?`

    if [ $? -e 0 ]; then
        echo "The endpoint at ${chacra_endpoint} already exists and FORCE_REBUILD was not used, Exiting..."
    fi

ceph-build-next/build/build_deb
```diff
@@ -6,6 +6,19 @@ if test -f /etc/redhat-release ; then
...
+    echo "The binary at ${chacra_endpoint} already exists, Exiting..."
```
alfredodeza: Like I mentioned in the comment above, we are not really checking for binaries, just endpoints to the url where the binaries would live. It is meant to work that way because a job like Ceph produces a lot of binaries, not just one. So this `echo` should really be: 

    echo "The endpoint at ${chacra_endpoint} already exists and FORCE_REBUILD was not used, Exiting..."

ceph-build-next/build/build_deb
andrewschoen: Would I still need the ``binary_exists`` var in that case or can I trash it?

PR #162, opened 2015-10-27
check the exit status on the chacractl exists command; add a FORCE flag to ceph-next

andrewschoen:
If FORCE is checked then binaries will be built and pushed to chacra even if they already exist in chacra.

### DIFF
--- a/ceph-build-next/build/build_deb
+++ b/ceph-build-next/build/build_deb
@@ -14,7 +14,7 @@ chacra_endpoint="ceph/${chacra_ref}/${distro}/{$dist}/${ARCH}"
 chacractl binary exists ${chacra_endpoint}
 
 # if the binary already exists in chacra, do not rebuild
-if [ $? -e 0 ] ; then
+if [ $? -e 0 ] && [ "$FORCE" = false ] ; then
     echo "The endpoint at ${chacra_endpoint} already exists, Exiting..."
     exit 0
 fi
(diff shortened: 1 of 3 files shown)


ceph-build-next/build/build_deb
```diff
@@ -11,11 +11,11 @@ vers=`cat ./dist/version`
 distro=`python -c "exec 'import platform; print platform.linux_distribution()[0].lower()'"`
 chacra_endpoint="ceph/${chacra_ref}/${distro}/{$dist}/${ARCH}"
```
alfredodeza: for `exists` it needs the full URL because you can check binaries or repos, while `chacractl binary` requires only the url from the project name forwards

ceph-build-next/build/build_deb
```diff
@@ -11,11 +11,11 @@ vers=`cat ./dist/version`
...
 
-binary_exists=`chacractl binary exists ${chacra_endpoint}`
+chacractl binary exists ${chacra_endpoint}
```
alfredodeza: which would make this fail.

ceph-build-next/build/build_deb
```diff
@@ -11,11 +11,11 @@ vers=`cat ./dist/version`
...
 
 # if the binary already exists in chacra, do not rebuild
-if [ "$binary_exists" = true ] ; then
-    echo "The binary at ${chacra_endpoint} already exists, Exiting..."
+if [ $? -e 0 ] && [ "$FORCE" = false ] ; then
+    echo "The endpoint at ${chacra_endpoint} already exists, Exiting..."
```
alfredodeza: `... already exists and "FORCE" was not set, Exiting...`

andrewschoen:
@alfredodeza thanks, updated

PR #170, opened 2015-10-28
force push to ceph-releases

alfredodeza:

### DIFF
--- a/ansible/roles/ceph-release/tasks/main.yml
+++ b/ansible/roles/ceph-release/tasks/main.yml
@@ -66,8 +66,12 @@
   # git tag -s "v{{ version }}"  -u 17ED316D -m "v{{ version }}" chdir=ceph
   command: git tag "v{{ version }}" -m "v{{ version }}" chdir=ceph
 
-- name: push changes to ceph-releases git repo
-  command: git push origin {{ branch }} chdir=ceph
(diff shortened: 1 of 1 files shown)


ansible/roles/ceph-release/tasks/main.yml
```diff
@@ -66,8 +66,12 @@
...
   command: git tag "v{{ version }}" -m "v{{ version }}" chdir=ceph
 
-- name: push changes to ceph-releases git repo
-  command: git push origin {{ branch }} chdir=ceph
+  # 'origin' in this case is ceph-releases, since this is our fork/copy of the
+  # official ceph repo it is ok to forcefully push "losing" changes. This may
+  # come up if we need to re-do a release which means doing a fresh commit to
+  # the debian changelog and rpm files and tagging.
+- name: force push changes to ceph-releases git repo
+  command: git push -f origin {{ branch }} chdir=ceph
```
andrewschoen: +1 to the force push, you might think about using the git module here, which allows for a "force" option.

http://docs.ansible.com/ansible/git_module.html

ktdreyer:
LGTM

ansible/roles/ceph-release/tasks/main.yml
alfredodeza: the git module doesn't have a push option that I could see.

PR #174, opened 2015-10-28
Setup releases

alfredodeza:
fix the github url


ktdreyer:
This has to be an SSH URL in order to authenticate to GitHub :(

PR #177, opened 2015-10-28
ansible: add github.com host key on slaves

ktdreyer:
This prevents "Host key verification failed" errors when operating with GitHub over SSH.


alfredodeza:
looks ok to me

For explicit clones with ansible we can do:

https://github.com/ceph/ceph-build/blob/master/ansible/roles/ceph-release/tasks/main.yml#L12

ansible/slave.yml
```diff
@@ -81,6 +81,13 @@
       # docs.ceph.com.pub is the output of `ssh-keyscan docs.ceph.com`
       key="{{ lookup('file', 'ssh/hostkeys/docs.ceph.com.pub') }}"
 
+    - name: add github.com host key
+      sudo: true
+      known_hosts: path='/etc/ssh/ssh_known_hosts'
```
andrewschoen: I don't think that key=value works here with newlines. I think you'll want key: value, each on a new line.

ktdreyer: You're right. I've fixed this to use the proper syntax

PR #188, opened 2015-10-29
remove old docs that relied on a convention that is not enforced

alfredodeza:


ktdreyer:
thanks!

PR #190, opened 2015-10-29
ceph-build-next: fix the grep that finds the binaries to push to chacra

andrewschoen:


ktdreyer:
LGTM

alfredodeza:
that grep will not work, it has an extra slash

I tried this and it seemed correct:

    find . | egrep "*\.(changes|deb|dsc|gz)$" | egrep -v "(Packages|Sources|Contents)"

PR #191, opened 2015-10-29
ansible: set HSTS header in nginx

ktdreyer:
This will cause a browser to automatically load the "https://" URL when the user enters "jenkins.ceph.com" into the browser.

We are using HTTP redirects from HTTP -> HTTPS already on jenkins.ceph.com, but this change should make it a bit more secure.

See https://www.owasp.org/index.php/HTTP_Strict_Transport_Security for more information


ktdreyer:
once your browser sees this `Strict-Transport-Security` HTTP header, it will remember to only load the site over HTTPS from that point on. In other words, the browser will seamlessly redirect any attempts to visit "http://jenkins.ceph.com" to the https:// version, without even making the plaintext HTTP request first.

PR #192, opened 2015-10-29
fix calamari jobs' "branches" definition

ktdreyer:
A bug in jenkins-job-wrecker screwed up the "branches" settings when we
imported these jobs from the internal Calamari jenkins server. Fix this.


alfredodeza:
this jenkins-job-wrecker is a problem we can't trust it

ktdreyer:
that's the problem with open-source, it's bugs all the way down

alfredodeza:
This wouldn't be an issue if you were bundling stuff.

ppdrsaw

ktdreyer:
BRB, bundling https://pypi.python.org/pypi/zope.cooties/

PR #193, opened 2015-11-02
ceph-release-rpm: use download.ceph.com GPG key URL

ktdreyer:
This should be more reliable than the Gitweb interface, and matches what we've done in ceph's docs, ceph-deploy, etc.


ktdreyer:
Eventually this package ought to ship a static copy of the GPG key, but let's get it ported to chacra first

PR #194, opened 2015-11-02
port ceph-release-rpm to use chacra

andrewschoen:


ktdreyer:
LGTM

alfredodeza:
@andrewschoen we need to be able to force push things to chacra, those need to be defined here like they were for ceph-next

andrewschoen:
@alfredodeza ok, I can add in the force option and the exists check. I don't think that chacractl has the ability to update a binary yet though, does it? At least that's what I remember seeing during testing of ceph-next.

alfredodeza:
@andrewschoen I just released 0.0.3 which has a `chacractl binary --force` flag to re-upload binaries if they exist already

andrewschoen:
@alfredodeza I've got the FORCE and TEST flags added to this now.

PR #196, opened 2015-11-03
Ceph retag

alfredodeza:
avoid re-tagging if we have already created one and we are re-building


ktdreyer:
Now that I see this, should we name this simply `TAG` instead of `RETAG`?
(Also, `precent` -> `prevent` in commit log)

alfredodeza:
what do you have against *precent*? I feel like it reads fine

PR #204, opened 2015-11-05
Create `rpmmacros` file

alfredodeza:
So that we don't end up with RPM binaries that have 'centos' in there

### DIFF
--- a/ansible/files/rpmmacros.j2
+++ b/ansible/files/rpmmacros.j2
@@ -1,1 +1,1 @@
-%dist .el{{ ansible_lsb.major_release|int }}
+%dist .el{{ ansible_lsb.major_release }}
--- a/ansible/files/rpmmacros.j2
+++ b/ansible/files/rpmmacros.j2
@@ -1,0 +1,1 @@
+%dist .el{{ ansible_lsb.major_release|int }}
--- a/ansible/slave.yml
+++ b/ansible/slave.yml
@@ -7,10 +7,10 @@
    - jenkins_user: 'jenkins-build'
   tasks:
     - name: create a jenkins-build user
-      user: name=jenkins-build comment="Jenkins Build Slave User"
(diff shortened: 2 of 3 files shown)


ansible/slave.yml
```diff
@@ -7,10 +7,10 @@
    - jenkins_user: 'jenkins-build'
   tasks:
     - name: create a jenkins-build user
-      user: name=jenkins-build comment="Jenkins Build Slave User"
+      user: name=jenkins-build comment="Jenkins Build Slave User" home="/home/{{ jenkins_user }}"
```
andrewschoen: Should we do ``name="{{ jenkins_user }}"`` here instead of hardcoding jenkins-build?

ansible/slave.yml
```diff
@@ -78,10 +85,11 @@
...
       sudo: true
-      known_hosts: path='/etc/ssh/ssh_known_hosts'
-      name='docs.ceph.com'
-      # docs.ceph.com.pub is the output of `ssh-keyscan docs.ceph.com`
-      key="{{ lookup('file', 'ssh/hostkeys/docs.ceph.com.pub') }}"
+      known_hosts:
+        path: '/etc/ssh/ssh_known_hosts'
+        name: 'docs.ceph.com'
+        # docs.ceph.com.pub is the output of `ssh-keyscan docs.ceph.com`
+        key: "{{ lookup('file', 'ssh/hostkeys/docs.ceph.com.pub') }}"
```
ktdreyer: gah, sorry about this, thanks for fixing it.

ansible/files/rpmmacros.j2
```diff
@@ -0,0 +1 @@
+%dist .el{{ ansible_lsb.major_release|int }}
```
ktdreyer: can we skip the cast to an `int` here?

ktdreyer: actually `{{ ansible_distribution_major_version }}` will do it

PR #215, opened 2015-11-09
do not activate virtualenv when building

alfredodeza:
So that we avoid these things:

   $ rpm -qpR ceph-9.2.0-0.el7.x86_64.rpm
   /bin/bash
   /bin/sh
   /bin/sh
   /bin/sh
   /bin/sh
   /bin/sh
   /home/jenkins-build/build/workspace/ceph-build-next/ARCH/x86_64/DIST/centos7/venv/bin/python
   /usr/bin/env
   ...

### DIFF
--- a/ceph-build-next/build/build_deb
+++ b/ceph-build-next/build/build_deb
@@ -6,8 +6,7 @@ if test -f /etc/redhat-release ; then
     exit 0
 fi
 
-# activate the virtualenv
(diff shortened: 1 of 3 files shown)


ktdreyer:
Longer-term, we need to make the RPM build process use `mock` (similar to Debian's `pbuilder`) to avoid the local environment entirely.

PR #216, opened 2015-11-09
do not source the venv in ceph-deploy and ceph-release-rpm jobs

andrewschoen:
When we source the virtualenv the virtualenv path is linked within the
binaries.


ktdreyer:
Longer-term, we need to make the RPM build process use `mock` (similar to Debian's `pbuilder`) to avoid the local environment entirely.

PR #218, opened 2015-11-09
use --upgrade when installing chacractl

andrewschoen:
This will allow us to not have to change the CHACRACTL_VERSION variable
each time we release a new version of chacractl.


ktdreyer:
awesome

PR #219, opened 2015-11-10
always update the pip cache before installing packages

andrewschoen:
We do this to get around a bug in the previous code that would never go
out to look for new packages if something in the cache satisfied the
requirements. Specifically, we wanted chacractl==0.0.5 installed, but
because 0.0.4 exists in the cache and satisified the requirement of
>=0.0.4 the cache was never updated to include 0.0.5.

### DIFF
--- a/ceph-deploy/build/setup
+++ b/ceph-deploy/build/setup
@@ -19,7 +19,7 @@ fi
 
 # Create the virtualenv
 virtualenv $WORKSPACE/venv
-source $WORKSPACE/venv/bin/activate
+VENV="$WORKSPACE/venv/bin"
 
 # Define and ensure the PIP cache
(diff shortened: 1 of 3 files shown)


ceph-deploy/build/setup
```diff
@@ -19,7 +19,7 @@ fi
 
 # Create the virtualenv
 virtualenv $WORKSPACE/venv
-source $WORKSPACE/venv/bin/activate
+VENV="$WORKSPACE/venv/bin"
```
alfredodeza: I thought this was already fixed, no?

andrewschoen: Nope, we apparently missed this one in the setup of ceph-deploy. We fixed it in the build script yesterday though.

PR #221, opened 2015-11-12
ceph-setup-next: correct $BRANCH variable in output

ktdreyer:
$Branch is undefined. Use $BRANCH.

arior to this change, Jenkins would not print the branch name to the
console log during the "ceph-setup-next" job.


ktdreyer:
whoops, "arior" -> "Prior" :-) Hopefully our future selves will know what I meant.

PR #223, opened 2015-11-12
creates a utility to install python packages and configure chacractl

andrewschoen:
This way we can have a single bit of code that installs python packages and configures chacractl as these are very common steps in our build jobs.

This also ports ceph-build, ceph-deploy and ceph-release-rpm to use these utils.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -30,3 +30,12 @@ install_python_packages () {
         $VENV/pip install --upgrade --exists-action=i --find-links="file://$PIP_SDIST_INDEX" --no-index $package
     done
 }
+
+make_chacractl_config () {
+    # create the .chacractl config file
+    cat > $HOME/.chacractl << EOF
+    url = "$CHACRACTL_URL"
+    user = "$CHACRACTL_USER"
+    key = "$CHACRACTL_KEY"
+    EOF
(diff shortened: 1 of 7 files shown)


scripts/build_utils.sh
```diff
@@ -0,0 +1,40 @@
...
+        echo $package
+        # download packages to the local pip cache
+        $VENV/pip install --upgrade --exists-action=i --download="$PIP_SDIST_INDEX" $package
+        # install packages from the local pip cache, ignoring pypi
+        $VENV/pip install --upgrade --exists-action=i --find-links="file://$PIP_SDIST_INDEX" --no-index $package
+    done
+}
+
+make_chacractl_config () {
+# create the .chacractl config file
```
ktdreyer: sorry this is really pedantic, but do you mind indenting the code in this function, so it matches the function above?

andrewschoen: I originally had it that way, but there was a syntax error related to the ``EOF``. Is there a better way to do that?

andrewschoen: Here's a link to that error I was seeing when the code was indented. https://jenkins.ceph.com/job/ceph-deploy/19/ARCH=x86_64,DIST=trusty/console

PR #224, opened 2015-11-13
ceph-build-pull-requests: check ansible syntax

ktdreyer:
With this change, the ceph-build-pull-requests job will run `ansible-playbook` with the `--syntax-check` flag on every playbook in the top-level of `./ansible/`. This will help us avoid simple Ansible syntax errors.

### DIFF
--- a/ceph-build-pull-requests/build/build
+++ b/ceph-build-pull-requests/build/build
@@ -29,3 +29,14 @@ for dir in `find . -maxdepth 1 -path ./.git -prune -o -type d -print`; do
         venv/bin/jenkins-jobs test $definitions_dir -o /tmp/output
     fi
 done
+
+# Install Ansible
+if ! venv/bin/pip install --find-links="file://$PIP_SDIST_INDEX" --no-index ansible; then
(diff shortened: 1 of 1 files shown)


ceph-build-pull-requests/build/build
```diff
@@ -29,3 +29,14 @@ for dir in `find . -maxdepth 1 -path ./.git -prune -o -type d -print`; do
...
 done
+
+# Install Ansible
+if ! venv/bin/pip install --find-links="file://$PIP_SDIST_INDEX" --no-index ansible; then
+    venv/bin/pip install --download-directory="$PIP_SDIST_INDEX" ansible
+    venv/bin/pip install --find-links="file://$PIP_SDIST_INDEX" --no-index ansible
+fi
+
+# Syntax-check each Ansible playbook
+for playbook in ansible/*.yml; do
```
alfredodeza: this is probably a bit overkill, but would you mind adding the `$WORKSPACE` variable here so that it reads as a full path? Like:

    for playbook in "$WORKSPACE/ansible/*.yml"; do

ktdreyer: Sure, amended and re-pushed.

PR #226, opened 2015-11-16
Make the radosgw-agent job push to chacra.ceph.com

andrewschoen:
This also moves a few more things into build_utils.sh. I'll make another PR to use those methods in the other jobs that need them.

This branch was tested here: https://jenkins.ceph.com/job/radosgw-agent/8/

### DIFF
--- a/radosgw-agent/build/build
+++ b/radosgw-agent/build/build
@@ -35,13 +35,7 @@ if [[ -f /etc/redhat-release || -f /usr/bin/zypper ]] ; then
 
         chacra_endpoint="radosgw-agent/${chacra_ref}/${DISTRO}/${DISTRO_VERSION}"
 
-        $VENV/chacractl exists binaries/${chacra_endpoint}/noarch ; exists=$?
-
-        # if the binary already exists in chacra, do not rebuild
-        if [ $exists -eq 0 ] && [ "$FORCE" = false ] ; then
-            echo "The endpoint at ${chacra_endpoint}/noarch already exists and FORCE was not set, Exiting..."
-            exit 0
-        fi
+        check_binary_existence $chacra_endpoint/noarch
 
         suse=$(uname -n | grep -ic -e suse -e sles || true)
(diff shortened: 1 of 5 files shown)


radosgw-agent/build/build
```diff
@@ -23,21 +44,22 @@ if [[ -f /etc/redhat-release || -f /usr/bin/zypper ]] ; then
...
+                find dist -name "*.noarch.rpm" | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpoint}/noarch
+                find dist -name "*.src.rpm" | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpoint}/source
             fi
         else
+            mkdir -p $HOME/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
+
             python setup.py clean
             python setup.py sdist --formats=gztar
-            rpmdev-setuptree
-            rpmdev-wipetree
```
ktdreyer: I don't feel good about this final change here; I'd rather we just install rpmdevtools on all the CentOS slaves

ktdreyer: The issue with re-using `$HOME/rpmbuild/` all the time is that files from one build (say, ceph-deploy) could contaminate another build. It's actually safer to wipe the tree before each build

andrewschoen: Alright, I've removed this commit and added ``rpmdevtools`` to our ansible for jenkins slaves instead.

We'll want to make the ceph-deploy job do this as well.

radosgw-agent/build/build
```diff
@@ -10,12 +10,33 @@ echo "  WS=$WORKSPACE"
 echo "  PWD=$(pwd)"
 ls -l
 
+
+# the following two methods exist in scripts/build_utils.sh
```
alfredodeza: thanks for this comment

alfredodeza: also, HI BASH

PR #227, opened 2015-11-16
ansible: add lxml deps for docs.ceph.com builds

ktdreyer:
The docs require these two packages in order to build.

### DIFF
--- a/ansible/slave.yml
+++ b/ansible/slave.yml
@@ -73,6 +73,8 @@
         - doxygen
         - ditaa
         - ant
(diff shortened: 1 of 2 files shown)


alfredodeza:
This means that the same slaves needed for Ceph are the ones that will build docs for Ceph too right? (a rebase is needed here too)

ktdreyer:
Yeah, let's just close this.

We should manage the entire "docs.dhc.sepia.ceph.com" VM with Ansible instead.

PR #230, opened 2015-11-17
ceph-build: separate deb build scripts from rpm build scripts

andrewschoen:
We need these specific scripts grouped into their own shell directives
so that if setup_deb exits on a centos node then setup_rpm will still be
ran against that same node.


ktdreyer:
Thanks guys

PR #231, opened 2015-11-17
add the 'huge' label to ensure competent machines are used for building

alfredodeza:


ktdreyer:
that looks right to me

PR #246, opened 2015-11-24
use ports mirror for arm packages in pbuilder

alfredodeza:


ceph-build/config/definitions/ceph-build.yml
```diff
@@ -21,6 +21,7 @@
           name: ARCH
           values:
             - x86_64
+            - arm64
```
ktdreyer: We're modifying the matrix here, but that doesn't really seem like it's "using ports mirror for arm packages in pbuilder" ... is this PR missing a commit?

PR #251, opened 2015-11-30
implement a ceph-deploy-tag job

alfredodeza:
So that it runs before the actual ceph-deploy release job, using the new ansible playbooks and allowing us to basically release with a single, push-button action.

### DIFF
--- a/ceph-deploy-tag/build/build
+++ b/ceph-deploy-tag/build/build
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -ex
+
+if [ "$TAG" = false ] ; then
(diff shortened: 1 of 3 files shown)


ceph-deploy-tag/build/build
```diff
@@ -0,0 +1,31 @@
...
+
+set -ex
+
+if [ "$TAG" = false ] ; then
+    echo "Assuming tagging process has succeeded before because TAG was set to false"
+    exit 0
+fi
+
+# Create the virtualenv
+virtualenv venv
```
andrewschoen: We should use the new scripts in scripts/build_utils.sh to make the virtualenv and install ansible

PR #264, opened 2015-12-08
also remove venv that is left behind

alfredodeza:

### DIFF
--- a/ceph-build/config/definitions/ceph-build.yml
+++ b/ceph-build/config/definitions/ceph-build.yml
@@ -37,6 +37,7 @@
       - shell: |
           echo "Cleaning up workarea"
           rm -rf dist
+          rm -rf venv
           rm -rf release
       - copyartifact:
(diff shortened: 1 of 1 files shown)


ktdreyer:
we don't want "wipe workspace" ?

alfredodeza:
I don't know. Maybe? I was just finding that venv was left behind and had to `rm -rf` manually

alfredodeza:
Can try using `wipe workspace`

alfredodeza:
@ktdreyer we can't use `wipe-workspace: true` because that is an `scm` option that will remove/wipe the previous cloned repository. This `venv` directory exists outside of the cloned repository so that option will not help in this particular case.

ktdreyer:
Ok that makes sense then!

ktdreyer:
https://github.com/ceph/ceph-build/pull/265 will clarify the "workarea" comment in this vicinity

PR #266, opened 2015-12-22
create a ceph-pull-requests job

alfredodeza:
This would replace the make-check-bot. 

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -1,0 +1,64 @@
+- job:
+    name: ceph-pull-requests
+    project-type: freestyle
+    defaults: global
+    display-name: 'ceph: Pull Requests'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - github:
+          url: https://github.com/ceph/ceph/
+    logrotate:
+      daysToKeep: 15
+      numToKeep: 30
+      artifactDaysToKeep: -1
+      artifactNumToKeep: -1
+
+    parameters:
+      - string:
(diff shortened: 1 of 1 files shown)


dachary:
:+1:

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    triggers:
+      - github-pull-request:
+          admin-list:
+            - alfredodeza
+            - ktdreyer
+            - andrewschoen
+            - dmick
+          org-list:
+            - ceph
```
ktdreyer: IIRC "org" never worked with our old Jenkins setup, but it would be awesome if it's finally working now.

Otherwise I suspect we'll want to expand "white-list" here.

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -0,0 +1,64 @@
...
+          auto-close-on-fail: false
+
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph.git
+          branches:
+            - ${sha1}
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
```
dachary: ok if it's related to extracting the git clone and it is in minutes. The make check itself including compilation can take up to 2 hours.

liewegas:
OT, but it would be awesome if one of these plugins nagged people about Signed-off-by: (and, say, set a tag missing-signed-off-by until they fix it).

alfredodeza:
@liewegas it might be a bit tricky to determine getting a subset of commits that belong to the PR only. But once we have this in place I think it will be easier to work a bit on that

PR #273, opened 2016-01-04
ansible playbook for sensu server and client setup

andrewschoen:
This playbook uses all third party ansible roles (except for a small redis role I included here).  I had to fork the sensu role to update it for the newest version of sensu and to fix a few things related to client setup and ruby gem plugin installation. I'm not sure we'll want to keep using that branch in my repo or move it elsewhere eventually. I submitted a patch upstream for some of that work, but the project doesn't seem terribly active anymore.

I also still need a good mechanism for overwriting the passwords defined in sensu.yml. Right now using ``--extra-vars`` would be the only way to accomplish that.


andrewschoen:
test this please

ansible/sensu.yml
```diff
@@ -0,0 +1,52 @@
...
+      sensu_client_subscription_names:
+        - common
+        - rabbitmq
+      # these are custom settings for this client
+      sensu_settings:
+        # used in the rabbitmq-alive check
+        rabbitmq:
+          user: sensu
+          password: secret
+          # we need to escape the / in /sensu
```
alfredodeza: I don't even want to know how you got here

PR #278, opened 2016-01-06
do not start with a newline when creating the xml

alfredodeza:

### DIFF
--- a/ceph-pr-commits/build/build
+++ b/ceph-pr-commits/build/build
@@ -25,10 +25,9 @@ def produce_junit(**kw):
     time = kw.get('time', 0)
     tests = kw.get('tests', [])
     failures = kw.get('failures', [])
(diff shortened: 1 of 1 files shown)


ceph-pr-commits/build/build
```diff
@@ -25,10 +25,9 @@ def produce_junit(**kw):
     time = kw.get('time', 0)
     tests = kw.get('tests', [])
     failures = kw.get('failures', [])
-    total_tests = len(tests)
+    total_tests = len(tests) or 1
```
ktdreyer: I don't think we need this, right?

alfredodeza: I think we do because we are at least running one test always :)

ktdreyer: ok, that works for me

PR #284, opened 2016-01-08
ceph-pull-requests: require a centos7 node and reboots after make check

andrewschoen:
Fixes: http://tracker.ceph.com/issues/14172


dachary:
:+1:

PR #287, opened 2016-01-08
ceph-pull-requests: use a postbuildscript to only reboot on failure

andrewschoen:
This way we can report the correct status to github and still reboot the node on failure.  Before this change the node would never reboot if the make check failed.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -47,7 +47,11 @@
 
     builders:
       - shell: "./run-make-check.sh"
-      - shell: "sudo reboot"
 
     publishers:
       - github-notifier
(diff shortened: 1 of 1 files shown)


ktdreyer:
LGTM

dachary:
:+1:

dachary:
For the record, I learnt about publishers ( http://docs.openstack.org/infra/jenkins-job-builder/publishers.html ) and the **postbuild** ( http://docs.openstack.org/infra/jenkins-job-builder/publishers.html#publishers.postbuildscript ).

Is there a reason to prefer **postbuild** over **post-task** ( http://docs.openstack.org/infra/jenkins-job-builder/publishers.html#publishers.post-tasks ) ? Not that I think **post-task** is more suited, just out of curiosity :-) I would have never discovered this solution by myself !

PR #290, opened 2016-01-12
Verbose commit check

alfredodeza:
So that we can also track when something goes wrong. Right now it is very difficult to tell what commands are being run so that we can debug this.

### DIFF
--- a/ceph-pr-commits/build/test_commits.py
+++ b/ceph-pr-commits/build/test_commits.py
@@ -14,6 +14,7 @@ ceph_checkout = os.path.join(workspace, 'ceph')
 
 
 def run(command):
+    print "Running command: %s" % ' '.join(command)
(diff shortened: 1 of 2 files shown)


ceph-pr-commits/build/test_commits.py
```diff
@@ -14,6 +14,7 @@
 
 
 def run(command):
+    print "Running command: %s" % ' '.join(command)
```
tchaikov: @alfredodeza do you want to print out the stdout.read() also just for debugging purpose?

tchaikov: oh, nvm, i guess that what cfcd1ac80aea06d6d29b683b60008b6f976a9dfc is for.

tchaikov:
lgtm.

PR #293, opened 2016-01-13
default to number of executors to just 1

alfredodeza:
So that we don't get nodes rebooted by script like make-check.sh while we are doing a release.

This is why rebooting a node because a test requires it is not just a matter of slapping "sudo reboot" on a script.

This option should at some point be made optional so that mita can toggle it with `huge` nodes only.


andrewschoen:
test this please

andrewschoen:
I think the failing check here is a problem with jenkins and not this actual PR.

alfredodeza:
test this please

PR #299, opened 2016-01-18
ceph-pr-commits: only check for divergent commits from master

theanalyst:
We are currently checking all the commits, which causes failure due to
a single commit that wasnt signed. Check only the commits in the pr vs
master for now

Signed-off-by: Abhishek Lekshmanan <abhishek@suse.com>


theanalyst:
#300 should fix this .. closing

PR #305, opened 2016-01-25
when the build is done, tell mita to destroy the node to ensure it gets a new one

alfredodeza:
This will require some mita configuration to understand what a "unique" machine means.

### DIFF
--- a/ceph-ansible-pull-requests/config/definitions/ceph-ansible-pull-requests.yml
+++ b/ceph-ansible-pull-requests/config/definitions/ceph-ansible-pull-requests.yml
@@ -63,7 +63,7 @@
     publishers:
       - github-notifier
       - postbuildscript:
-          script-only-if-succeeded: False
+          script-only-if-succeeded: True
           script-only-if-failed: True
           builders:
             - shell: "curl -X POST https://mita.ceph.com/api/nodes/${NODE_NAME}/delete/"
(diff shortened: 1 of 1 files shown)


ceph-ansible-pull-requests/config/definitions/ceph-ansible-pull-requests.yml
```diff
@@ -59,3 +59,11 @@
           !include-raw:
             - ../../../scripts/build_utils.sh
             - ../../build/build
+
+    publishers:
+      - github-notifier
+      - postbuildscript:
+          script-only-if-succeeded: False
```
andrewschoen: Don't we want to do this regardless of if the job succeeds or fails? The way it is now it'd only happen on failure.

ceph-ansible-pull-requests/config/definitions/ceph-ansible-pull-requests.yml
```diff
@@ -59,3 +59,11 @@
...
             - ../../../scripts/build_utils.sh
             - ../../build/build
+
+    publishers:
+      - github-notifier
+      - postbuildscript:
+          script-only-if-succeeded: True
+          script-only-if-failed: True
+          builders:
+            - shell: "curl -X POST https://mita.ceph.com/api/nodes/${NODE_NAME}/delete/"
```
andrewschoen: I think mita might need some work here to support a jenkins $NODE_NAME and not a mita UUID.

PR #315, opened 2016-01-26
add osd_auto_discovery to extra-vars; add -k to the curl call to mita

andrewschoen:


alfredodeza:
I am not sure we need to use auto discovery with the work you've done on adding a second device. Or am I just interpreting the option the other way around?

andrewschoen:
@alfredodeza you have to set that or pass '/dev/vdb' as ``devices`` to --extra-vars, afaik. The catch is that ``devices`` has to be a list and I've never figured out a way to use a list with --extra-vars and the k=v format, I think you have to use json. Setting ``osd_auto_discovery=true`` just seemed easier than converting everything we already had to json.

PR #325, opened 2016-02-17
ceph-installer pull request job

alfredodeza:
It removes the old mariner job that would actually build binaries and will now use tox instead.

Support for tox builds is in a ceph-installer pr.


ceph-installer-pull-requests/config/definitions/ceph-installer-pull-requests.yml
```diff
@@ -1,49 +1,36 @@
...
-          name: ARCH
-          type: label-expression
-          values:
-            - x86_64
-      - axis:
-          name: DIST
-          type: label-expression
-          values:
-            - trusty
+    name: ceph-installer-pull-requests
```
andrewschoen: I think we need to define ``node`` somewhere here. Maybe ``small && trusty``?

PR #330, opened 2016-03-01
ceph-installer coverage

alfredodeza:
It will create two reports: junit report from py.test and coverage report from pytest-coverage

Should run only on merges to master (similar to when we build docs)

### DIFF
--- a/ceph-installer-coverage/build/build
+++ b/ceph-installer-coverage/build/build
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# the following two methods exist in scripts/build_utils.sh
+pkgs=( "pytest" "pytest-coverage" )
+install_python_packages "pkgs[@]"
+
+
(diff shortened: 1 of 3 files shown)


ceph-installer-merged/config/definitions/ceph-installer-merged.yml
```diff
@@ -0,0 +1,47 @@
+- job:
+    name: ceph-installer-merged
```
andrewschoen: why not call this job ceph-installer-coverage?

alfredodeza: I was thinking of making this the "when it merges to master" job, but happy to update the name if you think it shouldn't have anything else piled onto it

alfredodeza:
@andrewschoen I updated everything to reflect the correct naming for the job

PR #337, opened 2016-03-22
ceph-build-arm64: Test for arm64 build, separate from ceph-build

dmick:
Some things want to be different in this build; let's keep them
away from the main ceph build until we're sure they can coexist
peacefully.

### DIFF
--- a/ceph-build-arm64/config/definitions/ceph-build.yml
+++ b/ceph-build-arm64/config/definitions/ceph-build.yml
@@ -1,0 +1,56 @@
+- job:
+    name: ceph-build-arm64
+    project-type: matrix
+    disabled: true
+    defaults: global
+    display-name: 'ceph-build-arm64'
+    block-downstream: false
+    block-upstream: false
+    properties:
+      - github:
+          url: https://github.com/ceph/ceph
+    execution-strategy:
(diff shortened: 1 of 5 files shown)


ceph-build-arm64/config/definitions/ceph-build.yml
```diff
@@ -0,0 +1,62 @@
...
+          values:
+            - xenial
+
+    builders:
+      - shell: |
+          echo "Cleaning up top-level workarea (shared among workspaces)"
+          rm -rf dist
+          rm -rf venv
+          rm -rf release
+      - copyartifact:
```
alfredodeza: I understand this is mostly copied from `ceph-build` but we don't need to copy artifacts any more (woo hoo!)

alfredodeza:
You have included the rpm scripts but this is only for Xenial. Is the plan to get arm64 for RPM servers to later on?

dmick:
Yes, Centos is next; I just haven't tried to make that happen yet, as my build slaves are in worse shape for it.  (The build slaves are not installable yet except by physical disk construction.  :( )

ceph-build-arm64/config/definitions/ceph-build.yml
dmick: cool

PR #343, opened 2016-03-29
ceph-setup-arm64: tweaks for building unreleased branches

dmick:
- pull from ceph.git, not ceph-releases
- always rebuild tarball
- name all source tars by $vers (x.y.z), not $cephvers (x.y.z-N-gSHA1)

### DIFF
--- a/ceph-setup-arm64/config/definitions/ceph-setup.yml
+++ b/ceph-setup-arm64/config/definitions/ceph-setup.yml
@@ -23,7 +23,7 @@
 
     scm:
       - git:
-          url: git@github.com:ceph/ceph-releases.git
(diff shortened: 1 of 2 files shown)


ceph-setup-arm64/config/definitions/ceph-setup.yml
```diff
@@ -23,7 +23,7 @@
 
     scm:
       - git:
-          url: git@github.com:ceph/ceph-releases.git
+          url: git@github.com:ceph/ceph.git
```
alfredodeza: would you mind explaining this change? we use ceph-releases to prevent publishing tags to ceph.git, this would revert that

PR #348, opened 2016-04-01
Remove separate -arm64 versions and integrate into main jobs

dmick:


ceph-build/config/definitions/ceph-build.yml
```diff
@@ -9,7 +9,7 @@
       - github:
           url: https://github.com/ceph/ceph
     execution-strategy:
-      combination-filter: (ARCH=="x86_64") || (DIST == "trusty" && ARCH == "arm64")
+      combination-filter: (ARCH=="x86_64") || (DIST == "xenial" && ARCH == "arm64")
```
alfredodeza: Since there aren't any other arm64 distros in our lineup, I think there is no need to specify `DIST == "xenial"` here.

dmick: No need, but it's a choice: we're only interested in building for xenial right now.  (and none of the rpm stuff is changed at all yet).

dmick:
I think this is ready to go now

PR #354, opened 2016-04-12
ceph-build: check chacra for ceph-osd package to test for "should I build"

dmick:


dmick:
I would like to find a good way to test this without merging to master, but I don't know what it would be

ceph-build/build/build_deb
```diff
@@ -57,6 +57,9 @@ gen_debian_version() {
 
 bpvers=`gen_debian_version $debian_version $DIST`
 
+# look for a specific package to tell if we can avoid the build
+chacra_endpoint="ceph/${chacra_ref}/${distro}/${DIST}/${ARCH}/ceph-osd_${vers}-${bpvers}_${ARCH}.deb"
```
alfredodeza: using `ceph-osd` here would make this check not work for Ceph versions where the split didn't exist (Hammer?)

ktdreyer:
> I would like to find a good way to test this without merging to master, but I don't know what it would be

The way I typically do this is that I run JJB's `jenkins-jobs` by hand on my local system, so that I've applied the change to jenkins.ceph.com prior to merging in GitHub. You can just modify the production "ceph-build" job this way, or maybe rename it... although using a renamed job is really hard.

It would be nice to have a staging environment for this sort of thing.

ceph-build/build/build_deb
dmick: argh, totally missed this comment.  OK, can change to librados2

PR #355, opened 2016-04-15
Dynamic axis for Ceph builds

alfredodeza:
This will allow to modify the DIST variable dynamically so that different distro versions can be used.

https://wiki.jenkins-ci.org/display/JENKINS/DynamicAxis+Plugin

### DIFF
--- a/ceph-build/config/definitions/ceph-build.yml
+++ b/ceph-build/config/definitions/ceph-build.yml
@@ -23,7 +23,7 @@
             - x86_64
             - arm64
       - axis:
-          type: label-expression
+          type: dynamic
           name: DIST
           values:
             - jessie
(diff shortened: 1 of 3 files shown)


dmick:
Alfredo explains: so that different distro axes can be used for different revisions of the project (say, hammer should build for different distros than jewel)

ceph-build/config/definitions/ceph-build.yml
```diff
@@ -23,16 +23,15 @@
             - x86_64
             - arm64
       - axis:
+          # the dynamic axis is implemented **oddly** in jenkins job builder.
+          # It will pick just the first item in `values` and assign that to the
+          # 'variable name' that is where `DIST` would get expanded too. This
+          # is not documented. This means that DIST_VERSIONS will need to exist
+          # as an environment variable passed onto this job.
```
dmick: since we've already got a comment, let's elaborate:  "..containing a space-separated list of all values to substitute for DIST in this job"

dmick:
Extra comment addressed with better descriptiong string in ceph.yml

PR #356, opened 2016-04-19
Add 'rebootable' label to ceph-pull-requests

dmick:
This is to signal node selection that the job will reboot the node
after it finishes.  Without an explicit label on the node and a request
for it, any 'trusty' node would satisfy, and be rebooted; this is
a problem for nodes that 'dial in' to the Jenkins master (such as
physical slaves on the sepia network).

This corresponds with the addition of 'rebootable' tags to mita's
configuration, so that mita nodes continue to suffice for this job.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -3,7 +3,7 @@
     project-type: freestyle
     defaults: global
     concurrent: true
-    node: huge && (centos7 || trusty)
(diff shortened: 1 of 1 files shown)


dmick:
This depends on the changes to mita's node config which I don't yet have a place to put

andrewschoen:
I think you might have meant to put this tag on the ceph-pull-requests job, it's the one that does a reboot after a make check failure.

dmick:
sigh, yep, will fix and repush

alfredodeza:
test this please

andrewschoen:
test this please

PR #357, opened 2016-04-20
Base url fix

alfredodeza:
define a `check_url` and an `endpoint` one for chacra interactions

This fixes the issue where the same url would be used to check DEB (or RPM) binary existence
and to POST new binaries causing this:

POST /binaries/ceph/jewel/ubuntu/trusty/x86_64/librados2_10.2.0-10.2.0-1trusty_x86_64.deb/ HTTP/1.1" 400

Which is the wrong URL for DEB binaries. Similar problem was for RPMs

### DIFF
--- a/ceph-build/build/build_deb
+++ b/ceph-build/build/build_deb
@@ -58,11 +58,12 @@ gen_debian_version() {
 bpvers=`gen_debian_version $debian_version $DIST`
 
 # look for a specific package to tell if we can avoid the build
-chacra_endpoint="ceph/${chacra_ref}/${distro}/${DIST}/${ARCH}/librados2_${vers}-${bpvers}_${ARCH}.deb"
+chacra_endpoint="ceph/${chacra_ref}/${distro}/${DIST}/${ARCH}"
+chacra_check_url="${chacra_endpoint}/librados2_${vers}-${bpvers}_${ARCH}.deb"
 
 if [ "$THROWAWAY" = false ] ; then
     # this exists in scripts/build_utils.sh
-    check_binary_existence $chacra_endpoint
+    check_binary_existence $chacra_check_url
 fi
 
 HOST=$(hostname --short)
--- a/ceph-build/build/build_rpm
+++ b/ceph-build/build/build_rpm
@@ -61,11 +61,14 @@ vers=`cat ./dist/version`
 # slap -rc to the ref if we are doing a release-candidate build
 [ "$RC" = true ] && chacra_ref="$BRANCH-rc" || chacra_ref="$BRANCH"
 [ "$TEST" = true ] && chacra_ref="test"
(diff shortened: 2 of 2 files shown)


ErwanAliasr1:
Maybe we could add deb|rpm in the commit title to avoid having two commits that share the same title ?

ceph-build/build/build_rpm
```diff
@@ -116,6 +119,6 @@ cd "$WORKSPACE"
 
 if [ "$THROWAWAY" = false ] ; then
     # push binaries to chacra
-    find release/${vers}/rpm/*/SRPMS | grep rpm | $VENV/chacractl binary ${chacra_flags} create ${chacra_baseurl}/source
-    find release/${vers}/rpm/*/RPMS/* | grep rpm | $VENV/chacractl binary ${chacra_flags} create ${chacra_baseurl}/${ARCH}
+    find release/${vers}/rpm/*/SRPMS | grep rpm | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpointl}/source
```
ErwanAliasr1: Sounds like a typo s/endpointl/endpoint/

ceph-build/build/build_deb
```diff
@@ -58,11 +58,12 @@ gen_debian_version() {
 bpvers=`gen_debian_version $debian_version $DIST`
 
 # look for a specific package to tell if we can avoid the build
-chacra_endpoint="ceph/${chacra_ref}/${distro}/${DIST}/${ARCH}/librados2_${vers}-${bpvers}_${ARCH}.deb"
+chacra_endpoint="ceph/${chacra_ref}/${distro}/${DIST}/${ARCH}"
+chacra_check_url="${chacra_endpoint}/librados2_${vers}-${bpvers}_${ARCH}.deb"
```
ErwanAliasr1: Seems to brake line 178 as chacra_endpoint is no more the rpm itself isn't it ?

alfredodeza: rpm? this is for debs only. We don't need an endpoint pointing to a binary, we need one pointing to `ARCH`

ceph-build/build/build_rpm
alfredodeza: Good catch!

alfredodeza:
test this please

ceph-build/build/build_deb
ErwanAliasr1: Ok forget my point.

ErwanAliasr1:
My last comment is that we have two very similar script but one is including the arch in the endpoint url and not the other. Could have been better to get both using the same path as they share the same name.

alfredodeza:
@ErwanAliasr1 right, they are different that way because for RPMs we are selecting binaries to go to `/source` and `/{ARCH}` whereas the DEBs are just going to `/{ARCH}`

PR #368, opened 2016-04-22
Add 'rebootable' label to ceph-pull-requests

dmick:
This is to signal node selection that the job will reboot the node
after it finishes.  Without an explicit label on the node and a request
for it, any 'trusty' node would satisfy, and be rebooted; this is
a problem for nodes that 'dial in' to the Jenkins master (such as
physical slaves on the sepia network).

This corresponds with the addition of 'rebootable' tags to mita's
configuration, so that mita nodes continue to suffice for this job.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -3,7 +3,7 @@
     project-type: freestyle
     defaults: global
     concurrent: true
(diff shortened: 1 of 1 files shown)


dmick:
Duplicating https://github.com/ceph/ceph-build/pull/356 with a local-to-ceph-build branch to see if that's why the github status update is failing for the syntax-check job

dmick:
now the original at #356 passed.  

I have no idea.

PR #370, opened 2016-04-25
Pbuilder support

alfredodeza:
Creates an axis for pbuilder so that all DEB distros will have it.

Updates the combination filter to exclude CentOS.


alfredodeza:
test this please

dmick:
Trying to understand the motivation for this.

PR #371, opened 2016-04-25
ceph-pull-requests: Override processor numbers

ErwanAliasr1:
By default, run-make-check divide by 2 the number of processors to be used
during the make process.

When building a PR on our build farm, we need to use the full potential of the
build hosts.

By overriding NPROC with the actual number of processors, we speedup the build
process and use in a more efficient way the build hosts.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -46,7 +46,7 @@
           wipe-workspace: true
 
     builders:
-      - shell: "timeout 7200 ./run-make-check.sh"
+      - shell: "export NPROC=$(nproc); timeout 7200 ./run-make-check.sh"
 
     publishers:
       - github-notifier


ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -46,7 +46,7 @@
           wipe-workspace: true
 
     builders:
-      - shell: "timeout 7200 ./run-make-check.sh"
+      - shell: "export NPROC=$(nproc); timeout 7200 ./run-make-check.sh"
```
alfredodeza: is `nproc` available everywhere?

alfredodeza:
OK to test

alfredodeza:
this will need a rebase

ErwanAliasr1:
Rebased.

ErwanAliasr1:
@alfredodeza This is now rebased.

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
ErwanAliasr1: That's from coreutils, so that's safe and this is what run-make-check is also using. So that's not a new dep.

ktdreyer:
It strikes me as odd that run-make-check.sh is not doing this?

dmick:
run-make-check used the "standard" nproc/2.  nproc jobs is not magic but I agree it's probably better for make only. Really not sure about check.

ErwanAliasr1:
@ktdreyer Sounds like a personal-taste setup when you build on your local system.
For the build hosts, that is just killing the perf and doesn't use properly the ressources.we buy.

alfredodeza:
@ErwanAliasr1 can you rebase once more

alfredodeza:
@dmick does that mean that you don't see this PR as something that will improve times for make-check ? Is there a drawback?

ErwanAliasr1:
@alfredodeza rebased.

ErwanAliasr1:
@alfredodeza @dmick as per my PR on ceph (https://github.com/ceph/ceph/pull/8131) make check is now very sensible of "make -j" value.

dmick:
I didn't quite catch the meaning of "is now very sensible of make -j value". 

What concerns me is that the make check tests themselves may use multiple threads per test, and so having nproc of them may be overall slower.  If you've tested sufficiently to know that -j $(nproc) is the right number, even including the actual tests, then by all means drive on.

ErwanAliasr1:
@dmick I mean that my PR have been improving the //ism of some make check targets. By using the same parallelism value (which was = nproc on my setup), that PR have been dividing by 2 the build time. So yes, on a make check, using more cpu cores = shorter build time. So using make -j $(nproc) is reasonable and saves time.

alfredodeza:
So this looks OK to me. @dmick since you are raising questions, I defer to you to merge.

dmick:
OK. If you have the data I have no other objection.

alfredodeza:
@ErwanAliasr1 can you rebase this so we can merge?

ErwanAliasr1:
@alfredodeza just rebased.

ErwanAliasr1:
thx for the merge

PR #374, opened 2016-04-26
Use --extra-vars instead of Mako templating for the prado playbook

andrewschoen:
There is no need to maintain to slave playbooks. We can use regular jinja2 variables in the playbook and prado can just use --extra-vars when calling ansible-playbook to provide those values.

This needs: https://github.com/alfredodeza/prado/pull/13


ansible/slave.yml
```diff
@@ -5,17 +5,23 @@
...
+   # jenkins API credentials:
+   - api_user: 'ceph-jenkins'
+   - token: '{{ token }}'
+   - api_uri: 'https://jenkins.ceph.com'
+   - nodename: '{{ nodename }}'
+   - labels: '{{ labels }}'
+
   tasks:
     - name: create a {{ jenkins_user }} user
-      user: name={{ jenkins_user }} comment="Jenkins Build Slave User" home="/home/{{ jenkins_user }}"
```
alfredodeza: why the removal of `home` ? I *think* there might be environments where it might not create that dir

alfredodeza: I realized that you were merging the good parts that we already use in the .j2 file here

PR #377, opened 2016-04-28
ensure the current hostname is defined in /etc/hosts

alfredodeza:
This will prevent errors from builders with improper hostnames:

    sudo: unable to resolve host vm


ktdreyer:
This should only be needed on Ubuntu, right?

alfredodeza:
@ktdreyer it doesn't hurt to apply it everywhere (it is also idempotent). But yes, it currently is broken on the Ubuntu Trusty image

PR #386, opened 2016-05-04
ceph-setup: build from tag, not from branch

alfredodeza:


dmick:
This will be another thing to revisit/make more flexible when we want to build non-tag commits.

PR #392, opened 2016-05-20
ceph-build-os: Adding ceph-build-os job

ErwanAliasr1:
This new job is about building VM image intented to build Ceph base code.

It uses the edeploy project to define the content of the VM by defining :
- the distro
- the distro version
- a release version of that build

The ceph-builder target of edeploy intent to create xfs-based VMs that have all
the usual packages required to build ceph or edeploy.

This will save time at build-time as most of the deps will be already installed.

### DIFF
--- a/ceph-build-os/build/build
+++ b/ceph-build-os/build/build
@@ -18,3 +18,4 @@ if [ -z "$REPOSITORY" ]; then
 fi
 
 sudo make ceph-builders DIST=${DIST} DVER=${DVER} VERSION=${VERSION} VIRTUALIZED=ceph-builders.virt NO_MGNIDS=True NO_COMPRESSED_FILE=True REPOSITORY=${REPOSITORY} KEEP_PKGMNGR=True COMPRESSED=yes && DVER=${DVER} VERSION=${VERSION} upload.sh
+
--- a/ceph-build-os/build/build
+++ b/ceph-build-os/build/build
@@ -1,6 +1,32 @@
 #!/bin/bash
 
+upload() {
+	IMAGE=$(ls /var/lib/debootstrap/install/$DISTRO_RELEASE-$VERSION/*qcow2)
+	if [ -z "$IMAGE" ]; then
+	   echo "upload: No IMAGE found in /var/lib/debootstrap/install/$DISTRO_RELEASE-$VERSION/, Exiting"
+	   exit 1
+	fi
+
+	IMAGE_NAME=$(basename `echo $IMAGE | sed -e "s/\.img\.qcow2//g"`)
+
+	echo "upload: About to upload $IMAGE as $IMAGE_NAME"
+
+	# If the same image already exists, let's delete it
+	openstack image list -f csv | grep -qw "$IMAGE_NAME" && echo "upload: Deleting remote image" && openstack image delete "$IMAGE_NAME"
+
+	# Upload the image
+	echo "upload: Uploading $IMAGE"
+	openstack image create $IMAGE_NAME --disk-format qcow2 --container-format bare --file $IMAGE
+}
+
 set +x
+
+export OS_AUTH_URL="$OVH_AUTH_URL"
+export OS_TENANT_ID="$OVH_TENANT_ID"
+export OS_TENANT_NAME="$OVH_TENANT_NAME"
+export OS_USERNAME="$OVH_USERNAME"
+export OS_PASSWORD="$OVH_PASSWORD"
+export OS_REGION_NAME="$OVH_REGION"
 
 if [ -z $DIST ]; then
     echo "build: No DIST found, exiting !"
@@ -14,6 +40,36 @@ fi
 
 if [ -z $VERSION ]; then
     echo "build: No VERSION found, exiting !"
+    exit 1
+fi
+
+if [ -z "$OS_AUTH_URL" ]; then
+    echo "upload: Cannot find any valid OS_AUTH_URL"
+    exit 1
+fi
+
+if [ -z "$OS_TENANT_ID" ]; then
+    echo "build: Cannot find any valid OS_TENANT_ID"
+    exit 1
+fi
+
+if [ -z "$OS_TENANT_NAME" ]; then
+    echo "build: Cannot find any valid OS_TENANT_NAME"
+    exit 1
+fi
+
+if [ -z "$OS_USERNAME" ]; then
+    echo "build: Cannot find any valid OS_USERNAME"
+    exit 1
+fi
+
+if [ -z "$OS_PASSWORD" ]; then
+    echo "build: Cannot find any valid OS_PASSWORD"
+    exit 1
+fi
+
+if [ -z "$OS_REGION_NAME" ]; then
+    echo "build: Cannot find any valid OS_REGION_NAME"
     exit 1
 fi
 
@@ -37,7 +93,7 @@ if [ -z "$REPO_URL" ]; then
     echo "build: Using repository $REPO_URL"
 fi
 
+
 sudo edeploy activate-pkgmngr
 
-sudo make ceph-builders DIST=${DIST} DVER=${DISTRO_RELEASE} VERSION=${VERSION} VIRTUALIZED=ceph-builders.virt NO_MGNIDS=True NO_COMPRESSED_FILE=True REPOSITORY=${REPO_URL} KEEP_PKGMNGR=True COMPRESSED=yes && DVER=${DISTRO_RELEASE} VERSION=${VERSION} upload.sh
-
+sudo make ceph-builders DIST=${DIST} DVER=${DISTRO_RELEASE} VERSION=${VERSION} VIRTUALIZED=ceph-builders.virt NO_MGNIDS=True NO_COMPRESSED_FILE=True REPOSITORY=${REPO_URL} KEEP_PKGMNGR=True COMPRESSED=yes && upload
--- a/ceph-build-os/build/build
+++ b/ceph-build-os/build/build
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-sudo edeploy activate-pkgmngr;
-cd build;
+sudo edeploy activate-pkgmngr; 
+cd build; 
 
 if [ -z "$REPOSITORY" ]; then
     case "$DIST" in
--- a/ceph-build-os/build/build
+++ b/ceph-build-os/build/build
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+sudo edeploy activate-pkgmngr;
+cd build;
+
+if [ -z "$REPOSITORY" ]; then
+    case "$DIST" in
+        "xenial"|"trusty")
+            REPOSITORY="http://ubuntu.mirrors.ovh.net/ubuntu/"
+        ;;
+        "jessie")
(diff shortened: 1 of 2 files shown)


ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,62 @@
...
+      artifactDaysToKeep: -1
+      artifactNumToKeep: -1
+
+    parameters:
+        - string:
+            name: DIST
+            description: "The distribution codename like xenial,centos,jessie"
+
+        - string:
+            name: DVER
```
alfredodeza: we've had a combination of these to refer to the distribution release (`dver`, `distrover`, `release`) could you use `DISTRO_RELEASE` so that it is clear what the variable is about (even it if is longer)

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,62 @@
...
+            description: "The distribution codename like C7.2, U16.04, D8"
+
+        - string:
+            name: VERSION
+            description: "The release of that particular build : 1.0.0, 1.1.0, ..."
+
+        - string:
+            name: REPOSITORY
+            description: "URL of the package repository, like http://debian.mirrors.ovh.net/debian/"
```
alfredodeza: I am guessing that this is something that changes from distro to distro. Why do we need to pass in the full URL for a repository? Can't the job figure out which one to use and default to that unless we want one explicitly? 

And what would the URL be for an RPM distro?

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,62 @@
...
+          branches:
+            - ceph-builders
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
+          skip-tag: true
+          wipe-workspace: true
+
+    builders:
+      - shell: "sudo edeploy activate-pkgmngr; cd build; sudo make ceph-builders DIST=${DIST} DVER=${DVER} VERSION=${VERSION} VIRTUALIZED=ceph-builders.virt NO_MGNIDS=True NO_COMPRESSED_FILE=True REPOSITORY=${REPOSITORY} KEEP_PKGMNGR=True && DVER=${DVER} VERSION=${VERSION} upload.sh"
```
alfredodeza: this might be a bit better as a script in `ceph-build-os/build/build` so that it can be included here.

ceph-build-os/config/definitions/ceph-build-os.yml
ErwanAliasr1: Yes, that's distro specific but also mirror specific. The project cannot determine what is _your_ preferred mirror. That's why we let people defining some unless that's determined by debootstrap. I saw many failures by using the central repo so using a local one is better.

For an centos, repository looks like http://centos.mirrors.ovh.net/ftp.centos.org/7/os/x86_64/Packages/centos-release-7-2.1511.el7.centos.2.10.x86_64.rpm

ceph-build-os/config/definitions/ceph-build-os.yml
ErwanAliasr1: Doing that.

ceph-build-os/config/definitions/ceph-build-os.yml
ErwanAliasr1: Done.

ceph-build-os/config/definitions/ceph-build-os.yml
ErwanAliasr1: Done.

ceph-build-os/config/definitions/ceph-build-os.yml
alfredodeza: It would be nice not to require a user to specify this all the time though. I understand that there is a preference depending on location and that is just not possible to compute, but it should have some defaults to that we aren't required to pass this in.

Can you also add the RPM repo url as an example to the help string?

alfredodeza: Can the option be renamed to `REPO_URL` ?

ErwanAliasr1:
@alfredodeza if you are ok with it, can we merge it to get the new behavior for build the 2nd gen of images ?

alfredodeza:
you need a rebase here

ErwanAliasr1:
Rebased

andrewschoen:
test this please

ErwanAliasr1:
@andrewschoen @alfredodeza Tested successfully as per https://jenkins.ceph.com/job/ceph-build-os/65/console

ceph-build-os/build/build
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+set +x
+
+if [ -z $DIST ]; then
+    echo "build: No DIST found, exiting !"
+    exit 1
+fi
+
+if [ -z $DVER ]; then
```
alfredodeza: `DVER` is no longer used I think?

ceph-build-os/build/build
```diff
@@ -0,0 +1,41 @@
...
+if [ -z "$REPOSITORY" ]; then
+    case "$DIST" in
+        "xenial"|"trusty")
+            REPOSITORY="http://ubuntu.mirrors.ovh.net/ubuntu/"
+        ;;
+        "jessie")
+            REPOSITORY="http://debian.mirrors.ovh.net/debian/"
+        ;;
+        "centos")
+            REPOSITORY="http://centos.mirrors.ovh.net/ftp.centos.org/7/os/x86_64/Packages/centos-release-7-2.1511.el7.centos.2.10.x86_64.rpm"
```
alfredodeza: The configuration is allowing `REPO_URL` not `REPOSITORY`

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,79 @@
...
+        - string:
+            name: REPO_URL
+            description: "Optional URL of a package repository to override the default one like (http://debian.mirrors.ovh.net/debian/ or http://centos.mirrors.ovh.net/ftp.centos.org/7/os/x86_64/Packages/centos-release-7-2.1511.el7.centos.2.10.x86_64.rpm"
+
+    scm:
+      - ceph-build
+      - edeploy
+
+    builders:
+      - shell: "cd edeploy/build; DIST=${DIST} DVER=${DISTRO_RELEASE} VERSION=${VERSION} REPOSITORY=${REPO_URL} ../../ceph-build/ceph-build-os/build/build"
```
ErwanAliasr1: @alfredodeza This is were I do the binding between the parameters name you asked for and the variable names I need for edeploy. I prefer keeping the edeploy variable name fwhen calling build to keep the consistency .

alfredodeza: this is going to be a problem, because your errors are going to report about a variable that the user is not passing ever (and not seeing) because of the abstraction :(

ErwanAliasr1: fair enough @alfredodeza 
I've added a commit to insure we do the translation only when calling make. Therefore every variable & messages reported to the users are then consistent.

andrewschoen:
test this please

ErwanAliasr1:
@andrewschoen that code is running fine

ceph-build-os/build/build
```diff
@@ -0,0 +1,99 @@
...
+	   echo "upload: No IMAGE found in /var/lib/debootstrap/install/$DISTRO_RELEASE-$VERSION/, Exiting"
+	   exit 1
+	fi
+
+	IMAGE_NAME=$(basename `echo $IMAGE | sed -e "s/\.img\.qcow2//g"`)
+
+	echo "upload: About to upload $IMAGE as $IMAGE_NAME"
+
+	# If the same image already exists, let's delete it
+	openstack image list -f csv | grep -qw "$IMAGE_NAME" && echo "upload: Deleting remote image" && openstack image delete "$IMAGE_NAME"
```
alfredodeza: how is `openstack` being installed (or made available) ?

ceph-build-os/build/build
```diff
@@ -0,0 +1,99 @@
...
+        ;;
+        *)
+            echo "build: no known favorite repository for $DIST"
+        ;;
+    esac
+    echo "build: Using repository $REPO_URL"
+fi
+
+
+sudo edeploy activate-pkgmngr
```
alfredodeza: same with `edeploy` 

I'm not finding where these executables are getting installed

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,79 @@
...
+        - string:
+            name: REPO_URL
+            description: "Optional URL of a package repository to override the default one like (http://debian.mirrors.ovh.net/debian/ or http://centos.mirrors.ovh.net/ftp.centos.org/7/os/x86_64/Packages/centos-release-7-2.1511.el7.centos.2.10.x86_64.rpm"
+
+    scm:
+      - ceph-build
+      - edeploy
+
+    builders:
+      - shell: "cd edeploy/build; DIST=${DIST} DISTRO_RELEASE=${DISTRO_RELEASE} VERSION=${VERSION} REPO_URL=${REPO_URL} ../../ceph-build/ceph-build-os/build/build"
```
alfredodeza: oh, maybe this "build" script is what is installing these things? is it not using a virtualenv? that would make it hard to contain pollution of the VM otherwise

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    publishers:
+      - postbuildscript:
+          script-only-if-succeeded: True
+          script-only-if-failed: True
+          builders:
+            - shell: "sudo poweroff"
```
alfredodeza: it looks like this is not getting executed (is this really the behavior you want?):

    [PostBuildScript] - Execution post build scripts.
    [PostBuildScript] Build is not failure : do not execute script
    Finished: SUCCESS

ceph-build-os/build/build
ErwanAliasr1: This script is executed on a ceph-builders image which contains all the requirements for that task.
Ceph-builders images are both building ceph & edeploy capable.

ceph-build-os/config/definitions/ceph-build-os.yml
```diff
@@ -0,0 +1,79 @@
+- scm:
+    name: ceph-build
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph-build.git
```
alfredodeza: doesn't look like you need to clone ceph-build, are you using anything in ceph-build?

Anything in the current job (ceph-build-os) can be "included", so no need to clone it

ceph-build-os/config/definitions/ceph-build-os.yml
alfredodeza: So for example, instead of calling ` ../../ceph-build/ceph-build-os/build/build` you could include the `build` file  like we do here for example: https://github.com/ceph/ceph-build/blob/master/ceph-deploy-docs/config/definitions/ceph-deploy-docs.yml#L34

PR #394, opened 2016-05-23
ansible: Enforce rights on /home/jenkins-build

ErwanAliasr1:
On a fresh new image, the /home/jenkins-build relies to root.root making ccache
failing at creating a file here.

That means that anything trying to write a file in the home dir will fail.

This patch simply enfore the user/group of the home dir

### DIFF
--- a/ansible/slave.yml
+++ b/ansible/slave.yml
@@ -34,8 +34,8 @@
     - name: ensure the build dir exists
       file: path=/home/{{ jenkins_user }}/build state=directory owner={{ jenkins_user }}
 
-    - name: ensure the build dir has the right owner permissions
-      file: path=/home/{{ jenkins_user }}/build state=directory owner={{ jenkins_user }}
+    - name: ensure the home dir has the right owner permissions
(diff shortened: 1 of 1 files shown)


ansible/slave.yml
```diff
@@ -34,8 +34,8 @@
     - name: ensure the build dir exists
       file: path=/home/{{ jenkins_user }}/build state=directory owner={{ jenkins_user }}
 
-    - name: ensure the build dir has the right owner permissions
-      file: path=/home/{{ jenkins_user }}/build state=directory owner={{ jenkins_user }}
+    - name: ensure the home dir has the right owner permissions
+      file: path=/home/{{ jenkins_user }} state=directory owner={{ jenkins_user }} group={{ jenkins_user }} recursive=yes
```
alfredodeza: s/recursive/recurse/

And I am not 100% we are creating a `{{ jenkins_user }}` group

ErwanAliasr1: We do admin@trusty-huge--9b6e6c57-9d1f-411e-bb44-fd5c00cb7865:~$ grep jenkins /etc/group
jenkins-build:x:1001:

PR #403, opened 2016-06-06
teuthology-docs: Move package dependencies into the job definition

zmc:
Even though we can't actually *install* them, since we don't have tty-less sudo on ``docs.dhc``, it's useful to put jobs' dependencies in the job definition.


alfredodeza:
this looks good. I'll merge after a rebase :)

zmc:
@alfredodeza great, rebased

PR #404, opened 2016-06-07
Improving build script

ErwanAliasr1:
This PR is about making the build process as a script with enhanced debugging when something gets wrong.

It does reports 
- the deleted & leftover files
- the remaining processes
- detect the segfault & extract debug info with addr2line

This makes devs's life easier as it helps understanding the output of the build.

NOTE: THIS PR IS STILL TARGETING MY PERSONAL BRANCH AND NEEDS AN UPDATE BEFORE THE MERGE.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -1,3 +1,30 @@
+- scm:
+    name: ceph-build
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph-build.git
+          browser-url: https://github.com/ceph/ceph-build
+          branches:
+                  - evelu-build
+          timeout: 20
+          skip-tag: true
+          wipe-workspace: false
+          basedir: "ceph-build"
+
+- scm:
+    name: ceph
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph.git
+          branches:
+                - ${sha1}
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
+          skip-tag: true
+          wipe-workspace: true
+          basedir: "ceph"
+
 - job:
     name: ceph-pull-requests
     project-type: freestyle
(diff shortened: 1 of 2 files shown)


ErwanAliasr1:
Anyone ?

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -1,3 +1,30 @@
+- scm:
+    name: ceph-build
```
andrewschoen: You shouldn't need to clone ceph-build. Just use ``!include-raw`` with the shell builders below.

See: https://github.com/ceph/ceph-build/blob/master/ceph-build/config/definitions/ceph-build.yml#L56

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -35,22 +62,15 @@
...
-          browser: auto
-          timeout: 20
-          skip-tag: true
-          wipe-workspace: true
+      - ceph-build
+      - ceph
 
     builders:
-      - shell: "ccache -M 5G; ccache -z; export NPROC=$(nproc); timeout 7200 ./run-make-check.sh; ccache -s"
+      - shell: "cd ceph; ../ceph-build/ceph-pull-requests/build/build build"
```
andrewschoen: You can use ``!include_raw`` here.  See: https://github.com/ceph/ceph-build/blob/master/ceph-build/config/definitions/ceph-build.yml#L56

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -35,22 +62,15 @@
...
 
     publishers:
       - postbuildscript:
           script-only-if-succeeded: False
           script-only-if-failed: True
           builders:
-            - shell: "sudo reboot"
+            - shell: "cd ceph; ../ceph-build/ceph-pull-requests/build/build failure"
```
andrewschoen: You can use ``!include_raw`` here as well.

PR #419, opened 2016-07-21
Letsencrypt support for load balanced Shaman

alfredodeza:
Tricky because this deploys when Grafana gets deployed (shaman gets load balanced, grafana does not)

All with letsencrypt support


ansible/roles/nginx/tasks/letsencrypt.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: unlink nginx configs
+  file:
+    path: "/etc/nginx/sites-enabled/{{ item.app_name }}.conf"
+    state: "absent"
+  sudo: true
+  with_items: nginx_hosts
+
+- name: create temporary nginx config
+  template:
+    src: "../templates/nginx_tmp_site.conf"
```
andrewschoen: You don't need ``../templates/`` here.

ansible/roles/nginx/tasks/main.yml
```diff
@@ -37,57 +37,26 @@
 
 - name: create nginx site config
   template:
-    src: "../templates/nginx_balancer.conf"
-    dest: "/etc/nginx/sites-available/{{ item.name }}.conf"
+    src: "../templates/nginx_site.conf"
```
andrewschoen: ditto, don't need ``../templates/`` here.

PR #422, opened 2016-08-04
ceph-dev: provide the ability to choose which distros are used

andrewschoen:
This PR allows for the user to select which distros ceph will be built for. It also fixes a few things that were found while testing.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -2,7 +2,8 @@
 
 set -ex
 
-VENV="/tmp/jenkins-venv/bin"
+TEMPVENV=$(mktemp -td venv.XXXXXXXXXX)
+VENV="$TEMPVENV/bin"
 
 install_python_packages () {
     # Use this function to create a virtualenv and install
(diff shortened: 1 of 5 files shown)


scripts/build_utils.sh
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-VENV="$WORKSPACE/venv/bin"
+VENV="/tmp/jenkins-venv/bin"
```
ktdreyer: Can we use `TEMPVENV=$(mktemp -td venv.XXXXXXXXXX)` instead of hard-coding `/tmp/jenkins-venv` ? hard-coded `/tmp` paths make Shadowman weep

andrewschoen: You are right, I should have been more considerate of Shadowman's feelings. Will fix.

alfredodeza: This is not only important for avoiding hard-coding, but it is entirely possible that we get into trouble on a subsequent build attempting to use the same venv

PR #423, opened 2016-08-05
use the value of VENV from build_utils.sh everywhere

andrewschoen:
We need to do this now as the virtualenv is being created in a temp directory and not a standard path.


andrewschoen:
I've updated this to include a commit that will ping chacra after all binaries are done uploading to start the repo creation.

PR #424, opened 2016-08-08
scripts: disable chacractl ssl verification temporarily

alfredodeza:


ktdreyer:
that's what it does, but why?

ktdreyer:
why

andrewschoen:
@ktdreyer we were having trouble with letsencrypt so we thought it'd be important to also support self-signed certificates.

ktdreyer:
Thank you!

PR #433, opened 2016-08-17
calamari-server-trusty: Install multitee before we use it

b-ranto:
The calamari-server-trusty build fails because there is no multitee available. Fix this by installing it before running the multitee'd salt command.


b-ranto:
This can be seen for instance here:

https://jenkins.ceph.com/job/calamari-server-trusty/lastFailedBuild/consoleText

Look for the line:

/tmp/hudson1094321908020410961.sh: line 14: multitee: command not found

PR #435, opened 2016-08-22
ceph-dev-setup: produce the tarball with cmake if make-dist exists

alfredodeza:
Just working for the tarball (ceph-dev-setup) process. The actual binary builds still need work.


alfredodeza:
@alimaredia you will probably be interested in this PR.

PR #436, opened 2016-08-23
calamari-server-trusty: Upload the packages to chacra

b-ranto:

### DIFF
--- a/calamari-server-trusty/config/definitions/calamari-server-trusty.yml
+++ b/calamari-server-trusty/config/definitions/calamari-server-trusty.yml
@@ -2,7 +2,10 @@
     block-downstream: false
     block-upstream: false
     builders:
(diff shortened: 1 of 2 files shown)


calamari-server-trusty/config/definitions/calamari-server-trusty.yml
```diff
@@ -2,7 +2,10 @@
     block-downstream: false
     block-upstream: false
     builders:
-    - shell: !include-raw ../../build/build
+    - shell:
+      !include-raw:
```
alfredodeza: this needs to be outdented 2 spaces it seems. Like:

    - shell:
           !include-raw:

b-ranto: That seemed to have helped, thanks!

PR #444, opened 2016-08-29
support for concurrent ceph-dev-setup and ceph-dev-build jobs

andrewschoen:


alfredodeza:
would you mind splitting 834792d one for each file?

PR #449, opened 2016-08-30
Job cleanup

alfredodeza:
There are old jobs that are not only unused but there is no plan to use them at all.


alfredodeza:
This will never pass as it is removing stuff that the PR-syntax-check job will want to check

PR #454, opened 2016-09-01
teuthology PRs: require a huge Trusty host

zmc:


teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
```diff
@@ -70,3 +70,6 @@
           !include-raw:
             - ../../../scripts/build_utils.sh
             - ../../build/build
+
+    publishers:
+      - github-notifier
```
alfredodeza: This is already taken care of by `github-pull-request`. None of the current jobs use this because it causes a conflict since it is two different plugins.

Are you seeing a problem with the current approach?

teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
```diff
@@ -1,6 +1,6 @@
 - job:
     name: teuthology-pull-requests
-    node: trusty && amd64 && small
+    node: trusty
```
alfredodeza: this change sounds reasonable to me

teuthology-pull-requests/config/definitions/teuthology-pull-requests.yml
zmc: It may be that I'm doing it wrong, but after I enabled "Require status checks..." [here](https://github.com/ceph/teuthology/settings/branches/master), the status wasn't being updated.

Now I see that there are two actual status checks (I swear there was only the one before); perhaps if that were checked we wouldn't need this second plugin?

zmc: (Testing that theory now: https://github.com/ceph/teuthology/pull/948)

alfredodeza: For pull requests to display something other than 'default' this needs to be added to the `github-pull-request` section:

    status-context: "custom-text-here"

And I don't see that has ever been set there to "teuthology-pull-requests". Is it possible that is coming from Travis?

zmc: Ah, it is possible that came from Travis, when we used it. Still testing...

zmc: Well, that's very confusing: it works now. I'm not sure why it didn't before, but unless it breaks again I won't worry about it.

PR #457, opened 2016-09-01
adds a ceph-dev-trigger job

andrewschoen:
This new job will respond to pushes to the ceph repo and trigger the ceph-dev job.

### DIFF
--- a/ceph-dev-build/config/definitions/ceph-dev-build.yml
+++ b/ceph-dev-build/config/definitions/ceph-dev-build.yml
@@ -1,6 +1,6 @@
 - job:
     name: ceph-dev-build
-    node: trusty || centos7 || xenial || master
+    node: huge_xenial_builder
     project-type: matrix
     defaults: global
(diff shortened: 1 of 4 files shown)


ceph-dev-build/config/definitions/ceph-dev-build.yml
```diff
@@ -1,6 +1,6 @@
 - job:
     name: ceph-dev-build
-    node: trusty || centos7 || xenial || master
+    node: huge_xenial_builder
```
alfredodeza: we will need to make this generic, like `executor` so that we can fire more of these light weight executor nodes

ceph-dev-setup/build/build
```diff
@@ -191,3 +193,7 @@ mv release/version dist/.
 cat > dist/sha1 << EOF
 SHA1=${GIT_COMMIT}
 EOF
+
+cat > dist/branch << EOF
```
alfredodeza: excellent

PR #466, opened 2016-09-09
Filter slashes out of branches

alfredodeza:
This was already done in the setup job, but the variable doesn't percolate when modified it seems, so we need to do this filtering everywhere

### DIFF
--- a/ceph-dev-build/build/build_deb
+++ b/ceph-dev-build/build/build_deb
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -ex
 
+BRANCH=`branch_slash_filter $BRANCH`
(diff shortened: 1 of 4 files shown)


andrewschoen:
I think what's happening here is that we populate $BRANCH with the properties file from ceph-dev-setup and *then* the parameters from ceph-dev are injected which overwrites $BRANCH

alfredodeza:
lets merge this as-is for now while I work on porting the changes to the function in build_utils.sh

alfredodeza:
Went ahead and implemented the function in `build_utils.sh` So this is good to get reviewed/merged

PR #469, opened 2016-09-09
jenkins-job-builder: should not use the master node

andrewschoen:
master is not setup for building anything and does not have virtualenv
installed which this job needs. In general, we don't want jobs that do
any "building' to use master.

### DIFF
--- a/jenkins-job-builder/config/definitions/jjb.yml
+++ b/jenkins-job-builder/config/definitions/jjb.yml
@@ -1,6 +1,6 @@
 - job:
     name: jenkins-job-builder
-    node: (small && trusty) || master
+    node: small
     project-type: freestyle
     defaults: global
     display-name: 'Jenkins Job Builder'


jenkins-job-builder/config/definitions/jjb.yml
```diff
@@ -1,6 +1,6 @@
 - job:
     name: jenkins-job-builder
-    node: (small && trusty) || master
+    node: small && trusty
```
alfredodeza: can we just use `small` here? I don't think we need `trusty`

andrewschoen: that would potentially give us wheezy nodes though, unless we changed mita's config to remove wheezy

alfredodeza: we can't use wheezy? (they are currently disabled, but curious)

andrewschoen: we can, they just take forever to provision. did you disable wheezy manually? I checked here and they still look enabled https://github.com/ceph/mita/blob/master/deploy/playbooks/roles/common/templates/prod_nodes.py.j2#L26

alfredodeza: they are manually disabled. And you are right, they take for ever. We should just get them removed or commented out "officially"

alfredodeza:
this looks good to me

PR #473, opened 2016-09-12
ceph-dev-build: further massage ceph.spec to insist on using sha1 in version name

alfredodeza:


alfredodeza:
already making the kraken branch pass in production

PR #477, opened 2016-09-13
ansible: set utf-8 locale for the build user

alfredodeza:


ktdreyer:
The description tells us what it does, but **why** ?

why

b-ranto:
@ktdreyer We were running into issues when trying to build calamari-client in jenkins. There were some utf8-named files and jenkins fails to remove them if the locale is not in utf8 (jvm apparently works in ascii only mode and is unable to recognize the characters in the file names). We were hitting the following issue and this is a workaround:

https://issues.jenkins-ci.org/browse/JENKINS-33478

PR #478, opened 2016-09-13
A playbook to configure static jenkins slaves

andrewschoen:
I used this playbook to add smithi118 - smith128 (except smithi126) to jenkins. 

I decided to go with a different playbook than slave.yml because I wasn't going to use this with prado and some of the paths to templates and files were different when running this outside of prado.


alfredodeza:
Could you add what pr #477 does as well? to prevent JavaIO errors because of encodings being POSIX

andrewschoen:
@alfredodeza I've updated this to include the work from PR #477

PR #484, opened 2016-09-16
set BUILD_HOME to the right path

alfredodeza:
It was making `.pbuilderrc` (the configuration **file**) a directory. This makes it right


ktdreyer:
For posterity: I think this was to handle a change between Trusty's pbuilder version and Xenial's pbuilder version, which set $HOME differently.

ktdreyer:
Whoops, that's already explained in better detail in https://github.com/ceph/ceph-build/pull/479 :)

PR #485, opened 2016-09-16
cmake: modified several scripts for cmake support

alimaredia:
This is a first crack at release builds supporting cmake and autotools.

To Do: modify ceph/ceph.spec.in to support CEPH_EXTRA_CMAKE_ARGS

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -81,6 +81,7 @@
         - autoconf
         - redhat-lsb-core
         - automake
+        - cmake
         - binutils
         - bison
         - flex
--- a/ceph-dev-setup/build/build
+++ b/ceph-dev-setup/build/build
@@ -43,25 +43,43 @@ echo "Running submodule update ..."
 git submodule update --init
 
 # Flavor Builds support
-# TODO: enable CMAKE
-#export CEPH_EXTRA_CMAKE_ARGS
 
-FLAVOR_FLAGS="--without-cryptopp"
+#cryptopp is not default for CMake
+AUTOCONF_FLAVOR_FLAGS="--without-cryptopp"
 
 if [ "${FLAVOR}" == "notcmalloc" ]
 then
-    echo "Detected notcmalloc flavor: will use flag: --without-tcmalloc"
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --without-tcmalloc"
-    #export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=libc"
+    echo "Detected notcmalloc flavor: will use flag: --without-tcmalloc or -DALLOCATOR=libc"
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --without-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=libc"
 else
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --with-tcmalloc"
-    #export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --with-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
 fi
 
-if [ "${FLAVOR}" == "blkin" ]
-then
-    echo "Detected bklin flavor: will use flag: --with-blkin"
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --with-blkin"
+
+# make-dist does not exist for autotools builds
+if [ ! -x make-dist ]
+    echo "Using autotools build system"
+    echo "Running autogen.sh ..."
+    ./autogen.sh
+    echo "Running configure ..."
+    ./configure \
+      --disable-option-checking \
+      '--prefix=/usr' \
+      '--sbindir=/sbin' \
+      '--localstatedir=/var' \
+      '--sysconfdir=/etc' \
+      '--with-debug' \
+      '--with-nss' \
+      '--with-radosgw' \
+      '--disable-static' \
+      '--without-lttng' \
+      $AUTOCONF_FLAVOR_FLAGS \
+      'CFLAGS= -Wno-unused-parameter' \
+      'CXXFLAGS=  -Wno-unused-parameter' \
+      --cache-file=/dev/null \
+      --srcdir=.
 fi
 
 mkdir -p release
@@ -104,38 +122,18 @@ else
 
     if [ -x make-dist ]
     then
-      ./make-dist $cephver
-      vers=`ls ceph-*.tar.bz2 | cut -c 6- | sed 's/.tar.bz2//'`
-      extension="tar.bz2"
-      extract_flags="jxf"
-      compress_flags="jcf"
+        ./make-dist $cephver
+        vers=`ls ceph-*.tar.bz2 | cut -c 6- | sed 's/.tar.bz2//'`
+        extension="tar.bz2"
+        extract_flags="jxf"
+        compress_flags="jcf"
     else
-      echo "Running autogen.sh ..."
-      ./autogen.sh
-      echo "Running configure ..."
-      ./configure \
-        --disable-option-checking \
-        '--prefix=/usr' \
-        '--sbindir=/sbin' \
-        '--localstatedir=/var' \
-        '--sysconfdir=/etc' \
-        '--with-debug' \
-        '--with-nss' \
-        '--with-radosgw' \
-        '--disable-static' \
-        '--without-lttng' \
-        $FLAVOR_FLAGS \
-        'CFLAGS= -Wno-unused-parameter' \
-        'CXXFLAGS=  -Wno-unused-parameter' \
-        --cache-file=/dev/null \
-        --srcdir=.
-
-      make dist
-      make dist-bzip2
-      vers=`ls ceph-*.tar.gz | cut -c 6- | sed 's/.tar.gz//'`
-      extension="tar.gz"
-      extract_flags="zxf"
-      compress_flags="zcf"
+        make dist
+        make dist-bzip2
+        vers=`ls ceph-*.tar.gz | cut -c 6- | sed 's/.tar.gz//'`
+        extension="tar.gz"
+        extract_flags="zxf"
+        compress_flags="zcf"
     fi
 
     echo tarball vers $vers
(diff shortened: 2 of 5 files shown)


alimaredia:
From what I can tell I don't need to check the version number to make sure autotools or cmake is enabled here (https://github.com/ceph/ceph-build/blob/master/ceph-dev-setup/build/build#L113) because make-dist is in repo's that work for cmake?

ansible/examples/slave.yml
```diff
@@ -81,6 +81,7 @@
         - autoconf
         - redhat-lsb-core
         - automake
+        - cmake
```
andrewschoen: Could you also add this to ``ansible/examples/slave_static.yml``

ceph-dev-setup/build/build
```diff
@@ -43,25 +43,44 @@ echo "Running submodule update ..."
 git submodule update --init
 
 # Flavor Builds support
-# TODO: enable CMAKE
-#export CEPH_EXTRA_CMAKE_ARGS
 
-FLAVOR_FLAGS="--without-cryptopp"
```
andrewschoen: $FLAVOR_FLAGS is still being used later in the script at https://github.com/ceph/ceph-build/blob/wip-cmake/ceph-dev-setup/build/build#L146

ansible/examples/slave.yml
alimaredia: done

ceph-dev-setup/build/build
alimaredia: done

ceph-dev-setup/build/build
```diff
@@ -43,25 +43,44 @@ echo "Running submodule update ..."
...
 fi
 
-if [ "${FLAVOR}" == "blkin" ]
-then
-    echo "Detected bklin flavor: will use flag: --with-blkin"
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --with-blkin"
+ceph_release_num=`git describe | sed 's/^v\([[:alnum:]]\+\)\..*$/\1/'`
+
+#get an integer for the ceph release version number to check for building with autotools
+if [ "$ceph_release_num" -le 10 ]; then
```
tchaikov: why shall we run configure here? we repeat this step at https://github.com/ceph/ceph-build/blob/wip-cmake/ceph-dev-setup/build/build#L132 for preparing the dist tarball. and also, we'd better be consistent on when to use cmake: we use `[ -x make-dist ]` at https://github.com/ceph/ceph-build/blob/wip-cmake/ceph-dev-setup/build/build#L124.

ceph-dev/config/definitions/ceph-dev.yml
```diff
@@ -56,9 +56,8 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           choices:
             - default
             - notcmalloc
-            - blkin
           default: "default"
-          description: "Type of Ceph build, choices are: notcmalloc, and blkin. Defaults to: 'default'"
+          description: "Type of Ceph build, choices are: notcmalloc. Defaults to: 'default'"
```
tchaikov: better off putting

> choices are: notcmalloc, default (i.e. with tcmalloc). Defaults to: 'default'.

ceph-dev-setup/build/build
alfredodeza: I agree with using `-x make-dist` is what we have elsewhere. Unless there is an inconsistency we should be aware of?

This whole block does repeat itself.

ceph-setup/build/build
```diff
@@ -38,24 +38,44 @@ fi
...
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=libc"
+else
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --with-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
+fi
+
+#get an integer for the ceph release version number to check for building with autotools
+ceph_release_num=`git describe | sed 's/^v\([[:alnum:]]\+\)\..*$/\1/'`
+
+if [ "$ceph_release_num" -le 10 ]; then
```
alfredodeza: same comment as above, I think you can detect this easily with `-x make-dist`

ceph-dev/config/definitions/ceph-dev.yml
alimaredia: changed

ceph-dev-setup/build/build
```diff
@@ -43,25 +43,45 @@ echo "Running submodule update ..."
...
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
 fi
 
-if [ "${FLAVOR}" == "blkin" ]
-then
-    echo "Detected bklin flavor: will use flag: --with-blkin"
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --with-blkin"
+
+# make-dist was renamed in conjuntion with cmake cutover
+if [ -x make-dist ]
```
alimaredia: @alfredodeza @tchaikov does this conditional look better? The exact same on is in ceph-setup/build/build

ceph-dev-setup/build/build
```diff
@@ -43,25 +43,45 @@ echo "Running submodule update ..."
...
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --without-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=libc"
 else
-    FLAVOR_FLAGS="$FLAVOR_FLAGS --with-tcmalloc"
-    #export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --with-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
 fi
 
-if [ "${FLAVOR}" == "blkin" ]
```
alimaredia: @jdurgin just want to make sure this is ok to take out. I submitted a PR to remove it from autobuild-ceph too (https://github.com/ceph/autobuild-ceph/pull/82).

ceph-dev-setup/build/build
alfredodeza: yep, this is correct. But you are still missing the actual call to `make-dist` right?

alimaredia: @alfredodeza yes, the other build script has one, this one doesn't. Looking at both of the scripts, is there any reason they shouldn't be identical or possibly symlinked?

ceph-dev-setup/build/build
jdurgin: seems fine to remove the flavor. We'll probably want to include blkin by default rather than in a separate flavor once https://github.com/ceph/ceph/pull/11053 merges

ceph-setup/build/build
```diff
@@ -92,10 +114,22 @@ else
     echo building tarball
     rm ceph-*.tar.gz || true
     rm ceph-*.tar.bz2 || true
-    make dist
-    make dist-bzip2
+    if [ -x make-dist ]
```
alimaredia: @alfredodeza make-dist has been added

ceph-dev-setup/build/build
```diff
@@ -104,38 +124,18 @@ else
...
-        --srcdir=.
-
-      make dist
-      make dist-bzip2
-      vers=`ls ceph-*.tar.gz | cut -c 6- | sed 's/.tar.gz//'`
-      extension="tar.gz"
-      extract_flags="zxf"
-      compress_flags="zcf"
+        make dist
+        make dist-bzip2
```
tchaikov: is there any reason we want to do the "make dist" twice? maybe we can just always release the .bz2 tarball. and share the following code in both branches (cmake and autotools)
```sh
  vers=`ls ceph-*.tar.bz2 | cut -c 6- | sed 's/.tar.bz2//'`
  extension="tar.bz2"
  extract_flags="jxf"
```

and maybe we can dispense with `extension` and `extract_flags`, as they are fixed.
and we don't use  `compress_flags` anywhere, so it can be dropped?

ceph-dev-setup/build/build
tchaikov: the most notable difference in the output of `diff ceph-setup/build/build ceph-dev-setup/build/build`

```diff
2a3,6
> BRANCH=`branch_slash_filter $BRANCH`
> 
> # split on '/' to get just 'wip-mybranch' when input is like: origin/wip-mybranch
> 
150a198,201
> 
> cat > dist/branch << EOF
> BRANCH=${BRANCH}
> EOF
```

maybe @alfredodeza could shed some light on it.

tchaikov:
needs rebase, modulo the nits, lgtm.

ceph-dev-setup/build/build
alfredodeza: those should be the same. We don't support branches with forward slashes in them because we use a binary store in an HTTP API. Since the "branch" is part of the URL it cannot have a slash in it

ceph-setup/build/build
```diff
@@ -38,24 +38,46 @@ fi
...
+    echo "Detected notcmalloc flavor: will use flag: --without-tcmalloc or -DALLOCATOR=libc"
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --without-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=libc"
+else
+    AUTOCONF_FLAVOR_FLAGS="$AUTOCONF_FLAVOR_FLAGS --with-tcmalloc"
+    export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
+fi
+
+# make-dist was renamed in conjuntion with cmake cutover
+if [ -x make-dist ]
```
alfredodeza: if `make-dist` is going to be called later, no need to do the `if/else` here. This can just check if `make-dist` does *not* exist so that it can just run `autogen.sh`

alimaredia: done

ceph-dev/config/definitions/ceph-dev.yml
```diff
@@ -56,9 +56,8 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           choices:
             - default
             - notcmalloc
-            - blkin
           default: "default"
-          description: "Type of Ceph build, choices are: notcmalloc, and blkin. Defaults to: 'default'"
+          description: "Type of Ceph build, choices are: notcmalloc, default (i.e. with tcmalloc). Defaults to: 'default'".
```
alfredodeza: this is breaking the build. You have a `.` outside of the string

PR #486, opened 2016-09-20
allow network access in pbuilder builds

alfredodeza:
Because ceph components may want to install things that are cached and buggy in certain environments. Hi Xenial. Hello there pip 8.1.1. How are you urllib3


ktdreyer:
For posterity, one example of something that installed over the network: http://tracker.ceph.com/issues/15260

PR #495, opened 2016-09-29
Add template job and script to generate new build jobs

b-ranto:
The commit adds three new template jobs and create_new_jobs_layout.sh
script that will use these templates and will help generate the new
build jobs layout for a new upstream project.

Signed-off-by: Boris Ranto branto@redhat.com

btw: There is also an upstream video tutorial on how to use this:

https://asciinema.org/a/07u4onkwuymw5k2k7061upslh

### DIFF
--- a/detemplate.sh
+++ b/detemplate.sh
@@ -1,0 +1,30 @@
+#! /usr/bin/env bash
+
+set -e
+
+echo "Enter project name"
+read PROJECT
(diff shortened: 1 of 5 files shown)


b-ranto:
@ktdreyer @alfredodeza Please review and discuss the suggested layout --  I've tried to create a fairly simple and generic build job layout, maybe we need something more complex.

PR #500, opened 2016-10-03
ceph-dev-build: split setup into setup_deb and setup_rpm

andrewschoen:
This allows use to post earlier to shaman while still having access to all the distro information needed when making that update.


andrewschoen:
test this please

PR #516, opened 2016-10-10
ceph-dev-trigger: enable notcmalloc flavor

alfredodeza:


andrewschoen:
This looks good but I'm hesitant to merge until we've addressed storage on our chacra nodes. This will essentially double the amount of repos we're producing.

PR #518, opened 2016-10-11
ansible: use pgp.mit.edu to fetch the jessie release key

alfredodeza:
This would fix having to host this key in download.ceph.com

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -141,10 +141,10 @@
       when: ansible_pkg_mgr  == "apt"
       apt_key: id=C857C906 url=https://ftp-master.debian.org/keys/archive-key-8-security.asc keyring=/etc/apt/trusted.gpg state=present
 
-    - name: Add the Debian Jessie Stable Key
+    - name: Add the Debian Jessie Stable Release Key
       sudo: yes
       when: ansible_pkg_mgr  == "apt"
-      apt_key: id=518E17E1 url=http://download.ceph.com/keys/jessie-stable-release.asc keyring=/etc/apt/trusted.gpg state=present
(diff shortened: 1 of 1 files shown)


ktdreyer:
Oh man, yeah, we don't want to have responsibility for hosting Debian's keys.

Can you use the longer key ID, too? Those short key IDs are becoming [easy to brute force](http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html)

ktdreyer:
Want to specify the URL for safety? https://ftp-master.debian.org/keys/archive-key-8.asc

ktdreyer:
:cry: I'm wrong, we already import `archive-key-8.asc`, above. The "Debian Jessie Stable Key" is totally different

ansible/examples/slave.yml
```diff
@@ -141,10 +141,10 @@
       when: ansible_pkg_mgr  == "apt"
       apt_key: id=C857C906 url=https://ftp-master.debian.org/keys/archive-key-8-security.asc keyring=/etc/apt/trusted.gpg state=present
 
-    - name: Add the Debian Jessie Stable Key
+    - name: Add the Debian Jessie Stable Release Key
       sudo: yes
       when: ansible_pkg_mgr  == "apt"
-      apt_key: id=518E17E1 url=http://download.ceph.com/keys/jessie-stable-release.asc keyring=/etc/apt/trusted.gpg state=present
+      apt_key: keyserver=pgp.mit.edu id=518E17E1 state=present
```
ktdreyer: `id=75DDC3C4A499F1A18CB5F3C8CBF8D6FD518E17E1`

PR #521, opened 2016-10-21
release_tarball: use make-dist if available

tchaikov:
this enables us to prepare release tarball for versions without
autotools support, where we use make-dist to create the release tarball
instead.


liewegas:
lgtm

PR #522, opened 2016-10-25
kernel: Add initial implementation

b-ranto:
Add initial implementation for kernel jenkins builds. We build and prep
kernel in one phase with a single 'make' target and the Makefile does
not support creating just source rpm/deb files so we need to do all the
work in a single job.


b-ranto:
@andrewschoen ah, yeah, you are right, I have completely missed that when doing the commit, it should be fixed, now.

PR #527, opened 2016-10-26
kernel: various fixes

b-ranto:
- rename ARCH to DEB_ARCH to avoid overriding kernel ARCH variable in
  Makefile
- rename SHA1 to GIT_COMMIT, we do not set SHA1 anywhere
- download the kernel config files, they are not easily available
  locally
- deb files are stored in ../, clean ../, clean-up first to upload only
  new ../*.deb files
- rpm files are stored in ~/rpmbuild, clean-up first to upload only new
  rpm files
- remove precise from default distros as we do not seem to have any huge
  precise machines lying around at the moment


b-ranto:
FYI: If I did not miss anything, this make kernel buildable in Jenkins.

kernel/config/definitions/kernel.yml
```diff
@@ -14,7 +14,7 @@
       - string:
           name: DISTROS
           description: "A list of distros to build for. Available options are: xenial, centos7, centos6, trusty, precise, wheezy, and jessie"
-          default: "centos7 precise xenial"
+          default: "centos7 xenial"
```
andrewschoen: Can we add trusty here as well? For repos that teuthology might use for testing we've been building for centos7, xenial and trusty.

b-ranto: Sure, I've added trusty. I've removed precise from defaults just because there does not seem to be any nodes that would be providing it at the moment -- at least I did not get any in ~15 minutes or so before I cancelled the job.

kernel/build/build_deb
```diff
@@ -88,11 +88,17 @@ echo "*****"
 # Clean the git repo
 git clean -fxd
 
+# Install build dependencies
+sudo apt-get install -y bc
+
 # Prep the config
 echo "Preparing the kernel config"
-cp "$(dirname $0)/kernel-config-deb.basic" .config
+wget https://raw.githubusercontent.com/ceph/ceph-build/master/kernel/build/kernel-config-deb.basic -O .config
```
andrewschoen: Instead of using wget here you can use multiple ``scm`` entires in the config to have jenkins also clone ceph-build as well as ceph-client. I think that's the approach we should take here.

This job is a good example of that: https://github.com/ceph/ceph-build/blob/master/shaman-pull-requests/config/definitions/shaman-pull-requests.yml#L1

Then in the build script for that job you can see how I use different paths for ceph-build vs shaman.

https://github.com/ceph/ceph-build/blob/master/shaman-pull-requests/build/build#L9

b-ranto: Actually, I was hoping I could use this plugin

https://wiki.jenkins-ci.org/display/JENKINS/Config+File+Provider+Plugin

which seems to be designed to cover exactly this kind of use case but I'm not familiar with the yaml syntax for it and it does not seem to be enabled in our Jenkins instance at the moment. I've seen the plugin being available in the UI but I did not change anything there since I didn't want to break anything -- I'm not sure about what we use underneath the Jenkins to generate these things.  Cloning the entire ceph-build repo just for some config files seems like a bit of an overkill to me.

andrewschoen: Looks like the jjb for that plugin is here: http://docs.openstack.org/infra/jenkins-job-builder/builders.html#builders.config-file-provider

That approach would require manually entering that config into the jenkins UI though, unless I'm missing something. I'd much rather have that in source control then manually imported into the jenkins UI.

ktdreyer: It does seem to be a problem we've hit a couple times - "how do I quickly get additional file(s) into the workspace for Jenkins".

Unfortunately I think cloning the ceph-build.git repository is the least-worst option :cry:

b-ranto: Well, there is one other way of achieving this and it is as simple as making the config file a shell script, i.e. adding
`cat << EOF > file` 
at the beginning of the file,
`EOF`
at the end of the file and calling the script in the builders.

In fact, in my case, I think this might be the best way to achieve this. It adds only a minor overhead (few bytes for cat and EOF bits), it is tracked in scm and you can modify the config files quite easily in the Jenkins UI if needed.

b-ranto:
I've reworked the commit to use simple `cat` shell scripts that generate the config files -- it is fairly simple to use, just check the commit diff.

alfredodeza:
@b-ranto that was clever :)

PR #529, opened 2016-11-01
Add template job and script to generate new build jobs

b-ranto:
The commit adds a new template job and detemplate.sh script that will
use the template and will help new users generate a new build job for an
upstream project.

I've reworked the template job so that we get a much simpler template layout. A lot of time, we can't get the source package in an easy way (take e.g. kernel builds or calamari debian builds) so we convert the sources directly to the packages which we want to upload. This template job still encourages the use of mock and pbuilders and makes it possibly to directly build the binaries without the need for source rpms/debs.

### DIFF
--- a/detemplate.sh
+++ b/detemplate.sh
@@ -1,0 +1,30 @@
+#! /usr/bin/env bash
+
+set -e
+
+echo "Enter project name"
+read PROJECT
+echo "Enter github name as in github.com/ceph/<github_name>"
+read GITHUB
+
+echo "You entered:"
+echo "PROJECT=$PROJECT"
+echo "GITHUB=$GITHUB"
+
+if test -e ${PROJECT}; then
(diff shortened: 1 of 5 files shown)


b-ranto:
An updated screencast:

https://asciinema.org/a/ae66ivj75kwdvmw8mro39ub6a

ilc:
As a contributor.  How can I pre-test my changes to make sure that they work?

b-ranto:
Good question, currently, we prepare the script to the point where we think it is functional (no testing necessary, it only has to pass the basic automated sanity check), push the PR so that we get the initial layout in the Jenkins web UI, tune it up there and then we create the second PR with the fixes.

djgalloway:
@b-ranto I'm going through and cleaning up old PRs.  I don't understand what this one's for.  Is it still needed?

PR #530, opened 2016-11-01
calamari: Simplify to add debian build

b-ranto:
The debian build does not currently support creating source packages so
we need to merge the build and setup jobs into a single job. This patch
also simplifies various other bits and adds the support for the debian
build.


alfredodeza:
@b-ranto a single commit that makes changes to multiple files has caused us pain before :( The removal of the files is fine, but could you make the additions be as granular as you think it is reasonable?

b-ranto:
@alfredodeza this is pretty much a complete rewrite, what I did is that I removed the three old jobs and created a new one based on the new template that I've created. I do not expect this to work right away, I just want to get the new job structure into the jenkins so that I could test it all out and update the template job accordingly.

I can make this into probably two commits (removal of the old jobs and addition of the new job?) if that helps in any way.

PR #540, opened 2016-11-14
Various tweaks for the kernel job

idryomov:


idryomov:
Jenkins check failure looks unrelated.

kernel/build/build_deb
```diff
@@ -1,11 +1,6 @@
 #!/bin/bash
 set -ex
 
-# Only do actual work when we are a DEB distro
```
alfredodeza: why remove this? it will execute if this is running on an RPM box

kernel/build/build_rpm
```diff
@@ -1,11 +1,6 @@
 #!/bin/bash
 set -ex
 
-# Only do actual work when we are an RPM distro
```
alfredodeza: same here

idryomov: @alfredodeza Oh, should have been more clear - because we already have validate_{deb,rpm} in the definition.

alfredodeza:
jenkins test this please

PR #541, opened 2016-11-15
ceph{-dev}-{setup,build}: pass CMAKE_EXTRA_CONFIG_ARGS

dmick:
Inject another file, this one a 'catchall' of other environment variables
that -setup may need to pass to -build.  At the moment, that contains
CEPH_EXTRA_CMAKE_ARGS, but there doesn't seem to be much point in creating
one file per variable.

### DIFF
--- a/ceph-build/config/definitions/ceph-build.yml
+++ b/ceph-build/config/definitions/ceph-build.yml
@@ -57,6 +57,8 @@
           which-build: multijob-build
       - inject:
           properties-file: ${WORKSPACE}/dist/sha1
+      - inject:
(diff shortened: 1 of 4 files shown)


andrewschoen:
I agree there is no need to have individual files per environment variable. This PR looks correct, but I guess I'm missing the intent here. Will CMAKE_EXTRA_CMAKE_ARGS eventually be used in the build scripts?

liewegas:
@andrewschoen see http://tracker.ceph.com/issues/17869.  the env var is consumed by ceph.spec.in and debian/rules.

PR #542, opened 2016-11-15
kernel: repo-extra json with kernelrelease

idryomov:

### DIFF
--- a/kernel/build/build_rpm
+++ b/kernel/build/build_rpm
@@ -84,6 +84,19 @@ if [ "$THROWAWAY" = false ] ; then
     # push binaries to chacra
     find ~/rpmbuild/ | egrep '\.rpm$' | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpoint}/$ARCH/
 
+    # write json file with build info
+    cat > $WORKSPACE/repo-extra.json << EOF
+{
+    "version":"$kernelrelease",
+    "build_url":"$BUILD_URL",
+    "root_build_cause":"$ROOT_BUILD_CAUSE",
+    "node_name":"$NODE_NAME",
+    "job_name":"$JOB_NAME"
+}
+EOF
+    # post the json to repo-extra json to chacra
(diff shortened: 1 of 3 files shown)


idryomov:
This puts kernelrelease string into version, since it's not necessarily a package manager version.  teuthology only cares about package_manager_version right now though - @zmc what do you think about a shaman-only builder.scmversion or so?

kernel/build/build_rpm
```diff
@@ -84,6 +84,19 @@ if [ "$THROWAWAY" = false ] ; then
     # push binaries to chacra
     find ~/rpmbuild/ | egrep '\.rpm$' | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpoint}/$ARCH/
 
+    # write json file with build info
+    cat > $WORKSPACE/repo-extra.json << EOF
+{
+    "version":"$kernelrelease",
```
andrewschoen: You'll also need a ``package_manager_version`` field as that's what teuthology will use. The value there should match the version included in the binary name.

idryomov: @andrewschoen Currently we have builder.version property that corresponds to extra:package_manager_version. @zmc and I agreed to add another property, builder.scm_version, that would correspond to extra:version, to be used by the kernel task. builder.scm_version can either be ShamanProject-specific or just return builder.version in the GitbuilderProject case.

The rationale here is that the kernel task has always been special and kernel gitbuilder /version file never contained an actual package version.

andrewschoen: @idryomov that's totally fine with me if it works for @zmc

zmc: @idryomov @andrewschoen I'm a bit confused - we do still need `package_manager_version` - but this PR doesn't contain it. If we also need `version` (and it sounds like we do), we can use that too.

idryomov: @zmc Your wip-shaman-kernel branch uses `builder.version` to assign to `need_version[role]`.  That should become `builder.scm_version`.

Unless packaging.py *requires* `package_manager_version` - which I think it doesn't - we shouldn't need `package_manager_version` for the kernel task at all.

PR #544, opened 2016-11-15
ceph-{dev-}setup: Apparently quotes are not acceptable

dmick:
caused the expansion to look like (all quotes literal)
   '"' '-DALLOCATOR=libc"'


dmick:
@liewegas @alfredodeza

liewegas:
lgtm

PR #554, opened 2016-11-18
kernel: a new kernel-trigger job

idryomov:


kernel/config/definitions/kernel.yml
```diff
@@ -80,9 +83,7 @@ If this is checked, then the binaries will be built and pushed to chacra even if
 
     scm:
```
andrewschoen: you don't want ``scm`` in this job at all because that will make this job respond to pushes to ceph-client.git as well.

PR #555, opened 2016-11-21
Ceph PRs submodules check

alfredodeza:
If the pull request has modified submodules, then fail the build

### DIFF
--- a/ceph-pr-submodules/build/build
+++ b/ceph-pr-submodules/build/build
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+cd "$WORKSPACE"
+modified_submodules=$(git submodule foreach 'cd $toplevel; git diff --ignore-submodules=dirty origin/master -- $path')
+
+if [[ $modified_submodules ]]; then
+    echo "Project has modified submodules!"
+    echo $modified_submodules
+    exit 1
+fi
+
+exit 0
--- a/ceph-pr-submodules/build/build
+++ b/ceph-pr-submodules/build/build
@@ -1,6 +1,12 @@
 #!/bin/bash
 
 cd "$WORKSPACE"
+
+# The command below has access to the variables $name, $path, $sha1 and $toplevel:
+# $name is the name of the relevant submodule section in .gitmodules, $path is
+# the name of the submodule directory relative to the superproject, $sha1 is
+# the commit as recorded in the superproject, and $toplevel is the absolute
+# path to the top-level of the superproject.
 modified_submodules=$(git submodule foreach 'cd $toplevel; git diff --ignore-submodules=dirty origin/master -- $path')
 
 if [[ $modified_submodules ]]; then
--- a/ceph-pr-submodules/config/definitions/ceph-pr-commits.yml
+++ b/ceph-pr-submodules/config/definitions/ceph-pr-commits.yml
@@ -1,5 +1,6 @@
 - job:
     name: ceph-pr-submodules
+    node: centos || trusty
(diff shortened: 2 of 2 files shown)


ktdreyer:
I'm guessing this was put in place to avoid accidental submodule updates?

And administrators will have to override this failure when they want to explicitly update a submodule?

alfredodeza:
@ktdreyer correct

ceph-pr-submodules/build/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+cd "$WORKSPACE"
+modified_submodules=$(git submodule foreach 'cd $toplevel; git diff --ignore-submodules=dirty origin/master -- $path')
```
andrewschoen: Where does ``$toplevel`` come from here?

ceph-pr-submodules/config/definitions/ceph-pr-commits.yml
```diff
@@ -0,0 +1,54 @@
...
+            - ceph
+          trigger-phrase: ''
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "Unmodifed Submodules"
+          started-status: "checking if PR has modified submodules"
+          success-status: "submodules for project are unmodified"
+          failure-status: "found modified submodules"
```
ktdreyer: Maybe add "admin must override" here?

alfredodeza: That would potentially get out of sync easily when the configuration changes. We don't do any 'admin must override' type of messaging anywhere for that reason as well. It can get confusing to add something like that and have the option change

ktdreyer: I mean just alter the text message here from "found modified submodules" to "found modified submodules, admin must override this if its intentional" ?

We do expect this test to fail occasionally, right? Might as well give a hint to newcomers?

alfredodeza: That might be a good idea, but I would prefer that to be an all-or-nothing change across all PR jobs for this project. Otherwise it gets confusing when the messaging is not consistent.

ktdreyer: To me there's a distinction between "tests that we expect to occasionally fail as part of the regular development process" and "tests that should never fail". Distinguishing these categories makes it more user-friendly so we don't perpetuate the "just ignore Jenkins failures" idea :) That's why I think it makes sense for this particular change.

ceph-pr-submodules/config/definitions/ceph-pr-commits.yml
```diff
@@ -0,0 +1,54 @@
+- job:
+    name: ceph-pr-submodules
```
andrewschoen: Do you want to lock this to a certain node type? I think without this defined it'll use master which I don't think we want.

ceph-pr-submodules/config/definitions/ceph-pr-commits.yml
alfredodeza: Ah OK. I see what you mean. Will make that change. Makes sense

ceph-pr-submodules/build/build
yehudasa: @andrewschoen the git command itself parses it, it's an option of `git submodule foreach`

PR #564, opened 2016-11-23
ceph-ansible-galaxy: update the splitup script

leseb:
We now use the latest version of the splitup.sh script that now pushes
branches and tags from ceph-ansible to ansible-ceph-* repositories.

### DIFF
--- a/ceph-ansible-galaxy/build/build
+++ b/ceph-ansible-galaxy/build/build
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-# the following method exists in scripts/build_utils.sh
-pkgs=( "ansible" )
-install_python_packages "pkgs[@]"
-
-cd "$WORKSPACE"/ceph-ansible
 # propagates the change in the necessary Ansible Galaxy repos.
 # i.e. https://github.com/ceph/ansible-ceph-common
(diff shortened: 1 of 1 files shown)


ceph-ansible-galaxy/build/build
```diff
@@ -1,10 +1,6 @@
...
 
-# the following method exists in scripts/build_utils.sh
-pkgs=( "ansible" )
-install_python_packages "pkgs[@]"
-
 cd "$WORKSPACE"/ceph-ansible
 # propagates the change in the necessary Ansible Galaxy repos.
 # i.e. https://github.com/ceph/ansible-ceph-common
-$VENV/ansible-playbook -i dummy-ansible-hosts contrib/splitup.yml --tags update 
+bash contrib/splitup.sh
```
alfredodeza: this looks like it lives in the ceph-ansible source, to avoid path issues, can you specify the full path? this is provided by jenkins with `$WORKSPACE` so it would look like:

    bash "$WORKSPACE/contrib/splitup.sh"

That would make line 3 not necessary as well

leseb: Thanks, so is it `"$WORKSPACE"/ceph-ansible/contrib/splitup.sh` or `$WORKSPACE"/contrib/splitup.sh`

PR #568, opened 2016-11-28
Prado support

alfredodeza:
Prado needs these files at the top level because they depend on the directory structure relative to their location which doesn't work if they exist only on `examples/`

It will make it easier for deploying prado.


ansible/examples/slave.yml
```diff
@@ -194,28 +193,23 @@
...
-      failed_when: false
 
-    - name: update the etc cloud templates for debian /etc/hosts
+    - name: update etc cloud templates for red hat /etc/hosts
       sudo: true
       lineinfile:
+        backup: yes
         dest: /etc/cloud/templates/hosts.redhat.tmpl
         line: '127.0.1.1 ceph-builders'
-      failed_when: false
```
andrewschoen: Why did you remove ``failed_when: false`` from these last two tasks? Those template files won't exist on all slaves and I think ``lineinfile`` will fail if the ``dest`` doesn't exist.

ansible/examples/slave_libvirt.yml
```diff
@@ -74,7 +73,7 @@
       become_user: "{{ jenkins_user }}"
 
     - name: set the authorized keys
-      authorized_key: user={{ jenkins_user }} key="{{ lookup('file', 'files/ssh/keys/jenkins_build.pub') }}"
+      authorized_key: user={{ jenkins_user }} key="{{ lookup('file', 'playbook/files/ssh/keys/jenkins_build.pub') }}"
```
andrewschoen: Why was ``playbook`` added here? Is this something for ``prado`` support?

alfredodeza: ah this might be specific to prado then, I will add it back to examples (but not to the top level yaml files)

ansible/examples/slave.yml
alfredodeza: that is how this file currently exists in prado though

PR #577, opened 2016-11-30
Ceph dev setup new

alfredodeza:
Because ceph-dev-setup uses ceph.git and we need to do ceph-ci.git


ceph-dev-setup-new/config/definitions/ceph-dev-setup.yml
```diff
@@ -0,0 +1,59 @@
+- job:
```
andrewschoen: Would you change the filename here to ``ceph-dev-setup-new.yml``?  The config file for the ``ceph-dev-new`` job is also still named ``ceph-dev.yml``

PR #579, opened 2016-12-01
ceph-ansible-prs: add 2 scenarios

leseb:
Both:

* xenial_mon_osd: tests a 1 mon 1 osd xenial cluster using
raw_multi_journal OSD scenario

* xenial_conf_tests: test a 3 mon cluster and ensures
ceph.conf is rendered correctly

were missing, so adding them to the CI.

### DIFF
--- a/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
+++ b/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
@@ -3,6 +3,8 @@
     scenario:
       - ansible2.2-centos7_cluster
       - ansible2.2-xenial_cluster
+      - ansible2.2-xenial_conf_tests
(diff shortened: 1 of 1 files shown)


andrewschoen:
I thought we had decided to only test OSD scenarios on centos as there is little to no difference between the OS's and the OSD scenarios?

Also, if we bump ``xenial_cluster`` up to 3 mons (which might be a good idea anyway) we don't have a need for ``xenial_conf_tests``.

leseb:
@andrewschoen you're right, let's bump to 3 mon by default and revert this commit then. How does that sound?

andrewschoen:
@leseb that sounds good to me. We'd also need to change the ``scenario.py`` file for those cluster tests so they run the python test to ensure the conf is created correctly.

PR #581, opened 2016-12-01
Ceph dev new build

alfredodeza:
Because we need to get the artifacts from ceph-dev-new-setup, not ceph-dev-setup.

It also normalizes all the names because the previous changes where using the '-new' part in different places for jobs and configs


andrewschoen:
It's a real bummer to have all this duplication, I hope we won't have to run both of these setups for very long.

PR #583, opened 2016-12-02
Pass CEPH_EXTRA_CMAKE_ARGS through sudo to pbuilder build

dmick:
This should finally allow notcmalloc builds on debian.

Setting the variable on the sudo command line does affect the
resulting pbuilder debian/rules execution (tested with a dummy
debian package)

### DIFF
--- a/ceph-build/build/build_deb
+++ b/ceph-build/build/build_deb
@@ -72,8 +72,10 @@ echo deb vers $bpvers
 
 
 echo building debs for $DIST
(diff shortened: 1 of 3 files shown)


dmick:
My thought is that being explicit here is safest, and we can always add more variables if/when we need them.  (one not being passed now is SHA1)

alfredodeza:
Thank you @dmick !

PR #586, opened 2016-12-08
add ceph-ansible-rpm job

ktdreyer:
This job builds RPMs for each ceph-ansible GitHub branch.

(This is a 99% copy and paste from ceph-installer-rpm, so we probably need to come up with a centralized area for these types of jobs and use JJB templates. I can't come up with a good directory name at the moment, though.)

### DIFF
--- a/ceph-ansible-rpm/build/build
+++ b/ceph-ansible-rpm/build/build
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+set -ex
+
+# Sanity-check:
+[ -z "$GIT_BRANCH" ] && echo Missing GIT_BRANCH variable && exit 1
+[ -z "$JOB_NAME" ] && echo Missing JOB_NAME variable && exit 1
+
+# Strip "-rpm" off the job name to get our package's name
+PACKAGE=${JOB_NAME%-rpm}
+
+sudo yum -y install epel-release
+sudo yum -y install fedpkg
+
+# Add the Jenkins slave UID to the mock group.
+sudo usermod -a -G mock $(whoami)
+
+# Attempt the build. If it fails, print the mock logs to STDOUT.
+make rpm || ( tail -n +1 {root,build}.log && exit 1 )
+
+# Chacra time
+
+pkgs=( "chacractl>=0.0.4" )
+install_python_packages "pkgs[@]"
+
+# ask shaman which chacra instance to use
+chacra_url=`curl -u $SHAMAN_API_USER:$SHAMAN_API_KEY https://shaman.ceph.com/api/nodes/next/`
+# create the .chacractl config file using global variables
+make_chacractl_config $chacra_url
+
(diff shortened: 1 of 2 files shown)


ErwanAliasr1:
I don't get how you integrate the commit id or the branch name in the rpm versioning. Isn' t that required ?

ktdreyer:
Yes, that logic for integrating the commit ID is in ceph-ansible itself, specifically https://github.com/ceph/ceph-ansible/blob/master/Makefile

ErwanAliasr1:
As you use git-describe isn't that misleading as we use the last tag + commit-id ?

That's sounds perfect when you run on the tag itself but once you're in a branch, we are not more 1.08 and especially when running in a branch. Why not keeping the branch name ?

Using something like 'git branch 2> /dev/null | sed -e '/^[^*][\t]*/d' -e 's/*\( .*\)/ \1/''

That could be useful to get from what branch/commit the rpm is about.

ktdreyer:
Erwan, `git describe --tags ` shows "v2.0.0-163-g2c98b1d" on ceph-ansible's master branch, rather than 1.0.8. "2.0.0 plus some commits" sounds right to me.

This convention of using `git describe` for the package's version originates from what Ceph itself has done for a long time, and it works well imho.

You can still find out the branch name by checking shaman / chacra, since Jenkins will upload to `chacra_endpoint="${PACKAGE}/${BRANCH}/${GIT_COMMIT}/centos/7"`

ErwanAliasr1:
My point more about the rpm itself. If I get a rpm on a machine, how can I know from what branch it comes from ?

Rpms are also made for keeping a kind of tracability. We have it by the version + commit id but that doesn't give the branch name.

Maybe I'm misleading about the usage of those rpms. If no human use them or report a bug regarding this rpm, that's ok. But if humans are processing the rpms, it could be interesting to add the branch name.

ktdreyer:
> If I get a rpm on a machine, how can I know from what branch it comes from ?

You can run `yum -C list ceph-ansible`, and it will tell you the name of the Yum repository from which the RPM was installed. Even if the repo name is something generic, you can then look at the repository's definition in `/etc/yum.repos.d`, and the chacra URL there will include the branch name.

Alternatively, since many branch heads can point at a single sha1, you can also use Git:
* `git show-ref --heads | grep <sha1>` to find the branch(es) that point at this sha1
* `git branch --contains <sha1>` to find all the branch(es) that contain this sha1

ErwanAliasr1:
I meant from the system. If the branch died or being rebased, the commit id have no value. I'll not block that PR, just wondering about the usage ;)

ktdreyer:
> I meant from the system.

Yeah, in that case you can use the `yum list` thing above and inspect the originating repo's URL.

ktdreyer:
Whoops, I used the wrong YAML file directory. Fix is https://github.com/ceph/ceph-build/pull/592

PR #587, opened 2016-12-09
samba: Fix-up the build

b-ranto:
I was able to perform a centos7 build with these changes. The ubuntu
builds fail because the libgnutls-dev version is too knew [1] -- this should
be fixable in samba sources.

[1] I hit the same issue as https://lists.samba.org/archive/samba/2015-April/191253.html but the workaround did not work for me as the libgnutls28-dev was in the failing 3.4 version as well :-/

### DIFF
--- a/samba/build/build_deb
+++ b/samba/build/build_deb
@@ -8,77 +8,48 @@ fi
 
 
 ## Install any setup-time deps
-# TODO -- upload install-deps.sh script that installs the setup time deps
-# to upstream repo or update this section to install these packages
+# We need these for the build
+sudo apt-get install -y build-essential equivs libgnutls-dev libacl1-dev libldap2-dev ruby ruby-dev
 
-# We need this for mk-build-deps
-sudo apt-get install -y equivs
-
-# Run the install-deps.sh upstream script if it exists
-if [ -x install-deps.sh ]; then
-    echo "Ensuring dependencies are installed"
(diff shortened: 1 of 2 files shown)


samba/build/build_deb
```diff
@@ -8,77 +8,48 @@ fi
 
 
 ## Install any setup-time deps
-# TODO -- upload install-deps.sh script that installs the setup time deps
-# to upstream repo or update this section to install these packages
+# We need these for the build
+sudo apt-get install -y build-essential equivs libgnutls-dev libacl1-dev libldap2-dev ruby ruby-dev
```
alfredodeza: ruby? :(

build-essential should already be there

samba/build/build_deb
```diff
@@ -8,77 +8,48 @@ fi
...
 
-# We need this for mk-build-deps
-sudo apt-get install -y equivs
+# We use fpm to create the deb package
+sudo gem install fpm
```
alfredodeza: oh because fpm

b-ranto:
The thing we had before did not actually work, it was just a template for ira to fill it in. This version is based on the version from gitbuilders, see the https://github.com/ceph/autobuild-ceph/blob/master/build-samba-deb.sh script.

Yes, it needs ruby for the fpm. I'd prefer to use a more standardized way to build this but this is probably as good as we can do in a reasonably short time.

PR #588, opened 2016-12-09
samba: Minor fix-ups

b-ranto:
This also fixes the get_distro_and_target function to get the proper
DIST and DISTRO for the new trusty-pbuilder-like tags.


alfredodeza:
Do you want to address the chacractl issue that pushes trusty-pbuilder to the wrong place?

b-ranto:
This should address it as well by overriding the DIST variable.

alfredodeza:
I can't understand BASH of course :)

PR #591, opened 2016-12-12
fixes for ceph-docker-nightly

alfredodeza:


ceph-docker-nightly/build/build
```diff
@@ -1,14 +1,17 @@
 #!/bin/bash
 
 # the following two methods exist in scripts/build_utils.sh
-pkgs=( "pytest" )
+pkgs=( "pytest" "docker-py==1.10.3" )
```
ktdreyer: I think I know why you're doing this (API in 2 is different?) and it's not related to this commit, right?

alfredodeza: The docker API changes very very fast, and I am pinning here because I want to prevent having issues while we try to implement/stabilize

ceph-docker-nightly/config/definitions/ceph-docker-nightly.yml
```diff
@@ -1,6 +1,6 @@
 - job:
     name: ceph-docker-nightly
-    node: small && trusty
+    node: xenial && amd64
```
andrewschoen: These labels don't look correct. Did you mean ``xenial && x86_64``?

alfredodeza: omg let me fix

alfredodeza:
@andrewschoen this is good to go I think

PR #592, opened 2016-12-12
ceph-{ansible,installer}-rpm: fix config location

ktdreyer:
Our jenkins-job-builder job expects the JJB YAML to be located in a `config/definitions` directory. Add the child "definitions" directory here.


ktdreyer:
Thanks for cluing me in on that one. I've updated the PR to fix that issue, as well as the identical problem in the ceph-installer-rpm job.

PR #595, opened 2016-12-14
ceph-ansible-prs: move Ansible version out into separate var

ktdreyer:
Move the ansible version definition out into its own variable.

The purpose of this change is to make it easier to add multiple Ansible
versions.

This is code reorganization only, and this should not introduce a
behavior change.


alfredodeza:
@ktdreyer I think this was already addressed, can we close this PR?

ktdreyer:
yep thanks

PR #596, opened 2016-12-14
ceph-setup: move the git diff check before calling autogen.sh

alfredodeza:

### DIFF
--- a/ceph-setup/build/build
+++ b/ceph-setup/build/build
@@ -53,6 +53,28 @@ else
     export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
 fi
 
+
+# When using autotools/autoconf it is possible to see output from `git diff`
+# since some macros can be copied over to the ceph source, triggering this
+# check. This is why this check now is done just before running autogen.sh
+# which calls `aclocal -I m4 --install` that copies a system version of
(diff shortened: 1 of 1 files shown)


ktdreyer:
LGTM, and it sounds like we can revert after Hammer is EOL

smithfarm:
@alfredodeza Can you add a `Fixes: http://tracker.ceph.com/issues/17901` line before Signed-off-by ?

dmick:
@ktdreyer don't see any need to revert; it was always a risk for no good reason.  If anything, I'd put a commit into hammer with the newer m4 file(s).

smithfarm:
I guess this isn't a problem in jewel because `src/gmock` is a submodule there?

ceph-setup/build/build
```diff
@@ -53,6 +53,25 @@ else
     export CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS -DALLOCATOR=tcmalloc"
 fi
 
+
+# When using autotools/autoconf it is possible to see output from `git diff`
```
dmick: if we're gonna add three lines of comment (great!) let's actually call specific attention to "autogen.sh calls aclocal with --install, which can..." for documentation/archaeologists' sakes

dmick:
a tweak from me and one from @smithfarm and it's ready to go afaic

ktdreyer:
LGTM

smithfarm:
Thanks again, everyone.

PR #598, opened 2016-12-15
ceph-{ansible,installer}-rpm: fix mock group addition

ktdreyer:
When an ephemeral slave is first created, we have to add the jenkins-build UID to the `mock` group so that it can run mock.

The problem is that `usermod -a -G ...` will not affect the current login shell. When the ceph-*-rpm jobs run `usermod` for the first time, the jenkins-build UID obtains those group rights for any future jobs, but not the current job.

The result is that every time an ephemeral slave boots, it will fail its first ceph-*-rpm build job (and future ones succeed).

Call `newgrp` to refresh the mock group for this login shell, so that all mock operations succeed, including the first one on each slave.

### DIFF
--- a/ceph-ansible-rpm/build/build
+++ b/ceph-ansible-rpm/build/build
@@ -14,6 +14,7 @@ sudo yum -y install fedpkg
 
 # Add the Jenkins slave UID to the mock group.
 sudo usermod -a -G mock $(whoami)
+newgrp mock
 
 # Attempt the build. If it fails, print the mock logs to STDOUT.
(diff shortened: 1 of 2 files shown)


alfredodeza:
This might still not work for slaves that use JNLP. For reference, this might be useful to see:

https://github.com/ceph/ceph-build/commit/f9cc49aec026c2ae17b1db2603d262f5cb1c8a00

ktdreyer:
You're right, #epicfail

PR #599, opened 2016-12-15
ceph-pr-submodules: increase verbosity to try and get values out of vars

alfredodeza:
The check is returning false positives and it is hard to tell why without these -ex flags


ktdreyer:
I rebased this so the tests pass

PR #601, opened 2016-12-16
ceph-{ansible,installer}-rpms: do not use sepia slaves or sg

andrewschoen:
Our slaves in sepia, which use JNLP, are having troubles building these repos. The first issue we found was that the ``jenkins-build`` user was not added to the ``mock`` group properly on these sepia slaves. The attempted solution to that was proposed in: https://github.com/ceph/ceph-build/pull/600

The solution from https://github.com/ceph/ceph-build/pull/600 worked in that the mock command was able to execute without permissions errors but the builds were failing to build the 'noarch' rpm.  Here's an example of one of those repos: https://1.chacra.ceph.com/r/ceph-installer/master/1cd19fb1217c4f1b2083d85f3d148375e4106eb6/centos/7/flavors/default/noarch/

Let's just use the OVH slaves which we know can build these rpms until we have some automated tests in place, then we can use those tests to verify changes in these build scripts are working correctly.

### DIFF
--- a/ceph-ansible-rpm/build/build
+++ b/ceph-ansible-rpm/build/build
@@ -14,9 +14,10 @@ sudo yum -y install fedpkg
 
 # Add the Jenkins slave UID to the mock group.
 sudo usermod -a -G mock $(whoami)
+newgrp mock
 
 # Attempt the build. If it fails, print the mock logs to STDOUT.
-sg mock -c "make rpm" || ( tail -n +1 {root,build}.log && exit 1 )
+make rpm || ( tail -n +1 {root,build}.log && exit 1 )
 
 # Chacra time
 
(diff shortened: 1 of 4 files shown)


ktdreyer:
My wild guess is that sg eats the exit code when "make rpm" fails.

alfredodeza:
we need to get to the bottom of this for a fix. I am sure we will hit issues with anything else that needs a group created on a JNLP slave

ktdreyer:
Does restarting the JNLP daemon fix it?

andrewschoen:
It might, but if we restart the JNLP daemon during a job the slave will lose connection with jenkins and the job would fail I think.

ktdreyer:
Oh, sure, I mean this would just be a one-time restart.

andrewschoen:
@ktdreyer maybe we should just add ``jenkins-build`` to the ``mock`` group in the ansible playbook that configures these slaves.

ktdreyer:
Agreed, I think that would be the best option, from what we know today

PR #602, opened 2016-12-19
Adds a nightly job that tests ceph-installer

andrewschoen:
This job is parametrized so that we can also run it manually or
trigger it easily from other jobs.


alfredodeza:
For nightlies I think we need to send some emails out when we see failures. Could you add those to this PR? If they aren't working or Jenkins needs additional configuration, lets get that done afterwards

PR #604, opened 2016-12-20
nfs-ganesha: first run at a build script for rpms and debs

alimaredia:
Need to quickly address issues marked "ALI TO DO" before able to merge.

### DIFF
--- a/nfs-ganesha/build/build_deb
+++ b/nfs-ganesha/build/build_deb
@@ -1,0 +1,85 @@
+#! /usr/bin/bash
+set -ex
+
+# Only do actual work when we are a DEB distro
+if test "$DISTRO" != "debian" -a "$DISTRO" != "ubuntu"; then
+    exit 0
+fi
+
+
+## Install any setup-time deps
+# TODO -- upload install-deps.sh script that installs the setup time deps
+# to upstream repo or update this section to install these packages
+
+# We need this for mk-build-deps
+sudo apt-get install -y equivs cmake git
+
+## Setup the pbuilder
+# TODO remove if you do not want to use pbuilders
+setup_pbuilder
+
+
+## Build the source tarball
+
+cd $WORKSPACE/nfs-ganesha
+
+git submodule update --init || git submodule sync
+
+mkdir build
+cd build
+
+# make source tarball
+cmake -DCMAKE_BUILDER_TYPE=Maintainer $WORKSPACE/nfs-ganesha/src && make dist || exit 0
+
+## Get some basic information about the system and the repository
+DEB_ARCH=$(dpkg-architecture -qDEB_BUILD_ARCH)
+VERSION=`grep -om 1 "[0-9]\.[0-9]-dev-[0-9]" include/config.h`
+
+# terrible hack
+cd $WORKSPACE/nfs-ganesha-debian
+mv $WORKSPACE/nfs-ganesha/build/nfs-ganesha-*.tar.gz .
+tar -xvf nfs-ganesha-*.tar.gz
+rm *.tar.gz
+mv nfs-ganesha* src
+
+## Prepare the debian files
+# Bump the changelog
+dch -v "$VERSION" "New release ($VERSION)"
+
+# Create .dsc and source tarball
+sudo dpkg-buildpackage -S -us -uc
+
+## Build with pbuilder
+echo "Building debs"
+
+PBUILDDIR="/srv/debian-base"
+
+sudo pbuilder --clean
+
+mkdir -p dist/deb
+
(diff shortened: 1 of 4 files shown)


nfs-ganesha/build/build_deb
```diff
@@ -0,0 +1,90 @@
...
+## Get some basic information about the system and the repository
+# TODO -- update this to get the proper VERSION/REVISION
+DEB_ARCH=$(dpkg-architecture -qDEB_BUILD_ARCH)
+# ALI TO DO GET VERSION
+VERSION="$(./get-version.sh)"
+
+
+## Build the source tarball
+# ALI TO DO: find out if this can be set in config/definitions/nfs-ganesha.yaml
+NFS_GANESHA_BRANCH="next"
```
alfredodeza: yes, you can specify what branch(es) you want to build in the configuration. See http://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git

nfs-ganesha/build/build_deb
```diff
@@ -0,0 +1,90 @@
...
+## Build with pbuilder
+echo "Building debs"
+
+PBUILDDIR="/srv/debian-base"
+
+sudo pbuilder --clean
+
+mkdir -p dist/deb
+
+# ALI TO DO: find out if there's a env var for CEPH_BRANCH
```
alfredodeza: there isn't one available unless you make it a parameter or define it as a variable for this job

nfs-ganesha/build/build_deb
```diff
@@ -0,0 +1,90 @@
...
+CEPH_BRANCH="master"
+SHA1="latest"
+
+# ALI TO DO: see if the repo url is correct
+REPO_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$SHA1/$DISTRO/$DIST"
```
alfredodeza: this is not the repo url, it is a url that knows where to find the repository file and it is missing the `/repo` at the end. So it should be:

    SHAMAN_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$SHA1/$DISTRO/$DIST/repo"

So that you can `curl -L` later, unless what you need is to find out the actual repo, in which case you can use the API to fetch that information (it returns JSON)

nfs-ganesha/build/build_deb
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Get some basic information about the system and the repository
+# TODO -- update this to get the proper VERSION/REVISION
+DEB_ARCH=$(dpkg-architecture -qDEB_BUILD_ARCH)
+VERSION=`grep -om 1 "[0-9]\.[0-9]-dev-[0-9]" include/config.h`
+
+## Build the source tarball
+
+git checkout $NFS_GANESHA_BRANCH
```
alfredodeza: you may want to do this in the yaml configuration file to get the desired branch already checked out (instead of doing it manually here)

nfs-ganesha/build/build_deb
```diff
@@ -0,0 +1,85 @@
...
+git submodule update --init || git submodule sync
+
+mkdir build
+cd build
+
+# make source tarball
+cmake -DCMAKE_BUILDER_TYPE=Maintainer ../src && make dist || exit 0
```
alfredodeza: you should use absolute paths using `$WORKSPACE` so that you avoid issues when changing paths or getting the relative paths mixed up. It also makes it very clear to anyone else reading the script

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -0,0 +1,128 @@
...
+          values:
+            - ARCHS
+
+    scm:
+      - git:
+          url: git@github.com/nfs-ganesha/nfs-ganesha.git
+          # Use the SSH key attached to the ceph-jenkins GitHub account.
+          credentials-id: '39fa150b-b2a1-416e-b334-29a9a2c0b32d'
+          branches:
+            - $NFS_GANESHA_BRANCH
```
alfredodeza: so you are already defining here the branch, so maybe don't manually checkout that branch again in the script?

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -0,0 +1,128 @@
...
+          skip-tag: true
+          wipe-workspace: true
```
alfredodeza: you should add the `basedir` here to avoid having the `nfs-ganesha-debian` living in the same topdir. This will also mean you need to update the scripts to execute in the right location(s)

PR #606, opened 2017-01-04
ceph-deploy-build: do not use python to detect the distro

andrewschoen:
We're using pbuilder in many build jobs now so we can't depend on the
host to be the actual distro version that the binaries are being built
for. This uses the value from the JJB matrix to determine what distro
value to use when posting to chacra.


ktdreyer:
#bashchampion

alfredodeza:
For future reference, it looks like we already have helpers to deal with this `$DIST` varuable in `scripts/build_utils.sh` which includes everything that we may be using

PR #607, opened 2017-01-04
report when a build is triggered

alfredodeza:
This will (unfortunately) need to also be ported to `ceph-dev-new-trigger` once it is up and running. It is safer to just use it on ceph-dev-trigger for now

### DIFF
--- a/ceph-dev-trigger/build/notify
+++ b/ceph-dev-trigger/build/notify
@@ -1,0 +1,9 @@
+#!/bin/bash -ex
+
+# update shaman with the triggered build status. At this point there aren't any
+# architectures or distro information, so we just report this with the current
+# build information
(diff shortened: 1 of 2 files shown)


ceph-dev-trigger/build/notify
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -ex
+
+# update shaman with the triggered build status. At this point there aren't any
+# architectures or distro information, so we just report this with the current
+# build information
+BRANCH=`branch_slash_filter ${GIT_BRANCH}`
+SHA1=${GIT_COMMIT}
+
+create_build_status "triggered" "ceph"
```
andrewschoen: I think ``queued`` might be a better status name here.

PR #609, opened 2017-01-06
nfs-ganesha: fix scm section, add absolute paths

alimaredia:
Made changes to scm section to configure both git
repos.

Made all references to "dist" dir absolute paths

Made sure directories being git cleaned are
actual git repos

### DIFF
--- a/nfs-ganesha/build/build_rpm
+++ b/nfs-ganesha/build/build_rpm
@@ -41,12 +41,12 @@ cmake -DCMAKE_BUILDER_TYPE=Maintainer -DUSE_FSAL_RGW=YES -DUSE_FSAL_CEPH=YES $WO
 ## Create the source rpm
 echo "Building SRPM"
 rpmbuild \
-    --define "_sourcedir ./dist" \
+    --define "_sourcedir $WORKSPACE/dist" \
     --define "_specdir ." \
     --define "_builddir ." \
     --define "_srcrpmdir ." \
     --define "_rpmdir ." \
-    --nodeps -bs ../src/nfs-ganesha.spec
+    --nodeps -bs $WORKSPACE/nfs-ganesha/src/nfs-ganesha.spec
 SRPM=$(readlink -f *.src.rpm)
 
 # add repo file to mock config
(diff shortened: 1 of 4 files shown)


nfs-ganesha/build/build_rpm
```diff
@@ -41,12 +41,12 @@ cmake -DCMAKE_BUILDER_TYPE=Maintainer -DUSE_FSAL_RGW=YES -DUSE_FSAL_CEPH=YES $WO
 ## Create the source rpm
 echo "Building SRPM"
 rpmbuild \
-    --define "_sourcedir ./dist" \
+    --define "_sourcedir $WORKSPACE/dist" \
     --define "_specdir ." \
```
alimaredia: @b-ranto, do i need to change the paths for _specdir, _builddir, _scrpmdir, and _rpmdir to $WORKSPACE/dist in this command?

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -61,13 +61,9 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           type: label-expression
           name: AVAILABLE_DIST
           values:
-            - centos6
```
alimaredia: @alfredodeza were these values necessary? I'm confused about which distros get triggered to build when I push the "Build" button in "Build With Parameters" section of the jenkins UI.

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -83,26 +104,8 @@ If this is checked, then the binaries will be built and pushed to chacra even if
...
-          url: https://github.com/nfs-ganesha/nfs-ganesha-debian.git
-          # Use the SSH key attached to the ceph-jenkins GitHub account.
-          credentials-id: '39fa150b-b2a1-416e-b334-29a9a2c0b32d'
-          branches:
-            - xenial-nfs-ganesha-ceph-fsals
-          skip-tag: true
-          wipe-workspace: true
-          basedir: "nfs-ganesha-debian"
+      - nfs-ganesha
+      - nfs-ganesha-debian
```
alfredodeza: this looks correct

nfs-ganesha/config/definitions/nfs-ganesha.yml
alfredodeza: this section would list everything that you think the project is able to build. That is why this is named `AVAILABLE_DIST`. It allows for filtering the list from user input, e.g. 'just build in centos7'. You can remove centos6 if that will never ever be built (I think that may be the case here)

nfs-ganesha/build/build_rpm
b-ranto: @alimaredia It depends on where your .spec file is, where you want your (src.)rpms to be located, etc. The . is used in place of $PWD in these cases if you were asking about that.

PR #625, opened 2017-01-26
nfs-ganesha: fixes to enable "Builds" section in shaman

alimaredia:

### DIFF
--- a/nfs-ganesha/config/definitions/nfs-ganesha.yml
+++ b/nfs-ganesha/config/definitions/nfs-ganesha.yml
@@ -38,6 +38,11 @@
           default: "next"
 
       - string:
+          name: BRANCH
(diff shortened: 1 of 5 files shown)


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -38,6 +38,11 @@
           default: "next"
 
       - string:
+          name: BRANCH
```
andrewschoen: Does this need to be exposed as a parameter? Why not just set ``BRANCH=$NFS_GANESHA_BRANCH`` in the ``build/setup``?

You'll also need to make sure you use the ``branch_slash_filter`` method from ``scripts/build_utils.sh`` on the branch value because a branch name with ``/`` in the name will break the shaman urls.

PR #630, opened 2017-02-08
Samba+calamari: minor fixes

b-ranto:
We used DEB_ARCH instead of ARCH when pushing to chacra in
samba/calamari. This meant that we pushed to amd64 instead of x86_64.


b-ranto:
I have also pushed a commit that should help fix ~random build failures when installing dependencies.

ktdreyer:
Mind updating the title of this PR? the string goes into the Git commit log when we merge.

b-ranto:
I'll go ahead and merge these so that they don't get overwritten by next merge (I used them in the web UI).

PR #634, opened 2017-02-15
nfs-ganesha: Debian builds working various fixes

alimaredia:
.debs building with pbuilder

remove $BRANCH variable

exit 0's are now exit 1

various minor fixes

### DIFF
--- a/nfs-ganesha/build/build_deb
+++ b/nfs-ganesha/build/build_deb
@@ -6,11 +6,16 @@ if test "$DISTRO" != "debian" -a "$DISTRO" != "ubuntu"; then
     exit 0
 fi
 
+REPO_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$CEPH_SHA1/$DISTRO/$DIST/repo"
+if `curl --fail -L $REPO_URL`; then
+  echo "Ceph debian lib repo exists in shaman"
+else
+  echo "Ceph debian lib repo does NOT exist in shaman"
+  exit 1
+fi
 
-# We need this for system, to build ganesha, and to enable FSAL_CEPH and 
-# FSAL_RGW in the .spec file when cmake command runs
+# We need this for system and to run the cmake
 sudo apt-get update
-sudo apt-get -y install cmake equivs libkrb5-dev bison flex quilt librgw-dev libcephfs-dev
 
 # Normalize variables across rpm/deb builds
 NORMAL_DISTRO=$DISTRO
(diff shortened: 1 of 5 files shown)


alimaredia:
@andrewschoen do you know if this line is necessary? https://github.com/ceph/ceph-build/blob/master/nfs-ganesha/build/build_deb#L104

nfs-ganesha/build/build_deb
```diff
@@ -3,14 +3,21 @@ set -ex
...
+else
+  echo "Ceph debian lib repo does NOT exist in shaman"
+  exit 1
+fi
 
 # We need this for system, to build ganesha, and to enable FSAL_CEPH and 
 # FSAL_RGW in the .spec file when cmake command runs
 sudo apt-get update
-sudo apt-get -y install cmake equivs libkrb5-dev bison flex quilt librgw-dev libcephfs-dev
+sudo apt-get -y install cmake equivs libkrb5-dev bison flex quilt ceph
```
ktdreyer: Why do we need the full ceph package here instead of just libcephfs-dev and librgw-dev?

andrewschoen:
@alimaredia yes, that is a necessary as teuthology needs it.

nfs-ganesha/build/build_deb
alimaredia: strange i was getting errors telling me I should install the full ceph package instead of just those two libraries but I just did a run of those two and it passed.

I'll make that change and send it back over here

alimaredia:
@andrewschoen @ktdreyer I added some more changes, and have nothing else to add to the PR apart from any issues you find in it.

nfs-ganesha/build/build_deb
```diff
@@ -3,14 +3,21 @@ set -ex
...
+REPO_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$CEPH_SHA1/$DISTRO/$DIST/repo"
+if `curl --fail -L $REPO_URL`; then
+  echo "Ceph debian lib repo exists in shaman"
+else
+  echo "Ceph debian lib repo does NOT exist in shaman"
+  exit 1
+fi
 
 # We need this for system, to build ganesha, and to enable FSAL_CEPH and 
 # FSAL_RGW in the .spec file when cmake command runs
```
ktdreyer: This comment mentions "the .spec file", but this script is for building Debian packages. Want to update this?

It would also help to make it clearer in the comments exactly why we need this huge list of -dev packages when we're just going to use pbuilder to resolve them automatically. Does the `cmake` command or `make dist` command fail when those are not installed on the host?

nfs-ganesha/build/build_deb
```diff
@@ -36,15 +44,24 @@ cd build
 # make source tarball
 cmake -DCMAKE_BUILD_TYPE=Maintainer -DUSE_FSAL_GLUSTER=OFF -DUSE_FSAL_CEPH=ON -DUSE_FSAL_RGW=ON $WORKSPACE/nfs-ganesha/src && make dist || exit 1
```
ktdreyer: We run `make dist` here, but then we also `tar the "src" up` later on? Why don't we choose one or the other?

nfs-ganesha/build/build_deb
alimaredia: @ktdreyer some of the packages are needed for the cmake run, others are needed because I don't pass the '-d' flag to dpkg-buildpackage which aborts by default if the build dependencies aren't on the host machine. I can go ahead and add it back in, but I was told it's bad practice. Would you have me change another way?

I'll fix the ".spec file" typo as well!

nfs-ganesha/build/build_deb
alimaredia: we need the it to be in a directory called "src" because the debian/rules file looks for a directory with that name (https://github.com/nfs-ganesha/nfs-ganesha-debian/blob/xenial-nfs-ganesha-ceph-fsals/debian/rules#L30). 

I could simplify the process a little bit by removing the .git repo, tar'ing just the "src" by hand, and forgoing the cmake command, but I wanted to keep it similar for consistency. All the steps after that command are confusing though.

Would you recommend any changes?

nfs-ganesha/build/build_deb
```diff
@@ -27,6 +34,7 @@ setup_pbuilder
 ## Build the source tarball
 
 cd $WORKSPACE/nfs-ganesha
+PACKAGE_MANAGER_VERSION="`git describe --long | 'sed s/V//1'`-1xenial"
```
andrewschoen: Hardcoding ``xenial`` won't work if this job is used to build any other version. I think you can get the value you want here from the ``$DIST`` variable.  We should probably change that for every instance of ``xenial``.

nfs-ganesha/build/build_deb
ktdreyer: It's fine to use the `-d` flag when running `dpkg-buildpackage` *with* `-S`. In fact that is how the `git-buildpackage` tool does it.

It will save time in the job if you don't have to install all these dependency packages here, plus you'll avoid the maintenance burden of trying to keep this list in sync with whatever the nfs-ganesha `/debian/control` file has for `Build-Depends`.

nfs-ganesha/build/build_deb
ktdreyer: Is the tarball from `make dist` not sufficient because it lacks the `src` directory? It seems like we're papering over some bigger issue here if we're constructing two separate tarballs and throwing one away?

alimaredia: yes the tarball `make dist` creates doesn't have a src directory, it is the `src` directory. I can override some of the cpack config variables to make the process a little cleaner looking.

ktdreyer: I recommend checking with Kaleb Keithley to see how he creates the tarball for Debian packages [in the PPA](https://launchpad.net/~gluster/+archive/ubuntu/nfs-ganesha), because he must've dealt with this issue.

alimaredia:
@andrewschoen @ktdreyer a new wave of changes has been made!

PR #639, opened 2017-02-22
nfs-ganesha: remove nfs-ganesha-trigger job, try polling instead

alimaredia:

This is a placeholder for near term future work, if the nfs-ganesh-trigger job is deemed suboptimal.


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -105,6 +108,14 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           name: ARCH
           values:
             - ARCHS
+    triggers:
+      - github
+      - pollurl:
+          urls:
+            - url: https://shaman.ceph.com/api/repos/ceph/master/latest/centos/7/repo
```
andrewschoen: But what changes are you polling for here? I think you'll need to define ``check-content`` at least for these urls.

https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.pollurl

alimaredia: @andrewschoen added it, do you think i need to add the "text" content type underneath simple when the SHA1's change?

andrewschoen: @alimaredia nah, I think this ought to be enough.

PR #649, opened 2017-02-27
nfs-ganesha: add cron option to pollurl

alimaredia:


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -111,6 +111,7 @@ If this is checked, then the binaries will be built and pushed to chacra even if
     triggers:
       - github
       - pollurl:
+          cron: '*/30 * * * *'
```
alimaredia: @andrewschoen do you think polling every 30 minutes is enough?

andrewschoen: Yeah, I think so. It takes about ~ 1 to 1.5 hours to get a build finished and a repo created.

PR #653, opened 2017-03-03
nfs-ganesha: scenario based polling for ceph triggered builds

alimaredia:

### DIFF
--- a/nfs-ganesha/config/definitions/nfs-ganesha.yml
+++ b/nfs-ganesha/config/definitions/nfs-ganesha.yml
@@ -1,3 +1,11 @@
+- project:
+    name: nfs-ganesha
+    ceph-branch:
+      - master
+      - kraken
+    jobs:
+      - 'nfs-ganesha-{ceph-branch}'
+
 - scm:
     name: nfs-ganesha
     scm:
@@ -23,11 +31,12 @@
           skip-tag: true
           wipe-workspace: true
           basedir: "nfs-ganesha-debian"
-- job:
-    name: nfs-ganesha
+
+- job-template:
+    name: 'nfs-ganesha-{ceph-branch}'
     project-type: matrix
     defaults: global
-    display-name: 'nfs-ganesha'
+    display-name: 'nfs-ganesha: Triggered by {ceph-branch}'
     block-downstream: false
     block-upstream: false
     properties:
@@ -48,7 +57,7 @@
       - string:
           name: CEPH_BRANCH
           description: "The branch of Ceph to get the repo file of for libcephfs"
-          default: "master"
+          default: {ceph-branch}
 
       - string:
           name: DISTROS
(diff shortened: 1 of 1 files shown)


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -132,6 +141,9 @@ If this is checked, then the binaries will be built and pushed to chacra even if
       - nfs-ganesha-debian
 
     builders:
+      - inject:
+          properties-content: |
+            CEPH_BUILD={scenario}
```
alimaredia: @andrewschoen is this redundant or preferred over setting {scenario} as the default for $CEPH_BUILD?

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -23,11 +31,12 @@
           skip-tag: true
           wipe-workspace: true
           basedir: "nfs-ganesha-debian"
-- job:
-    name: nfs-ganesha
+
+- job-template:
```
alimaredia: @andrewschoen "job-template" is just supposed to substitute for "job" right?

nfs-ganesha/config/definitions/nfs-ganesha.yml
andrewschoen: I think because you've used it as the default value for ``CEPH_BRANCH`` you won't need this. You'll just need to use ``CEPH_BRANCH`` in your build scripts.

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -1,3 +1,11 @@
+- project:
+    name: nfs-ganesha
+    scenario:
```
andrewschoen: This doesn't need to be called ``scenario``, it could just be ``ceph-branch`` or something that identifies what this value better.

nfs-ganesha/config/definitions/nfs-ganesha.yml
andrewschoen: yes, that's right. here's some good docs on how ``project`` and ``job-template`` works together: https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job-template

nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -114,16 +123,16 @@ If this is checked, then the binaries will be built and pushed to chacra even if
       - pollurl:
           cron: '*/30 * * * *'
           urls:
-            - url: https://shaman.ceph.com/api/repos/ceph/master/latest/centos/7/repo
+            - url: https://shaman.ceph.com/api/repos/ceph/{scenario}/latest/centos/7/repo
```
andrewschoen: The job template is rendered for each additional field you've defined, in this case ``scenario``. This means that you'll have two duplicate polling urls here. The same problem exists for the urls looking for new xenial repos.

alimaredia:
@andrewschoen I think it's fine that there's two jobs because they should be both building with different ceph branches everytime a push happens, and there's only one update to the "next" branch every week, so the volume won't be very high. 

I made the changes you recommended. Do you know what the jenkins check is failing?

PR #654, opened 2017-03-07
nfs-ganesha: trigger builds from ceph-dev-build

alimaredia:

### DIFF
--- a/ceph-dev-build/config/definitions/ceph-dev-build.yml
+++ b/ceph-dev-build/config/definitions/ceph-dev-build.yml
@@ -79,6 +79,11 @@
             - ../../../scripts/build_utils.sh
             - ../../build/setup_rpm
             - ../../build/build_rpm
+      - trigger-builds:
+        - project: "nfs-ganesha"
+          current-parameters: True
+          predefined-parameters:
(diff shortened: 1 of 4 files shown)


alimaredia:
@andrewschoen I used the "trigger-builds" section from here: (https://docs.openstack.org/infra/jenkins-job-builder/builders.html#builders.trigger-builds) was that what you wanted me to use?

ceph-dev-build/config/definitions/ceph-dev-build.yml
```diff
@@ -79,6 +79,11 @@
             - ../../../scripts/build_utils.sh
             - ../../build/setup_rpm
             - ../../build/build_rpm
+      - trigger-builds:
+        - project: "nfs-ganesha"
+          current-parameters: True
```
alimaredia: is this what I'll need to use $BRANCH from ceph build?

andrewschoen: This might work. If it doesn't you might need to use ``properties-file`` and have ``ceph-dev-build`` write a file with everything in it that ``nfs-ganesha`` might need.

PR #660, opened 2017-03-10
setup_pbuilder: fix boolean test

dmick:
three times


dmick:
Had this saved up from a while ago, forgot to push it.  Der.

PR #661, opened 2017-03-13
nfs-ganesha: add flavors and assing BRANCH to NFS_GANESHA_BRANCH

alimaredia:

### DIFF
--- a/nfs-ganesha/build/setup
+++ b/nfs-ganesha/build/setup
@@ -46,6 +46,9 @@ install_python_packages "pkgs[@]"
 
 NFS_GANESHA_BRANCH=`branch_slash_filter $NFS_GANESHA_BRANCH`
 CEPH_BRANCH=`branch_slash_filter $CEPH_BRANCH`
+BRANCH=${NFS_GANESHA_BRANCH}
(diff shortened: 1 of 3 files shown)


nfs-ganesha/build/setup
```diff
@@ -46,6 +46,9 @@ install_python_packages "pkgs[@]"
 
 NFS_GANESHA_BRANCH=`branch_slash_filter $NFS_GANESHA_BRANCH`
 CEPH_BRANCH=`branch_slash_filter $CEPH_BRANCH`
+BRANCH=${NFS_GANESHA_BRANCH}
+# set flavor as ceph branch libs are coming from
+FLAVOR=${CEPH_BRANCH}
```
alfredodeza: a couple of questions here: are you always going to have `$CEPH_BRANCH` defined? and if so, would it be good to prefix it with `ceph_` so that it is clear that, say, `master` isn't really `master` but Ceph's master branch?

alimaredia: `$CEPH_BRANCH` defaults to master for when nfs-ganesha's github triggers the build, That's not a bad idea I'm going to do that.

PR #662, opened 2017-03-14
nfs-ganesha: rm SSH credential

ktdreyer:
This job doesn't use SSH, so there's no need to make this credential available to the job.


ktdreyer:
cc @alimaredia in case he wants to review

PR #663, opened 2017-03-15
build_utils: enable SSL verification

ktdreyer:
All our chacra endpoints (https://*.chacra.ceph.com) use HTTPS certs
signed by LetsEncrypt now. There's no need to disable SSL verification
here any more.

This reverts f5e28bdeb675f6e245c26ae7d0e7f662ac1ed0f1.

Signed-off-by: Ken Dreyer <kdreyer@redhat.com>

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -110,7 +110,7 @@ make_chacractl_config () {
 url = "$url"
 user = "$CHACRACTL_USER"
 key = "$CHACRACTL_KEY"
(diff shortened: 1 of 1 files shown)


alfredodeza:
Would you mind changing to:

    ssl_verify = True

So that we still have a reference on the flag to change it back when we need to?

ktdreyer:
Sure, amended and re-pushed with explicitly setting this to `True`.

PR #665, opened 2017-03-16
ceph-pr-submodules: fix submodule check.

dmick:
Prior code never generated any output at all.

Fixes: http://tracker.ceph.com/issues/19244

### DIFF
--- a/ceph-pr-submodules/build/build
+++ b/ceph-pr-submodules/build/build
@@ -4,14 +4,14 @@ set -ex
 
 cd "$WORKSPACE"
 
-# The command below has access to the variables $name, $path, $sha1 and $toplevel:
-# $name is the name of the relevant submodule section in .gitmodules, $path is
-# the name of the submodule directory relative to the superproject, $sha1 is
-# the commit as recorded in the superproject, and $toplevel is the absolute
-# path to the top-level of the superproject.
-modified_submodules=$(git submodule foreach 'cd $toplevel; git diff --ignore-submodules=dirty origin/master -- $path')
+# show diffs between $ghprbTargetBranch (where the merge is going) and
+# $ghprbActualCommit (the tip of the branch that's merging) with '...',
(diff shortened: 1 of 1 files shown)


dmick:
Still need to test.

ceph-pr-submodules/build/build
```diff
@@ -4,12 +4,13 @@ set -ex
...
-# the commit as recorded in the superproject, and $toplevel is the absolute
-# path to the top-level of the superproject.
-modified_submodules=$(git submodule foreach 'cd $toplevel; git diff --ignore-submodules=dirty origin/master -- $path')
+# show diffs between $ghprbTargetBranch (where the merge is going) and
+# $ghprbActiveCommit (the tip of the branch that's merging) with '...',
+# which is equivalent to diff $(git merge-base TB AC) AC, or "show
+# diff from common ancestor of the target branch and this branch with the
+# tip of this branch".  With --submodule, also show detail of diff in submodules.
+
+modified_submodules="$(git diff --submodule=log $ghprbTargetBranch...$ghprbActiveCommit | grep ^Submodule)"
```
tchaikov: @ktdreyer seems that's the way to go !

ktdreyer: :+1: cool

dmick:
and indeed there were some bugs:

* ghprbActualCommit, not Active
* ghprbTargetBranch does not include origin/
* if grep fails, that causes the command to fail, which, because of -e, causes the script to fail
* it's safer and clearer to use -n "$var" than just $var

All fixed, good result at

https://jenkins.ceph.com/job/ceph-pr-submodules/5829/console

dmick:
@ktdreyer, if you'll do a final review, I've assigned you; thanks @tchaikov too

dmick:
@ktdreyer, you wanna take this one?

tchaikov:
lgtm also.

PR #667, opened 2017-03-23
reconfigure apt during every "pbuilder update"

ktdreyer:
The nfs-ganesha pbuilder setup code configures the pbuilder caches to
point at repos from shaman. Since all Jenkins jobs uses the same
pbuilder cache tarballs, this means that these shaman repositories will
still be enabled in the next jobs that run on a slave.

  1) nfs-ganesha job runs on a slave
  2) nfs-ganesha job configures pbuilder cache to use shaman repo
  3) nfs-ganesha job finishes
  4) ceph-build job runs on same slave slave
  5) ceph-build job uses same pbuilder cache with shaman repo configured

In addition to polluting the pbuilder environment, this can also lead to
pbuilder failures when shaman/chacra garbage-collects old repositories,
because Apt wants to download repo metadata from old ceph repos that
are no longer available.

Force every job that runs "pbuilder update" to also run with
"--override-config" in order to wipe out any apt repositories that may
have been configured in the cache.

### DIFF
--- a/ceph-build/build/setup_pbuilder
+++ b/ceph-build/build/setup_pbuilder
@@ -91,7 +91,8 @@ if [ -e $basedir/$DIST.tgz ]; then
     sudo pbuilder update \
     --basetgz $basedir/$DIST.tgz \
     --distribution $DIST \
-    --mirror "$mirror"
+    --mirror "$mirror" \
+    --override-config
 else
     echo building $DIST base.tgz
     sudo pbuilder create \
--- a/ceph-dev-build/build/setup_pbuilder
+++ b/ceph-dev-build/build/setup_pbuilder
@@ -91,7 +91,8 @@ if [ -e $basedir/$DIST.tgz ]; then
     sudo pbuilder update \
     --basetgz $basedir/$DIST.tgz \
     --distribution $DIST \
-    --mirror "$mirror"
+    --mirror "$mirror" \
+    --override-config
 else
     echo building $DIST base.tgz
     sudo pbuilder create \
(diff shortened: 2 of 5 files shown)


ktdreyer:
I noticed this problem with this build failure: https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=xenial,DIST=xenial,MACHINE_SIZE=huge/1992//consoleFull .

Only that one Xenial build failed: https://shaman.ceph.com/builds/ceph/jewel-integration-ktdreyer/8883933fbebb142f3edd2132fb46c212e1703fd7/

and it was during the `pbuilder_setup` script.

alfredodeza:
I wasn't aware that ``--override-config`` would wipe the apt cache. I tried looking for a reference of this behavior but I couldn't see anything relevant to the cache configuration. Could you elaborate a bit more on how you got to this solution?

ktdreyer:
Sure, the [pbuilder(8) man page](http://manpages.ubuntu.com/manpages/xenial/en/man8/pbuilder.8.html) explains it a bit:

```
--update
       Updates  the  base.tgz for the specified distribution.  Also, by
       specifying the  --override-config  option,  it  is  possible  to
       install  a  new apt-line using the given options and settings in
       the configuration file for the base.tgz.
```

Specifically, the `--override-config` argument [sets](https://anonscm.debian.org/git/pbuilder/pbuilder.git/tree/pbuilder-checkparams) `OVERRIDE_APTLINES=yes`, and pbuilder [calls](https://anonscm.debian.org/git/pbuilder/pbuilder.git/tree/pbuilder-modules) the `installaptlines()` function.

PR #668, opened 2017-03-28
ceph-docker-prs: Moving image{s} directories to shm

ErwanAliasr1:
On the CI, the root disk is pretty slow (20-30MB/sec).
This patch is about moving the big directories into /dev/shm to speedup the builds.
We can count up to 7GB of storage that doesn't it the cold storage but RAM (~9000MB/sec).

### DIFF
--- a/ceph-docker-prs/build/build
+++ b/ceph-docker-prs/build/build
@@ -15,6 +15,17 @@ sudo gpasswd -a ${USER} docker
 sudo systemctl restart docker
 newgrp docker
 
+# Moving docker/image & libvirt/images to shm to speedup CI
+if [ ! -e /dev/shm/images ]; then
+  sudo mv /var/lib/libvirt/images /dev/shm/
+  sudo ln -sf /dev/shm/images /var/lib/libvirt/
+fi
+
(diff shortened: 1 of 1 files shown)


ktdreyer:
FYI this will affect all jenkins.ceph.com jobs that use libvirt

ErwanAliasr1:
@ktdreyer Why that ? I'm only touching ceph-docker-prs.

alfredodeza:
@ErwanAliasr1 ceph-docker-prs uses nodes labeled as `vagrant&&libvirt` which other jobs use as well

ErwanAliasr1:
Oh you meant as we don't make the node dying just after the job, the next job using the same class of node could be affected, right ?

ktdreyer:
That's right!

ktdreyer:
I don't know if we needed to close this. It seems like a good performance speedup to me. Is there a way to make it configurable, so that Vagrant and libvirt use /dev/shm for some jobs?

ktdreyer:
Maybe create a second libvirt storage pool named "ram" or something, then configure Vagrant to use it with `storage_pool_name` ?

PR #680, opened 2017-04-18
ceph-installer-tests: do not send emails on job completion

andrewschoen:
We don't have a mail server setup, so lets remove this for now.

### DIFF
--- a/ceph-installer-tests/config/definitions/ceph-installer-tests.yml
+++ b/ceph-installer-tests/config/definitions/ceph-installer-tests.yml
@@ -55,7 +55,3 @@
           !include-raw-escape:
             - ../../../scripts/build_utils.sh
             - ../../build/build
-
-    publishers:
-      - email:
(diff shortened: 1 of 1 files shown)


alfredodeza:
hang on one second, I think I might get this working :) I promise

alfredodeza:
I worked a bit on getting this on the master instance, and although it can send emails out, the problem is that slaves want to either send email directly or connect to an SMTP server. If the SMTP server is on the master it means it should have authentication enabled, and SMTP authentication is not very easy to setup.

This doesn't error anymore, but it doesn't send emails either.

alfredodeza:
Adding more nightlies to Jenkins means this needs to be revisited. We need to have some sort of email notifications. @djgalloway do you have any ideas on how we could try to make this work?

djgalloway:
@alfredodeza - Refresh my memory?  What made us think the slaves were trying to send mail directly?

alfredodeza:
ceph-installer no longer exists, and emails are starting to go out

PR #681, opened 2017-04-19
adds a simple script that feeds bz counts into grafana

GregMeno:


bz-mill/build/teardown
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# There has to be a better way to do this than this script which just looks
+# for every Vagrantfile in scenarios and then just destroys whatever is left.
```
alfredodeza: you don't need this file

bz-mill/config/definitions/map_namespace_to_query
```diff
@@ -0,0 +1,2 @@
+http://red.ht/2oiunCH bz.gmeno.ceph.new.customer.24hrs
+http://red.ht/2paAywM bz.gmeno.ceph.any.needinfo.24hrs
```
alfredodeza: is `24hrs` a query or just part of the namespace? if the latter, you might want to skip it because you get that for free with gradana

GregMeno: it's saying that the BZ query is collecting ticket that have not changed in 24hrs

vasukulkarni: is 'gradana' brother for grafana? :running_man:

bz-mill/build/build
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+cat config/definitions/map_namespace_to_query |\
```
alfredodeza: this is not going to work because the definitions is not part of the job. For this to work the job itself would need to be cloned (you would need to define an scm section in the config) and then call into it.

PR #695, opened 2017-04-28
attempt to make libvirt slaves more stable

andrewschoen:
We continue to have issues with ceph-ansible/docker/installer tests failing because of libvirt networking issues. This is an attempt to get libvirt back into a good state by delete old networks and restarting services before we run tests.

Here is an example of the error we see often: https://jenkins.ceph.com/view/ceph-installer/job/ceph-installer-tests-ansible2.2-nightly_xenial/124/console

### DIFF
--- a/ceph-installer-tests/build/build
+++ b/ceph-installer-tests/build/build
@@ -6,4 +6,11 @@ WORKDIR=$(mktemp -td tox.XXXXXXXXXX)
 
 cd $WORKSPACE/tests/functional
 
+# Sometimes, networks may linger around, so we must ensure they are killed:
+networks=`sudo virsh net-list --all | grep active | egrep -v "(default|libvirt)" | cut -d ' ' -f 2`
+for network in $networks; do
+    sudo virsh net-destroy $network || true
+    sudo virsh net-undefine $network || true
+done
+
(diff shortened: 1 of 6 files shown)


andrewschoen:
If this doesn't work we might want to try running these tests on centos7 instead of xenial. We run similar tests in our downstream CI on centos7 and don't seem to hit these issues as often.

I also found this vagrant-libvirt issue that seems related: https://github.com/vagrant-libvirt/vagrant-libvirt/issues/510

alfredodeza:
Seems like all of these jobs are already sourcing the `build_utils.sh` script, since it is identical everywhere, maybe we could add it there and re-use?

andrewschoen:
@alfredodeza ok, I've made that change.

ceph-installer-tests/build/build
```diff
@@ -6,4 +6,7 @@ WORKDIR=$(mktemp -td tox.XXXXXXXXXX)
 
 cd $WORKSPACE/tests/functional
 
+clear_libvirt_networks()
+restart_libvirt_services()
```
alfredodeza: unless my BASH is terribly wrong (which can certainly be the case) I **think** that these functions don't need to have `()` otherwise it will get you a syntax error.

andrewschoen: You are right, it's my BASH that is terribly wrong.

PR #700, opened 2017-05-02
nfs-ganesha: fix rpm version finder to work on dev and release candid…

alimaredia:
…ates

### DIFF
--- a/nfs-ganesha/build/build_rpm
+++ b/nfs-ganesha/build/build_rpm
@@ -92,7 +92,7 @@ echo "\"\"\"" >> nfs-ganesha.cfg
 echo "Building RPMs"
 sudo mock --verbose -r nfs-ganesha.cfg --define "dist .el7" --resultdir=$WORKSPACE/dist/RPMS/ ${SRPM} || ( tail -n +1 $WORKSPACE/dist/RPMS/{root,build}.log && exit 1 )
 
-VERSION=`grep -om 1 '[[:digit:]]\+\.[[:digit:]]\+-dev-[[:digit:]]\+' $WORKSPACE/nfs-ganesha/build/include/config.h`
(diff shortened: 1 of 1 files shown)


nfs-ganesha/build/build_rpm
```diff
@@ -92,7 +92,7 @@ echo "\"\"\"" >> nfs-ganesha.cfg
 echo "Building RPMs"
 sudo mock --verbose -r nfs-ganesha.cfg --define "dist .el7" --resultdir=$WORKSPACE/dist/RPMS/ ${SRPM} || ( tail -n +1 $WORKSPACE/dist/RPMS/{root,build}.log && exit 1 )
 
-VERSION=`grep -om 1 '[[:digit:]]\+\.[[:digit:]]\+-dev-[[:digit:]]\+' $WORKSPACE/nfs-ganesha/build/include/config.h`
+VERSION=`grep -R "#define GANESHA_VERSION \"" $WORKSPACE/nfs-ganesha/build/include config.h | sed -e 's/#define GANESHA_VERSION "//1; s/"//1;'`
```
alfredodeza: is the path here part of the problem this is fixing? e.g. `nfs-ganesha/build/include` and `config.h` or is there a typo here with a missing `/` so that the path is `nfs-ganesha/build/include/config.g` ?

alimaredia:
@alfredodeza the issues that triggered this PR is that up until a little while ago the ganesha versions were all X.X-dev-X (ex: 2.5-dev-3). Recently the version has been X.X-rcX (ex: 2.5-rc3). The grep command was failing and causing all the jobs to fail because it couldn't find any versions with the "dev-X" syntax. 

This PR should work on both versions since it is searching for and piping the line GANESHA_VERSION is defined on and taking out all the text that isn't the version.

alfredodeza:
I see the slash got fixed :)

PR #713, opened 2017-05-12
ceph-docker-lint: exclude codes

leseb:
Excluding the following code:

* SC1091: since our entrypoint copies files in / during this shellcheck
the files can not be found. Basically we source ./file but presently the
file is under osd_directory/file so can not be found.
* SC2015: because we use 'set -e' we don't have other choice to use '||
true' here.
* SC2009: pgrep returns processes ID where we are looking for a specific
string in the process line.
* SC2001: our sed expressions are too complex we can not substitute them
with bash ${variable//search/replace}.

### DIFF
--- a/ceph-docker-lint/build/build
+++ b/ceph-docker-lint/build/build
@@ -4,16 +4,18 @@ set -e
 set -x
 
 IGNORE_THESE_CODES="SC1091,SC2015,SC2009,SC2001"
+IGNORE_THESE_FILES="variables_entrypoint.sh" # pipe-separated file names, e.g: foo|bar|foobar, this avoids shellcheck complaining that vars are not used (see: SC2034)
 
 function generate_filelist(){
    if [[ "$pull_request_id" -eq "" || "${ghprbCommentBody:-}" = "jenkins lint all" ]]
    then
-       find . -name '*.sh'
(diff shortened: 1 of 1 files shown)


alfredodeza:
This seems more like omitting codes that are indicating complexity/difficulty to get a pass. I am fine with merging this if the intent is to slowly make the changes needed to improve the situation, and not just set them as a new default.

For example on `SC2034`, the sourcing of variables could be done separately, by different paths, vs. having a big file where all of them are defined and everything is just sourced.

We believe that there is a lot of complexity in these BASH files that the linter is conveying into actual issues.

ktdreyer:
Can we use inline directives to disable the warnings on the
pre-existing problematic lines?
https://github.com/koalaman/shellcheck/wiki/Directive

Eg.

# shellcheck disable=SC2001

That way we don't error on today's code, but we avoid merging new PRs
with these issues in order to avoid incurring further tech debt.

alfredodeza:
I'll merge to allow this to move forward for now. I've opened:

https://bugzilla.redhat.com/show_bug.cgi?id=1450476
https://bugzilla.redhat.com/show_bug.cgi?id=1450480

To help track the work needed to get these checks back into this job

PR #720, opened 2017-05-17
scripts: Add function to delete leftover vagrant VMs

djgalloway:
Some of the network errors thrown by libvirt when running
Docker builds were because the slave had leftover Vagrant VMs that didn't
get destroyed.

This PR ensures all slaves being used for Docker tests will have VMs and disk images cleaned up if the build fails.

### DIFF
--- a/ceph-docker-nightly/build/teardown
+++ b/ceph-docker-nightly/build/teardown
@@ -11,3 +11,5 @@ for scenario in $scenarios; do
     vagrant destroy -f
     cd -
 done
+
+delete_vagrant_docker_vms
--- a/ceph-docker-prs/build/teardown
+++ b/ceph-docker-prs/build/teardown
@@ -11,3 +11,5 @@ for scenario in $scenarios; do
     vagrant destroy -f
     cd -
 done
+
+delete_vagrant_docker_vms
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -407,6 +407,19 @@ setup_pbuilder() {
     fi
 }
 
+delete_vagrant_docker_vms() {
+    # Delete any vagrant/libvirt VMs leftover from a failed docker build
+    libvirt_vms=`sudo virsh list --all --name | grep docker`
+    for vm in $libvirt_vms; do
+        # Destroy returns a non-zero rc if the VM's not running
+        sudo virsh destroy $vm || true
+        sudo virsh undefine $vm || true
+    done
+    # Clean up any leftover disk images
+    sudo rm -f /var/lib/libvirt/images/docker*.img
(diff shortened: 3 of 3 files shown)


andrewschoen:
This looks like it'd be helpful for all the jobs that use vagrant/libvirt. So, ceph-ansible-pr, ceph-ansible-scenario and ceph-installer-tests. Can we expand this to those jobs as well?

djgalloway:
> This looks like it'd be helpful for all the jobs that use vagrant/libvirt. So, ceph-ansible-pr, ceph-ansible-scenario and ceph-installer-tests. Can we expand this to those jobs as well?

We could and that's what I'd originally planned on doing but @alfredodeza posited that deleting *all* libvirt VMs and disk images could be too destructive.  At least with grepping for 'docker' we know we're only deleting docker-related VMs and libvirt images.

andrewschoen:
@djgalloway ok, I guess that makes sense. I doubt we'd ever hit that on these CI machines but you never know. Maybe if we grep for ``ansible`` it'll work in both cases? I think those are named after the scenario so they would all have the keyword ``ansible`` in them.

djgalloway:
@andrewschoen I'm not sure grepping for `ansible` will work.  I just found these leftover from a build on an OVH node.

```
 -     cluster_client0_1495036387_7cc5add861b1b43083cc shut off
 -     cluster_mds0_1495036388_70f0a27fd36998dc22c6 shut off
 -     cluster_mgr0_1495036386_330825b9730a159ce475 shut off
 -     cluster_mon0_1495036387_4ac7972357b0927db79a shut off
 -     cluster_mon1_1495036389_43d743984a62d1cd97db shut off
 -     cluster_mon2_1495036388_5df76e153d8caafd1cd1 shut off
 -     cluster_osd0_1495036388_af350d3262e440527b3e shut off
 -     cluster_rgw0_1495036389_978ce1f9ff310b09bdb0 shut off
```

I'm kinda of the opinion that if we want a fresh environment on any slave with a libvirt slave that libvirt should be in a pristine condition and no other libvirt VMs or networks should be present.

In other words, I think we should destroy and undefine all VMs, disks, and networks.

@alfredodeza thoughts?

andrewschoen:
> In other words, I think we should destroy and undefine all VMs, disks, and networks.

I'm +1 on this. The scripts that destroy the VMS, disks and networks are all part of the jenkins job so it shouldn't affect people running tests manually.

alfredodeza:
Thanks for this work @djgalloway

alfredodeza:
Errr, yes, we need a giant hammer and obliterate everything and anything at this point. So I guess this now needs another PR. We need to have some fireworks when this happens as well

hxihgck

PR #721, opened 2017-05-17
samba: remove inactive distros and erroneous variable definition

dmick:


dmick:
wrong branch.

PR #722, opened 2017-05-17
add ceph-docker-build job

ktdreyer:
Build docker images for every Git branch in https://github.com/ceph/ceph-docker .

Use the credentials-id plugin to authenticate to the Docker Hub and push images.

### DIFF
--- a/ceph-docker-build/build/build
+++ b/ceph-docker-build/build/build
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -eux
+
+sudo yum -y install docker
+
+sudo systemctl start docker
+
+# Authenticate to Docker Hub
+sudo docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
+
+# Copy the files to the root for the type of image we're going to build
+mkdir -p daemon demo
+cp -Lrv ceph-releases/"$CEPH_RELEASE"/centos/7/daemon/* daemon
+# TODO: make demo image optional in build_imgs.sh
+# (instead of building it here and throwing it away)
+cp -Lrv ceph-releases/"$CEPH_RELEASE"/centos/7/demo/* demo
+
+# TODO: manipulate the Dockerfile here to insert the chacra Yum repo?
+# Proj Atomic does that with https://github.com/DBuildService/dockerfile-parse
+
+sudo "$WORKSPACE"/travis-builds/build_imgs.sh
+
+# Show our new images, for debugging:
+sudo docker images
+
+REGISTRY=docker.io
+
+# Determine our image's tag for the registry:
+BRANCH=${GIT_BRANCH//origin\//}
+TAG=ci-build-${BRANCH}-${CEPH_RELEASE}-centos-7
+
+# Tag the image we just built for the registry:
+sudo docker tag ceph/daemon "${REGISTRY}/cephjenkins/daemon:${TAG}"
+
+# Apparently this avoids a race condition between the tagging and the push
+# which causes this to sometimes fail when run by Jenkins
+sleep 1
+sudo docker --debug push "${REGISTRY}/cephjenkins/daemon:${TAG}"
+
+# Clean up registry credentials
+sudo docker logout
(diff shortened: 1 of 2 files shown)


ceph-docker-build/config/definitions/ceph-docker-build.yml
```diff
@@ -0,0 +1,49 @@
...
+    properties:
+      - github:
+          url: https://github.com/ceph/ceph-docker
+    logrotate:
+      daysToKeep: 15
+      numToKeep: 30
+
+    triggers:
+      - github
+      #- parameterized-timer: '@daily % CEPH_RELEASE=kraken'
```
alfredodeza: why is this commented out?

ceph-docker-build/build/build
```diff
@@ -0,0 +1,40 @@
...
+BRANCH=${GIT_BRANCH//origin\//}
+TAG=tag-build-${BRANCH}-${CEPH_RELEASE}-centos-7
+
+# Tag the image we just built for the registry:
+sudo docker tag ceph/daemon "cephjenkins/daemon:${TAG}"
+
+# Apparently this avoids a race condition between the tagging and the push
+# which causes this to sometimes fail when run by Jenkins
+sleep 1
+sudo docker --debug push "cephjenkins/daemon:${TAG}"
```
alfredodeza: are we going to push to docker hub for these? or is the registry somewhere else? (I don't see it defined in here)

ktdreyer:
Images are pushed here: https://hub.docker.com/r/cephjenkins/daemon/tags/

ceph-docker-build/config/definitions/ceph-docker-build.yml
ktdreyer: Whoops, this was an earlier thing, left over from debugging. I've dropped it and amended this.

ceph-docker-build/build/build
ktdreyer: Yes, when no server is specified in the tag name here, docker defaults to operating on the hub (docker.io). I've amended the commit to make this explicit

ceph-docker-build/build/build
```diff
@@ -0,0 +1,42 @@
...
+sudo "$WORKSPACE"/travis-builds/build_imgs.sh
+
+# Show our new images, for debugging:
+sudo docker images
+
+REGISTRY=docker.io
+
+# Determine our image's tag for the registry:
+BRANCH=${GIT_BRANCH//origin\//}
+TAG=tag-build-${BRANCH}-${CEPH_RELEASE}-centos-7
```
andrewschoen: Should we prefix this with something like ``ci-build`` instead of ``tag-build``?

ceph-docker-build/config/definitions/ceph-docker-build.yml
```diff
@@ -0,0 +1,46 @@
...
+    logrotate:
+      daysToKeep: 15
+      numToKeep: 30
+
+    triggers:
+      - github
+
+    parameters:
+      - string:
+          name: CEPH_RELEASE
```
andrewschoen: What was the intention with this param? I think it'd be nice to get a jewel and kraken container for each branch push.

ktdreyer: Exactly, the intention was to do more than just kraken.

Thinking about this, we could probably just parameterize this job in JJB instead.

andrewschoen: I think using JJB to parameterize this job would be a good idea. Then we could have separate jobs building different versions of ceph.

ktdreyer: I've amended and re-pushed as a parameterized job now, so we build the jewel, kraken, and luminous containers.

ceph-docker-build/build/build
ktdreyer: Good call on `ci-build`, I've amended this commit.

ceph-docker-build/config/definitions/ceph-docker-build.yml
andrewschoen: hmm, I don't think ceph-docker can make a luminous container though can it?

leseb:
Sorry for the comment, not sure why I thought you just submitted that.
Let's close this since the new architecture can handle that use case.

ktdreyer:
leseb thanks for clarifying this.

PR #725, opened 2017-05-18
samba-trigger: Trigger build on commits

b-ranto:
This will trigger a samba build on push to any of master* ceph* or wip-*
branches.


b-ranto:
I have updated the PR to make samba builds more aligned to nfs-ganesha builds as per request by @jcsp.

PR #727, opened 2017-05-18
samba: Clean old ceph bits before getting new

b-ranto:
Debian/Ubuntu does not tie the package versions the same way the rpm
packagin does so upgrading a single package won't remove any old
packages. Remove the old bits before installing any new.

This should fix a build failure on xenial where there were two libcephfs
libraries present on the system at the time of build which broke the
build.

### DIFF
--- a/samba/build/build_deb
+++ b/samba/build/build_deb
@@ -34,8 +34,14 @@ fi
 sudo cp $WORKSPACE/shaman.list /etc/apt/sources.list.d/
 
 ## Install any setup-time deps
-# Make sure we use the latest version of libcephfs-dev
-sudo apt-get remove -y libcephfs-dev || true
+# Make sure we use the latest ceph versions, remove any old bits
+xargs sudo apt-get remove -y <<< "
+libcephfs-dev
+libcephfs1
+libcephfs2
+librados2
+librbd1
+"
 
 # We need these for the build
 sudo apt-get update && sudo apt-get install -y build-essential equivs libgnutls-dev libacl1-dev libldap2-dev ruby ruby-dev libcephfs-dev


samba/build/build_deb
```diff
@@ -34,8 +34,13 @@ fi
...
 ## Install any setup-time deps
-# Make sure we use the latest version of libcephfs-dev
-sudo apt-get remove -y libcephfs-dev || true
+# Make sure we use the latest ceph versions, remove any old bits
+xargs sudo apt-get remove -y <<< "
+libcephfs-dev
+libcephfs1
+libcephfs2
+librados2
+"
```
alfredodeza: we've seen issues with librdb1 as well (unsure if we now have a librbd2?)

https://github.com/ceph/ceph-deploy/blob/master/ceph_deploy/install.py#L307-L308

b-ranto: Added and re-pushed, afaik there is no librbd2, it was only libcephfs that was bumped to version 2 recently.

ktdreyer:
> Debian/Ubuntu does not tie the package versions the same way the rpm
> packagin does so upgrading a single package won't remove any old
> packages

This is a surprise to me, because this should definitely be happening now in jewel and newer. Can you please provide reproduction steps, so we can fix this in the debian packaging?

tchaikov:
@ktdreyer @b-ranto what we need is probably `apt-get autoremove`?

b-ranto:
@ktdreyer well, I removed libcephfs-dev because I wanted to make sure I get a new one. However, the version that was installed still shipped libcephfs1. Then I installed libcephfs-dev from latest master build which installed libcephfs2 so I had both libcephfs1 and libcephfs2 installed. However, the build failed (on linking) so I suppose it tried to link against libcephfs1 which did not contain the necessary symbols.

tchaikov:
@b-ranto libcepfs-dev offers `usr/lib/libcephfs.so` which is actually libcephfs2.so. so, if we are linking against `libcephfs` we should be not be linking against libcephfs1. am i right?

b-ranto:
@tchaikov maybe it was a bug in the samba build system, I did not investigate much further as removing the old package helped

PR #729, opened 2017-05-18
samba: Strip out origin/ for CEPH_BRANCH

b-ranto:
When we pass the CEPH_BRANCH from ceph-dev-build, it contains
origin/<branch>, we need to strip out the remote from the CEPH_BRANCH
before we can use it.

### DIFF
--- a/samba/build/setup
+++ b/samba/build/setup
@@ -38,6 +38,10 @@ export LC_ALL=C # the following is vulnerable to i18n
 pkgs=( "chacractl>=0.0.4" )
 install_python_packages "pkgs[@]"
 
+SAMBA_BRANCH=$(branch_slash_filter $SAMBA_BRANCH)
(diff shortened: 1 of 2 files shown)


samba/build/build_deb
```diff
@@ -7,6 +7,9 @@ if test "$DISTRO" != "debian" -a "$DISTRO" != "ubuntu"; then
 fi
 
 ## Get the desired CEPH_BRANCH/CEPH_SHA1 ceph repo
+# Filter out origin/
+CEPH_BRANCH=$(basename "$CEPH_BRANCH")
```
andrewschoen: we have a method for this in build_utils.sh https://github.com/ceph/ceph-build/blob/master/scripts/build_utils.sh#L8

b-ranto:
Updated to use the build_utils function and fix a rename from BRANCH to SAMBA_BRANCH.

PR #733, opened 2017-05-25
nfs-ganesha/samba: disable ceph triggered builds of nfs-ganesha & samba

alimaredia:

### DIFF
--- a/ceph-dev-build/config/definitions/ceph-dev-build.yml
+++ b/ceph-dev-build/config/definitions/ceph-dev-build.yml
@@ -80,15 +80,6 @@
             - ../../build/setup_rpm
             - ../../build/build_rpm
       - trigger-builds:
-        - project: "nfs-ganesha"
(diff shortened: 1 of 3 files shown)


alfredodeza:
can  you just disable this job for now? e.g.

    disabled: true

alimaredia:
@alfredodeza take a look at it now

alfredodeza:
@alimaredia you need to disable for the jobs that 'trigger' the builds. Yes: disabling will prevent you from doing that. I thought that for samba we had 'trigger' jobs?

alimaredia:
Changed the title so that this PR isn't merged unless the decision is made to disable those builds.

alimaredia:
@alfredodeza removed the DNM tag, this is good to merge on my end

alfredodeza:
@alimaredia you need a rebase I think

alfredodeza:
I think we are good with this now that they are time-based

PR #743, opened 2017-06-15
ceph-docs/build/build: don't try to skip if not a merge commit

dmick:
This touches on http://tracker.ceph.com/issues/19976, but doesn't
fix it; it just defaults to "build it if it's not a merge commit".
I'm still not sure the ideas in the bug are workable.  So this
mitigates the issue.

### DIFF
--- a/ceph-docs/build/build
+++ b/ceph-docs/build/build
@@ -1,21 +1,26 @@
 #!/bin/bash
+# vim: ts=4 sw=4 expandtab
 
 set -ex
 
 # did any doc/ files change?  
-# This assumes $GIT_COMMIT is a merge commit (it is usually, because that's how
-# we manage ceph, with PRs that include merges).  So, noting the three dots,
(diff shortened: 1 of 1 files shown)


dmick:
This looks much bigger than it is, because I reformatted a bit as well.

ceph-docs/build/build
```diff
@@ -1,21 +1,26 @@
...
-       echo "No doc files changed, skipping build"
-       exit 0
+# only (--name-only).
+#
+# Skip this optimization for non-merge commits
+
+if git rev-parse --verify ${GIT_COMMIT}^2; then
+    # enclosing doublequotes prevent newlines from disappearing, for grep below
+    files="$(git diff --name-only ${GIT_COMMIT}^1...${GIT_COMMIT}^2)"
+    echo -e "changed files:\n$files"
```
alfredodeza: thanks for adding this

PR #744, opened 2017-06-16
ceph-medic-pull-requests: adds a job to test ceph-medic PRs

andrewschoen:

### DIFF
--- a/ceph-medic-pull-requests/build/build
+++ b/ceph-medic-pull-requests/build/build
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# the following two methods exist in scripts/build_utils.sh
+pkgs=( "tox" )
+install_python_packages "pkgs[@]"
+
+cd "$WORKSPACE/ceph-medic"
+
+$VENV/tox -rv
--- a/ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
+++ b/ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
@@ -1,0 +1,64 @@
+- scm:
+    name: ceph-medic 
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph-medic
+          branches:
+            - ${sha1}
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
+          basedir: "ceph-medic"
+          skip-tag: true
+          wipe-workspace: true
+
+
+- job:
+    name: ceph-medic-pull-requests
+    description: Runs tox tests for ceph-medic on each GitHub PR
+    project-type: freestyle
+    node: python3
+    block-downstream: false
+    block-upstream: false
+    defaults: global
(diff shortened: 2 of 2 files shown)


ceph-medic-pull-requests/build/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# the following two methods exist in scripts/build_utils.sh
+pkgs=( "ansible" "tox" )
```
alfredodeza: I don't think you need ansible here

ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
```diff
@@ -0,0 +1,64 @@
...
+          admin-list:
+            - alfredodeza
+            - dmick
+            - ktdreyer
+            - andrewschoen
+          org-list:
+            - ceph
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: false
```
alfredodeza: lets `permit-all: true` so that we aren't whitelisting ? The `org-list: ceph` doesn't seem to do what we would want here (which is, whitelist everyone in the ceph org)

ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
```diff
@@ -0,0 +1,64 @@
...
+          basedir: "ceph-medic"
+          skip-tag: true
+          wipe-workspace: true
+
+
+- job:
+    name: ceph-medic-pull-requests
+    description: Runs tox tests for ceph-medic on each GitHub PR
+    project-type: freestyle
+    node: trusty && small
```
alfredodeza: you will want `python3` here, and possibly only that, so that whatever is available with that label can work

ceph-medic-pull-requests/build/build
andrewschoen: You're right, will fix.

ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
andrewschoen: There is a ``py27`` tox environment as well, why would we not test that as well?

ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
andrewschoen: Sure, I'll do that.

ceph-medic-pull-requests/config/definitions/ceph-medic-pull-requests.yml
alfredodeza: we already have python 2 in the slaves that have python 3 installed, so you would be able to test both

andrewschoen: I misunderstood, I thought you meant to test only ``python3``.

PR #746, opened 2017-06-20
nfs-ganesha: nfs-ganesha for download.ceph.com

alimaredia:

### DIFF
--- a/nfs-ganesha-stable/build/build_rpm
+++ b/nfs-ganesha-stable/build/build_rpm
@@ -1,0 +1,136 @@
+#! /usr/bin/bash
+set -ex
+
+# Only do actual work when we are an RPM distro
+if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; then
+    exit 0
+fi
+
+## Get some basic information about the system and the repository
+RELEASE="$(lsb_release --short -r | cut -d '.' -f 1)" # sytem release
+
+# Get .repo file from appropriate shaman build
+REPO_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$CEPH_SHA1/$DISTRO/$RELEASE/flavors/default/repo"
+TIME_LIMIT=1200
+INTERVAL=30
+REPO_FOUND=0
+
+# poll shaman for up to 10 minutes
+while [ "$SECONDS" -le "$TIME_LIMIT" ]
+do
+  if `curl --fail -L $REPO_URL > $WORKSPACE/shaman.repo`; then
+    echo "Ceph repo file has been added from shaman"
+    REPO_FOUND=1
+    break
+  else
+    sleep $INTERVAL
+  fi
+done
+
+if [[ "$REPO_FOUND" -eq 0 ]]; then
+  echo "Ceph lib repo does NOT exist in shaman"
+  exit 1
+fi
+
+# add shaman repos to /etc/yum.repos.d/ to install ceph libraries so enable
+#FSAL_CEPH (enabled by default) and FSAL_RGW in the .spec file when cmake command runs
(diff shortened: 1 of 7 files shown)


nfs-ganesha-stable/build/setup
```diff
@@ -0,0 +1,56 @@
...
+install_python_packages "pkgs[@]"
+
+NFS_GANESHA_BRANCH=`branch_slash_filter $NFS_GANESHA_BRANCH`
+CEPH_BRANCH=`branch_slash_filter $CEPH_BRANCH`
+BRANCH=${NFS_GANESHA_BRANCH}
+# set flavor as ceph branch libs are coming from
+FLAVOR="ceph_${CEPH_BRANCH}"
+
+# ask shaman which chacra instance to use
+chacra_url="https://chacra.ceph.com/"
```
alimaredia: @alfredodeza i hardcoded chacra.ceph.com here

nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
```diff
@@ -0,0 +1,150 @@
...
+    block-upstream: false
+    properties:
+      - github:
+          url: https://github.com/nfs-ganesha/nfs-ganesha
+    concurrent: true
+    parameters:
+      - string:
+          name: NFS_GANESHA_BRANCH
+          description: "The git branch (or tag) to build"
+          default: "V2.5-stable"
```
alimaredia: @alfredodeza I also hardcoded the NFS_GANESHA_BRANCH and CEPH_BRANCH for the binaries I want out at the moment, but these values will obviously change over time. Any idea on what I should set them too?

nfs-ganesha-stable/build/build_rpm
```diff
@@ -0,0 +1,132 @@
...
+libblkid-devel
+libcap-devel
+libnfsidmap-devel
+libwbclient-devel
+krb5-devel
+librgw-devel
+libcephfs-devel
+"
+
+# 1-17-17  will remove on next push of EPEL 7 stable, can remove in a couple days
```
alfredodeza: what is this?

alimaredia: something that should have been removed when I copied everything over from the nfs-ganesha job.

nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
alfredodeza: I think it is fine to default them like you do here, but I would be concerned about defaults that might have issues like using the 'latest' sha1 from a Ceph build, because I am assuming you will want to build against a released sha1, not whatever is the latest?

nfs-ganesha-stable/build/setup
alfredodeza: yep, that is fine

nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
alimaredia: yes, would it make sense for me to hardcode the sha1. At this point we don't have anything for luminous correct?

alfredodeza: it isn't really hard coding, just defaulting to one if it is not defined, which I think is good for this use case. We do have SHA1s for luminous that are released but those are "rc" releases.

These are the current SHA1s that we've released: https://chacra.ceph.com/r/ceph/luminous/

alimaredia: I've used the latest SHA1 as the value for CEPH_SHA1 right below this variable.

alfredodeza:
@alimaredia this needs a rebase

alimaredia:
@alfredodeza rebased!

PR #748, opened 2017-06-21
cephmetrics: Add new build job

b-ranto:

### DIFF
--- a/cephmetrics/build/build_rpm
+++ b/cephmetrics/build/build_rpm
@@ -1,0 +1,70 @@
+#! /usr/bin/bash
+set -ex
+
+# Only do actual work when we are an RPM distro
+if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; then
+    exit 0
+fi
+
+
+## Install any setup-time deps (to make dist package)
+
+# We need this to get the major version from lsb_release
+yum install -y redhat-lsb-core mock git
+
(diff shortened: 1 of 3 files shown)


cephmetrics/build/build_rpm
```diff
@@ -0,0 +1,70 @@
...
+RELEASE="$(lsb_release --short -r | cut -d '.' -f 1)" # sytem release
+VERSION="0.1"
+REVISION="$(git rev-list --count HEAD)-g$(git rev-parse --short HEAD)"
+RPM_RELEASE=$(echo $REVISION | tr '-' '_') # the '-' has a special meaning
+
+
+## Build the source tarball
+echo "Building source distribution"
+git archive --format=zip --prefix=cephmetrics-${VERSION}/ HEAD > dist/cephmetrics-${VERSION}.zip
+wget https://grafana.com/api/plugins/vonage-status-panel/versions/1.0.4/download -O dist/vonage-status-panel-1.0.4.zip
```
alfredodeza: this seems to be for actual releases right? in that case this `wget` here worries me, can we get some comments here as to why we are doing this like that? what happens in this case with this zip file?

b-ranto: It is another source, it is for a grafana plugin called vonage-status-panel that cephmetrics requires. I did not want to split this into two packages so I just renamed the module to a unique name (cephmetrics-vonage) to avoid conflicts and packaged it in a separate subpackage.

PR #755, opened 2017-06-22
samba: Cleanup added yum/apt repos

zmc:

### DIFF
--- a/samba/config/definitions/samba.yml
+++ b/samba/config/definitions/samba.yml
@@ -110,3 +110,10 @@ If this is checked, then the binaries will be built and pushed to chacra even if
       - inject-passwords:
           global: true
           mask-password-params: true
+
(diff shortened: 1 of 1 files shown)


samba/config/definitions/samba.yml
```diff
@@ -110,3 +110,10 @@ If this is checked, then the binaries will be built and pushed to chacra even if
       - inject-passwords:
           global: true
           mask-password-params: true
+
+    publishers:
+      - postbuildscript:
+          script-only-if-succeeded: False
+          script-only-if-failed: True
```
andrewschoen: This needs to be ``script-only-if-failed: False`` so it runs regardless of build success of failure.

zmc: Of course!

PR #758, opened 2017-06-23
Add ceph-pull-requests-arm64

dmick:
This has several XML sections; removing the two smaller ones seems safe,
but I'm concerned about the last one (the GitSCM clone behavior).  It
doesn't currently look like either jjb or jjw handle that option section
correctly, so I'm leaving the XML for now.

### DIFF
--- a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
+++ b/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
@@ -1,0 +1,92 @@
+- job:
+    block-downstream: false
+    block-upstream: false
+    builders:
+    - shell:
+        !include-raw:
+        - ../../build/build
+    concurrent: true
+    disabled: false
+    name: !!python/unicode 'ceph-pull-requests-arm64'
+    node: arm64
+    parameters:
+    - string:
+        default: origin/master
+        description: A pull request ID, like 'origin/pr/72/head'
+        name: sha1
+    project-type: freestyle
+    properties:
+    - build-discarder:
+        artifact-days-to-keep: -1
+        artifact-num-to-keep: -1
+        days-to-keep: 15
+        num-to-keep: 300
+    - raw:
+        xml: |
+          <com.sonyericsson.jenkins.plugins.bfa.model.ScannerJobProperty plugin="build-failure-analyzer@1.18.1">
+          <doNotScan>false</doNotScan>
+          </com.sonyericsson.jenkins.plugins.bfa.model.ScannerJobProperty>
+    - github:
+        url: https://github.com/ceph/ceph/
+    - raw:
+        xml: |
+          <com.sonyericsson.rebuild.RebuildSettings plugin="rebuild@1.25">
+          <autoRebuild>false</autoRebuild>
+          <rebuildDisabled>false</rebuildDisabled>
+          </com.sonyericsson.rebuild.RebuildSettings>
+    publishers: []
+    quiet-period: '5'
+    retry-count: '3'
+    scm:
+    - raw:
+        xml: |
+          <scm class="hudson.plugins.git.GitSCM" plugin="git@3.1.0">
+          <configVersion>2</configVersion>
+          <userRemoteConfigs>
+          <hudson.plugins.git.UserRemoteConfig>
+          <name>origin</name>
+          <refspec>+refs/pull/*:refs/remotes/origin/pr/*</refspec>
+          <url>https://github.com/ceph/ceph.git</url>
+          </hudson.plugins.git.UserRemoteConfig>
+          </userRemoteConfigs>
+          <branches>
+          <hudson.plugins.git.BranchSpec>
+          <name>${sha1}</name>
+          </hudson.plugins.git.BranchSpec>
+          </branches>
+          <doGenerateSubmoduleConfigurations>false</doGenerateSubmoduleConfigurations>
+          <submoduleCfg class="list" />
+          <extensions>
+          <hudson.plugins.git.extensions.impl.CheckoutOption>
+          <timeout>20</timeout>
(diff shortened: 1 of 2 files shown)


ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -0,0 +1,100 @@
...
+        NPROC=$(nproc)
+        testnproc=$(($NPROC / 4))
+        #export CHECK_MAKEOPTS="-R unittest* -j${testnproc}"
+        export CHECK_MAKEOPTS="-j${testnproc}"
+        export BUILD_MAKEOPTS="-j${NPROC}"
+        ./run-make-check.sh
+        sleep 5
+        ps -ef | grep ceph || true
+    concurrent: true
+    description: <!-- Managed by Jenkins Job Builder -->
```
alfredodeza: I don't think you need this line, this is automatically done for you

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -0,0 +1,100 @@
...
+    builders:
+    - shell: |-
+        bash -ex
+        NPROC=$(nproc)
+        testnproc=$(($NPROC / 4))
+        #export CHECK_MAKEOPTS="-R unittest* -j${testnproc}"
+        export CHECK_MAKEOPTS="-j${testnproc}"
+        export BUILD_MAKEOPTS="-j${NPROC}"
+        ./run-make-check.sh
+        sleep 5
```
alfredodeza: we usually don't have shell directives that are longer than 1-2 lines. When that is the case we create a script in <job name>/build/build and then pull it in. Makes it easier to read too. The way you do that is like:

      - shell:
          !include-raw:
            - ../..//build/build

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -0,0 +1,100 @@
...
+    - raw:
+        xml: |
+          <com.sonyericsson.jenkins.plugins.bfa.model.ScannerJobProperty plugin="build-failure-analyzer@1.18.1">
+          <doNotScan>false</doNotScan>
+          </com.sonyericsson.jenkins.plugins.bfa.model.ScannerJobProperty>
+    - github:
+        url: https://github.com/ceph/ceph/
+    - raw:
+        xml: |
+          <com.sonyericsson.rebuild.RebuildSettings plugin="rebuild@1.25">
```
alfredodeza: why aren't you allowing this to be rebuilt? Or was that just something that trickled down from the UI?

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -0,0 +1,100 @@
...
+          <autoRebuild>false</autoRebuild>
+          <rebuildDisabled>false</rebuildDisabled>
+          </com.sonyericsson.rebuild.RebuildSettings>
+    publishers: []
+    quiet-period: '5'
+    retry-count: '3'
+    scm:
+    - raw:
+        xml: |
+          <scm class="hudson.plugins.git.GitSCM" plugin="git@3.1.0">
```
alfredodeza: I don't see any unusual settings here that wouldn't translate to what I think you want in the job.

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -0,0 +1,100 @@
...
+        auth-id: 2f99aa21-4f3c-4a7e-93d9-f85361aff591
+        auto-close-on-fail: false
+        build-desc-template: null
+        cron: null
+        github-hooks: true
+        only-trigger-phrase: false
+        org-list:
+        - ceph
+        permit-all: true
+        trigger-phrase: null
```
alfredodeza: you may want to include a trigger phrase just for this, that way, with a comment on Github you can re-trigger it. For example:

    jenkins test arm64

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
dmick: Shrug.  OK.

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
dmick: will remove

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
dmick: I think this is "Rebuild Without Asking For Parameters" and "Disable Rebuilding for this job", both of which are unchecked (false), which I think is the default.  The help, helpfully, does not refer to the plugin, so I can't really be sure.

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
dmick: Yeah, it's just hudson.plugins.git.extensions.impl.CloneOption that's not supported by either jjw or jjb.

ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
dmick: yeah, I didn't understand what it did, so I didn't set anything.

dmick:
pushed a review-only commit; lmk if this looks OK and if so I'll squash

dmick:
repushed.  also rebased.

PR #761, opened 2017-06-26
Add cephmetrics-pull-requests

zmc:
Runs tox tests!

### DIFF
--- a/cephmetrics-pull-requests/config/definitions/cephmetrics-pull-requests.yml
+++ b/cephmetrics-pull-requests/config/definitions/cephmetrics-pull-requests.yml
@@ -1,0 +1,54 @@
+- job:
+    name: cephmetrics-pull-requests
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    display-name: 'Cephmetrics: Pull Requests'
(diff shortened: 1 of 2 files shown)


cephmetrics-pull-requests/config/definitions/cephmetrics-pull-requests.yml
```diff
@@ -0,0 +1,54 @@
...
+          admin-list:
+            - zmc
+            - pcuzner
+            - b-ranto
+            - GregMeno
+          org-list:
+            - ceph
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: false
```
alfredodeza: this will restrict tests to only the ones listed in the `admin-list`, For example, for ceph.git we `permi-all: true`

zmc: Oh, good to know. I'm not actually sure which we want here, but I suppose we could leave this as-is until we need to change it (if we ever do)

zmc:
Thanks @alfredodeza - I'll merge this and I can toggle `permit-all` if necessary later.

PR #766, opened 2017-06-27
DNM: cephmetrics: Build for all branches

zmc:
This kicked off a build, but it failed because `BRANCH` was unset :-/

https://jenkins.ceph.com/job/cephmetrics/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=small/42/console

### DIFF
--- a/cephmetrics/build/build_rpm
+++ b/cephmetrics/build/build_rpm
@@ -4,6 +4,8 @@ set -ex
 # Sanity-check:
 [ -z "$GIT_BRANCH" ] && echo Missing GIT_BRANCH variable && exit 1
 
+BRANCH=`branch_slash_filter $GIT_BRANCH`
+
 # Only do actual work when we are an RPM distro
 if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; then
     exit 0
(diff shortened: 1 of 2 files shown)


b-ranto:
That line just defines the default value of BRANCH argument and we need it, otherwise (with empty BRANCH) the build will fail.

zmc:
@andrewschoen thanks!! It's doing what I want now: https://jenkins.ceph.com/job/cephmetrics/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=small/48/console

alfredodeza:
arghhhh, you had DNM in there @zmc and I just merged :( hopefully I am not doing something super destructive?

zmc:
No worries @alfredodeza, I had just force-pushed with the commits I'd tested successfully. Only damage here is the title of the PR

@andrewschoen @b-ranto I'm not sure if it's worth removing all the parameters and matrix-related stuff, or if we might want to try to get it working later. I'm happy to follow-up as necessary.

b-ranto:
I would not worry about removing the stuff since it is already there. If we ever want to do a deb build, we have the basics there.

However, looking at the patch, it looks like we will be building for master only or at least the manual override for BRANCH won't work :-/.

PR #769, opened 2017-06-29
cephmetrics-release: when building for centos7 upload to chacra as rhel7

andrewschoen:
We need to do this because cephmetrics-deps is uploaded as rhel7 and if
we don't upload cephmetrics as the same chacra won't combine the repos.


andrewschoen:
We're taking a different approach, this isn't needed anymore.

PR #782, opened 2017-07-07
ceph-medic-tests: a job to run nightly tests for ceph-medic

andrewschoen:
This uses a job-template so as we build more testings scenarios
they can be added here easily.


andrewschoen:
This needs https://github.com/ceph/ceph-medic/pull/41

PR #794, opened 2017-07-26
ceph-ansible-prs: Change group for jenkins_user

guits:
Required to switch to Ubuntu 16.10.
The group `libvirtd` have been renamed to `libvirt`.

### DIFF
--- a/ansible/slave_libvirt.yml
+++ b/ansible/slave_libvirt.yml
@@ -48,10 +48,18 @@
         - vagrant
       when: ansible_pkg_mgr  == "apt"
 
+    - set_fact:
+        jenkins_group: 'libvirtd'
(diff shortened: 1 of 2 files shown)


ansible/slave_libvirt.yml
```diff
@@ -51,7 +51,7 @@
     - name: create a {{ jenkins_user }} user
       user:
         name: "{{ jenkins_user }}"
-        groups: libvirtd
+        groups: libvirt
```
alfredodeza: since we have a mix of machines that may or may not have this group I think you should check either the distro version or just try both ways ignoring errors

guits: @alfredodeza It should be ok that way

PR #801, opened 2017-07-26
Add option to skip build for qa scripts/yaml changes

vasukulkarni:
pattern to skip : qa tests .* skip build

Signed-off-by: Vasu Kulkarni <vasu@redhat.com>

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -29,6 +29,7 @@
           org-list:
             - ceph
           trigger-phrase: ''
+          skip-build-phrase: '^jenkins do not test.*'
           only-trigger-phrase: false
           github-hooks: true
           permit-all: true
--- a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
+++ b/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
@@ -77,6 +77,7 @@
         org-list:
         - ceph
         trigger-phrase: ''
+        skip-build-phrase: '^jenkins do not test.*'
         only-trigger-phrase: false
         github-hooks: true
         permit-all: true


vasukulkarni:
As per discussion from alfredo in IRC:

<alfredodeza> skip-build-phrase (string) – when filled, adding this phrase to the pull request title or body will not trigger a build (optional)

ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -28,7 +28,8 @@
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
-          trigger-phrase: ''
+          trigger-phrase: ''    
+          skip-build-phrase: 'qa tests .* skip build'
```
alfredodeza: I think this is not a regex or a list of regexes/phrases, but just a single string that if present in the title, will skip the build. So maybe something like "NO-TEST" or "SKIP-TEST" ? It will also need to be added to the other "make check" job for arm64

vasukulkarni: I just googled and it showed me an example with regex(http://yaks-all-the-way-down.com/2016/04/22/Presenting-Git-Hub-Pull-Request-Builder.html), but we stick to single word, we should also add QA to that string just to be explicit that its QA changes QA-SKIP-TEST probably?

alfredodeza: That link says this feature is kind of broken :( but it does say that regexes are OK. Since it can scan either title or body, then we could do a very lenient phrase that allows a brief description (which you seem to want). For example `skip-build-phrase: '^jenkins do not test(.*)'`

That way you could do this in the PR body:

    jenkins do not test: qa changes

vasukulkarni: Yeah i read that its broken too, but its also mid 2016 article, worth trying just to see. I will update the string, +1 for ^jenkins do not test(.*)' ( can be qa or doc changes)

vasukulkarni: hold on (.*) is for grouping, we just need .*

vasukulkarni:
@alfredodeza   anything else here?

alfredodeza:
@vasukulkarni looks great! only missing a rebase

vasukulkarni:
@alfredodeza had some minor issues but repushed.

PR #804, opened 2017-07-27
Abort builds on update

guits:
When a PR is updated, jobs that were launched by the previous PR are not
aborted. It can even happen that you update a PR several times in a
short time. So it's n*x jenkins slave locked until they end.

Adding `cancel-builds-on-update` to the jobs templates force the job to be
aborted to free a jenkins slave quicker.


guits:
see: https://docs.openstack.org/infra/jenkins-job-builder/triggers.html?highlight=abort#triggers.github-pull-request

alfredodeza:
I had no idea about this, sounds like a good feature to me! @andrewschoen any comments?

PR #809, opened 2017-08-01
Ceph volume fixes for PRs

alfredodeza:
It now requires a python3 node to be able to run py35 tests. Had to omit Python 3.6 tests for now as that is not available in build nodes (yet).

Fixes the `scm` section to correctly narrow down the scope to the PR only (before it would just build any PR or any merge even)

### DIFF
--- a/ceph-volume-pr/build/build
+++ b/ceph-volume-pr/build/build
@@ -5,4 +5,8 @@ set -ex
 pkgs=( "tox" )
 install_python_packages "pkgs[@]"
 
-cd src/ceph-volume && $VENV/tox -rv
+cd src/ceph-volume
+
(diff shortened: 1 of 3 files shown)


ceph-volume-pr/build/build
```diff
@@ -5,4 +5,14 @@ set -ex
 pkgs=( "tox" )
 install_python_packages "pkgs[@]"
 
-cd src/ceph-volume && $VENV/tox -rv
+cd src/ceph-volume
+
+# Python 2.7
+$VENV/tox -v -e py27
```
andrewschoen: ``$VENV/tox -v -e py27,py35,flake8`` would also work here I think.

PR #813, opened 2017-08-03
ceph-ansible-prs: add naginator configuration

leseb:
Auto-restart failed builds 3 times with 60 seconds interval.

Signed-off-by: Sébastien Han <seb@redhat.com>

PR #815, opened 2017-08-07
Revert "ceph-ansible-prs: add naginator configuration"

guits:
Reverts ceph/ceph-build#813


guits:
@alfredodeza @djgalloway 
Things are getting worse since this plugin has been installed.
Indeed, every failing jobs are restarted, though most of these failures can't be solved by simply restarting the job, most of the time, either we would need to reboot the slave or to get a new slave.
This results by locking nodes for more time than before without fixing anything.

The purpose of this plugin is cool I think we should think for a good configuration before enabling it (sometimes failures are valid).

I propose to revert these changes since I'm not sure someone has bandwidth to look at this configuration at the moment.

djgalloway:
Merged w/o PR check because Jenkins is pending shutdown.

PR #816, opened 2017-08-07
scripts: Don't sync any lockfiles leftover by reprepro

djgalloway:


ktdreyer:
AFAIK the entire `db` directory is a reprepro-specific implementation detail (Apt does not use it), so we could just avoid rsync'ing `db` directories too.

Issue #817, opened 2017-08-08
cancel-builds-on-update option is not working as expected


guits:
@alfredodeza @djgalloway
the option `cancel-builds-on-update` introduced in 18a1bbc doesn't work at all, I can see jobs gets queued without being canceled when I have to update existing PRs.

I suspect this is because we are using an old version of Jenkins Jobs Builder.
Indeed, this option is available from JJB 2.0 and later.

alfredodeza:
@guits you can use raw/plain XML like in here: 

https://github.com/ceph/ceph-build/blob/master/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml#L31-L36

PR #818, opened 2017-08-08
Abort on updated pr

guits:
Add CancelOnUpdate option in raw/xml format

    The option `cancel-builds-on-update` doesn't work in job template
    because we are not using a recent version of JJB, therefore we had to
    revert the previous commit.

    This commit adds the cancel-builds-on-update option but in raw/xml
    format instead.

    Fixes: #817


guits:
@djgalloway yes, the option in jenkins jobs builder is available on 2.0 and later so we need to implement it in another way

djgalloway:
@guits I mean, we could just upgrade JJB.

PR #821, opened 2017-08-08
ceph-volume-ansible-prs: do not use raw xml, use the jjb option for ghprb

alfredodeza:

### DIFF
--- a/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
+++ b/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
@@ -35,6 +35,7 @@
 
     triggers:
       - github-pull-request:
+          cancel-builds-on-update: true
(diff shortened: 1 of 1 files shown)


ktdreyer:
That's definitely the "what"

but why

In particular, are we working around a bug in JJB upstream that will soon be resolved, so we can eventually revert this?

alfredodeza:
the newer version of JJB allows to declare it as part of the plugin, the older version didn't

ktdreyer:
Why don't we upgrade JJB? What's the version that will allow us to revert?

djgalloway:
That's what we did here: https://github.com/ceph/ceph-build/pull/819/commits/4f10012571b120e01854fae49be4d43f382787f1

ktdreyer:
thanks @djgalloway !

PR #825, opened 2017-08-09
Updates to ansible-jenkins role

djgalloway:
Also updated the Jenkins master doc https://github.com/ceph/ceph-sepia-secrets/pull/251


ansible/examples/master.yml
```diff
@@ -36,11 +36,15 @@
       - 'postbuildscript'
       - 'jobConfigHistory'
 
-    - port: 8080
+    - jenkins_port: 8080
     - prefix: '/build'
     - xmx: 8192
```
alfredodeza: I know this was not set by you, but if this is doing what I *think* will do, then it will set the JAVA JVM max memory usage to 8GB of ram. At some point this worked, but I don't think you want that

ansible/roles/ansible-jenkins/templates/etc_default.j2
```diff
@@ -0,0 +1,48 @@
+# {{ ansible_managed }}
+
+# pulled in from the init script; makes things easier.
+NAME=jenkins
+
+# location of java
+JAVA=/usr/bin/java
+
+# From https://jenkins.io/blog/2016/11/21/gc-tuning/
+JAVA_ARGS="-Xmx20g -Xms20g -Djava.awt.headless=true -Dhudson.model.User.SECURITY_243_FULL_DEFENSE=false -Dhudson.model.ParametersAction.keepUndefinedParameters=true -server -XX:+AlwaysPreTouch -Xloggc:/var/log/jenkins/gc-%t.log -XX:NumberOfGCLogFiles=5 -XX:+UseGCLogFileRotation -XX:GCLogFileSize=20m -XX:+PrintGC -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:+PrintHeapAtGC -XX:+PrintGCCause -XX:+PrintTenuringDistribution -XX:+PrintReferenceGC -XX:+PrintAdaptiveSizePolicy -XX:+UseG1GC -XX:+ExplicitGCInvokesConcurrent -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication -XX:+UnlockExperimentalVMOptions -XX:G1NewSizePercent=20 -XX:+UnlockDiagnosticVMOptions -XX:G1SummarizeRSetStatsPeriod=1"
```
alfredodeza: ok, so the playbook doesn't seem to be setting `Xmx`, so again, here you need to be careful because this means it will want to use 20GB of RAM

djgalloway: That was copied directly from jenkins.ceph.com so I figured that's the amount you'd tuned it to.

PR #826, opened 2017-08-09
adds a tool to check node usage in the labs

GregMeno:

### DIFF
--- a/lab-cop/config/definitions/lab-cop.yml
+++ b/lab-cop/config/definitions/lab-cop.yml
@@ -1,0 +1,41 @@
+- scm:
+    name: lab-cop
+    scm:
+      - git:
+          url: https://github.com/ceph/teuthology.git
+          branches:
+            - ${sha1}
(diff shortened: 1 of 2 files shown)


lab-cop/config/definitions/lab-cop.yml
```diff
@@ -0,0 +1,41 @@
+- scm:
+    name: lab-cop
+    scm:
+      - git:
+          url: https://github.com/ceph/teuthology.git
+          branches:
+            - ${sha1}
```
alfredodeza: this is usually done for pull requests, are you looking to trigger this on PRs? if so you will need a `triggers:` section with that defined in. Otherwise you can trigger on merges, in which case you need to remove the `branches` section (it defaults to anything/everything) *or* define what branches you are interested in

GregMeno: @alfredodeza thank you 
yeah I already nuked the trigger section -- I plan to kick it manually for now. 
I'll get rid of this too

PR #839, opened 2017-08-22
nfs-ganesha-stable: fixes to get build scripts working

alimaredia:
-ntirpc scm wasn't applying properly
-clean up ntirpc directory
-various pbuilder related fixes
-directory ownership fix for creating local ntirpc repo

### DIFF
--- a/nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
+++ b/nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
@@ -19,6 +19,7 @@
           skip-tag: true
           wipe-workspace: true
           basedir: "nfs-ganesha-debian"
+
 - scm:
     name: ntirpc
     scm:
@@ -29,6 +30,7 @@
           skip-tag: true
           wipe-workspace: true
           basedir: "ntirpc"
+
 - job:
     name: nfs-ganesha-stable
     project-type: matrix
@@ -135,6 +137,7 @@ If this is checked, then the binaries will be built and pushed to chacra even if
     scm:
       - nfs-ganesha
       - nfs-ganesha-debian
+      - ntirpc
 
     builders:
       - shell: |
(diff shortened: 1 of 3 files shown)


nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
```diff
@@ -19,6 +19,7 @@
           skip-tag: true
           wipe-workspace: true
           basedir: "nfs-ganesha-debian"
+
```
alimaredia: @alfredodeza for some reason the ntirpc scm wasn't being applied here. The spacing was the only reason i can think of. Do you have any guess as to why it wasn't there? I added it in by hand on the jenkins gui to find the rest of the errors fixed in this commit.

alfredodeza: a spacing here would seem consistent, not sure, you can always apply this on the fly and test it out

alimaredia: i couldn't find a way to do it in the GUI the same way i can change around the build and setup scripts. If the rest of it is fine, we can just merge it and hope for the best?

alfredodeza: not in the gui, this is documented in the README of this repo

nfs-ganesha-stable/config/definitions/nfs-ganesha-stable.yml
```diff
@@ -135,6 +137,7 @@ If this is checked, then the binaries will be built and pushed to chacra even if
     scm:
       - nfs-ganesha
       - nfs-ganesha-debian
+      - ntirpc
```
alimaredia: @alfredodeza this missing line was what was causing the ntirpc scm to not be acknowledged.

PR #843, opened 2017-08-24
ceph-docker: run flake8 and lint jobs on 2.jenkins

djgalloway:

### DIFF
--- a/ceph-docker-flake8/build/build
+++ b/ceph-docker-flake8/build/build
@@ -28,8 +28,8 @@ function check(){
 }
 
 function main() {
-    # install some of our dependencies
-    if [ "${HUDSON_URL}" = "https://jenkins.ceph.com/" ]
+    # install some of our dependencies if running on a jenkins slave
+    if [[ -n "$HUDSON_URL" ]]
     then
         sudo yum -y install epel-release
         sudo yum -y install docker jq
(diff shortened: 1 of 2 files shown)


ceph-docker-flake8/build/build
```diff
@@ -29,7 +29,7 @@ function check(){
 
 function main() {
     # install some of our dependencies
-    if [ "${HUDSON_URL}" = "https://jenkins.ceph.com/" ]
+    if [ "${HUDSON_URL}" = "https://2.jenkins.ceph.com/" ]
```
andrewschoen: Why this condition at all? Was somebody trying to build this job so it could be run locally?

djgalloway: Looks like: https://github.com/ceph/ceph-build/commit/83b99762805edd3bf133ee1a2d02674b90c241c5

djgalloway: Maybe change it to check if the env var is set at all?

andrewschoen: Yeah, I think that makes better sense. Maintaining jenkins urls in jobs will get tedious if we need to add more master nodes.

djgalloway:
@andrewschoen updated to just check if var is set.

PR #844, opened 2017-08-28
tcmu-runner: remove UAPI patch and add Ceph repo for build

dillaman:

### DIFF
--- a/tcmu-runner/build/build_rpm
+++ b/tcmu-runner/build/build_rpm
@@ -2,11 +2,39 @@
 set -ex
 
 BRANCH=`branch_slash_filter $BRANCH`
+CEPH_BRANCH=$(branch_slash_filter $CEPH_BRANCH)
 
 # Only do actual work when we are an RPM distro
 if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; then
     exit 0
 fi
+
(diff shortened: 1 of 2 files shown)


tcmu-runner/build/build_rpm
```diff
@@ -2,12 +2,40 @@
...
+# Get .repo file from appropriate shaman build
+REPO_URL="https://shaman.ceph.com/api/repos/ceph/$CEPH_BRANCH/$CEPH_SHA1/$DISTRO/$RELEASE/flavors/default/repo"
+TIME_LIMIT=1200
+INTERVAL=30
+REPO_FOUND=0
+
+# poll shaman for up to 10 minutes
+while [ "$SECONDS" -le "$TIME_LIMIT" ]
+do
+  if `curl --fail -L $REPO_URL > $WORKSPACE/shaman.repo`; then
```
alfredodeza: we've had issues where `shaman.repo` files are leftover and then they break other builds because the repos are gone after a couple of weeks. Could you add a cleanup for this?

The samba job does it like:

https://github.com/ceph/ceph-build/blob/master/samba/config/definitions/samba.yml#L114-L119

dillaman: Thanks -- that was my inspiration. I'll fix before I do a build.

PR #845, opened 2017-08-28
tcmu-runner: remove shaman repo file after build

dillaman:

### DIFF
--- a/tcmu-runner/config/definitions/tcmu-runner.yml
+++ b/tcmu-runner/config/definitions/tcmu-runner.yml
@@ -103,3 +103,10 @@ If this is checked, then the binaries will be built and pushed to chacra even if
       - inject-passwords:
           global: true
           mask-password-params: true
+
+    publishers:
+      - postbuildscript:
+          script-only-if-succeeded: False
+          script-only-if-failed: False
+          builders:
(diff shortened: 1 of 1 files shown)


dillaman:
@alfredodeza Am I missing something? I see the following error when it tried to remove the repo file: https://jenkins.ceph.com/job/tcmu-runner/10/console

alfredodeza:
Yep, you are building on 'master':

    Building on master in workspace /var/lib/jenkins/jobs/tcmu-runner/workspace

Which can happen if you don't specify what type of node you want. For example:
https://github.com/ceph/ceph-build/blob/master/ceph-installer-coverage/config/definitions/ceph-installer-coverage.yml#L3

trusty, centos, small, large, x86_64, are a few of the labels you can use to describe what you need.

dillaman:
The samba definition doesn't have anything like that and it appears to work: https://github.com/ceph/ceph-build/blob/master/samba/config/definitions/samba.yml

alfredodeza:
That job is a "matrix" job:

    project-type: matrix

That means that it can run in master, but it is like a "parent" job, the `matrix` will extrapolate the `label-expression` into different machines:

https://github.com/ceph/ceph-build/blob/master/samba/config/definitions/samba.yml#L49-L68

You end up with something like: <parent (matrix) job> -> multiple distros/versions/archs builds+machines

PR #846, opened 2017-08-28
tcmu-runner: added missing distro release

dillaman:

### DIFF
--- a/tcmu-runner/build/build_rpm
+++ b/tcmu-runner/build/build_rpm
@@ -9,6 +9,8 @@ if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; t
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


tcmu-runner/build/build_rpm
```diff
@@ -9,6 +9,8 @@ if test "$DISTRO" != "fedora" -a "$DISTRO" != "centos" -a "$DISTRO" != "rhel"; t
     exit 0
 fi
 
+RELEASE="$(lsb_release --short -r | cut -d '.' -f 1)" # sytem release
```
alfredodeza: you might get into trouble here, since we use Xenial for all Debian-based builds like Jessie, or Trusty. Was this variable somehow not defined from the parent job?

dillaman: Copied-and-pasted from samba's build_rpm (which is not applicable for Debian-based systems).

PR #854, opened 2017-09-01
jenkins-job-builder: install a newer jenkins-job-builder

idryomov:
kernel-trigger job needs to specify do-not-fetch-tags, otherwise
cloning fails, "No changes" is reported up and nothing gets triggered
(see commit b34f8faa08d4 ("kernel: do not fetch git tags") for more
details).  do-not-fetch-tags was added to jenkins-job-builder in commit
badab0264717 ("Add `do-not-fetch-tags` to CloneOption for Git"), which
isn't in any tag.  The version we get from pypi, 2.0.0.0b2, simply
ignores it, leaving us with no kernel builds.

Let's install from github until a new version is pushed to pypi.

### DIFF
--- a/ceph-build-pull-requests/build/build
+++ b/ceph-build-pull-requests/build/build
@@ -3,7 +3,8 @@
 set -e
 
 # the following two methods exist in scripts/build_utils.sh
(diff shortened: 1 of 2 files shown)


idryomov:
Superseded by https://github.com/ceph/ceph-build/pull/858.

PR #858, opened 2017-09-01
kernel: use raw xml for configuring scm module

idryomov:
kernel-trigger job needs to specify do-not-fetch-tags, otherwise
cloning fails, "No changes" is reported and nothing gets triggered (see
commit b34f8fa ("kernel: do not fetch git tags") for more details).
do-not-fetch-tags was added to jenkins-job-builder in commit
badab0264717 ("Add do-not-fetch-tags to CloneOption for Git"), which
isn't in any tag.  The version we get from pypi, 2.0.0.0b2, simply
ignores it, leaving us with no kernel builds.

Use raw xml generated with jenkins-job-builder's master to get
<noTags>true</noTags> up there after all.

### DIFF
--- a/kernel-trigger/config/definitions/kernel-trigger.yml
+++ b/kernel-trigger/config/definitions/kernel-trigger.yml
@@ -20,16 +20,49 @@
       - github
 
     scm:
-      - git:
-          url: https://github.com/ceph/ceph-client.git
-          branches:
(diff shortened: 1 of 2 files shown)


ktdreyer:
:heart: this clear commit log, thanks @idryomov

PR #863, opened 2017-09-12
ceph-ansible-prs: only run lvm_osd tests on ceph@master

andrewschoen:
These lvm_osd tests won't pass until 12.2.1 is released, but if we have
them use ceph@master then they should pass. After 12.2.1 is out we can
move these back to be a 'luminous' test.


ktdreyer:
What was the Ceph PR (for v12.2.1) that resolved this?

alfredodeza:
There are two PRs that I think triggered this:

https://github.com/ceph/ceph/pull/17408
https://github.com/ceph/ceph/pull/17653

Both have links back to the PRs for master and their respective bugzilla tickets.

In short: when using non-lv devices (e.g. plain partitions) for journals, after a reboot the system may not bring up OSDs correctly

PR #871, opened 2017-09-26
ceph-ansible-prs: allow for tests to be auto started or only triggered

andrewschoen:
This creates two different projects, one which lists scenarios that will
auto start on every PR and another which lists scenarios that will only
start if a comment is made on the PR with its trigger phrase.


andrewschoen:
@leseb yes, I believe that would still work.

PR #872, opened 2017-09-26
ceph-ansible-prs: setting only-trigger-phrase with a var did not work

andrewschoen:
I tried setting the value of only-trigger-phrase as a variable on the
project so we didn't have to duplicate the job template but that didn't
seem to work. Duplicating the job template does allow us to create jobs
that do not auto start when PRs are created.

### DIFF
--- a/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
+++ b/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
@@ -8,10 +8,8 @@
     scenario:
       - centos7_cluster
       - xenial_cluster
-    only_trigger_phrase:
-      - false
     jobs:
-        - 'ceph-ansible-prs-{release}-{ansible_version}-{scenario}'
+        - 'ceph-ansible-prs-auto'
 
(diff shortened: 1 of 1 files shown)


andrewschoen:
@leseb it seems the change in https://github.com/ceph/ceph-build/pull/871 didn't quite get what I wanted. The var I used for ``only-trigger-phrase`` wasn't used and all jobs still started automatically. This does fix that, but it seems like commenting on a PR with 'jenkins test this please' only triggers jobs that have already started for that PR, meaning we couldn't use that phrase to trigger everything. You can see here that this did not trigger all jobs: https://github.com/ceph/ceph-build/pull/871

leseb:
Well, personally I think the 'jenkins test this please', still remains useful to do the "final validation" of a PR.

andrewschoen:
@leseb this is updated now so that you can either comment with ``jenkins test $scenario_name`` to run an individual scenario or ``jenkins test all`` to run all the scenarios.

leseb:
@andrewschoen thanks!

PR #877, opened 2017-09-29
ceph-ansible-nightly: test all luminous scenarios

andrewschoen:

### DIFF
--- a/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
+++ b/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
@@ -22,6 +22,46 @@
     jobs:
         - 'ceph-ansible-nightly-{release}-{ansible_version}-{scenario}'
 
+- project:
+    name: ceph-ansible-nightly-luminous
+    release:
+      - luminous
+    ansible_version:
+      - ansible2.3
+    scenario:
+      - centos7_cluster
+      - xenial_cluster
+      - docker_cluster
+      - update_cluster
+      - lvm_osds
+      - purge_lvm_osds
+      - journal_collocation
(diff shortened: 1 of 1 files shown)


leseb:
I like the nightly approach for testing all the scenarios. Unfortunately, I don't think we've reached the desired stability of the CI to get reliable results from that run.
"JavaErrors" and "vagrant sync" haven't been fixed yet, so they will create false issues on these nightly. Once we get ride of these errors I think we can consider the CI as reliable on each run and then we can move ahead with this.

Thoughts?

andrewschoen:
I think there is still value in running them and I haven't been seeing those errors you mentioned quite as much anymore. In looking over some of the nightly tests we have for jewel all the failures I looked at were valid. For example ``jewel-ansible2.3-centos7_cluster`` has been failing since Sept. 15th and all the ones I checked were not CI related: https://2.jenkins.ceph.com/view/ceph-ansible/job/ceph-ansible-nightly-jewel-ansible2.3-centos7_cluster/

I also worry that now that we're only starting a few tests automatically for each PR that some of these scenarios will not get run very often. These nightly tests are an easy way for us to keep that test coverage on master without burdening the CI on each pull request.

These jobs are set up so that they can be run ad-hoc against any ceph-ansible branch or tag, which is a really nice feature to have as well.

leseb:
Well, I see the JavaError on daily basis. I guess the problem is that there are very few people looking at the nightly (I don't, I probably should) and also sending patches to fix the bugs.
We can merge this and see if we get any valid errors.

Regarding "only starting a few tests automatically for each PR", the goal is prior to merging the PR is to run 'jenkins test all'. This won't happen every time though.

PR #884, opened 2017-10-10
Support CentOS static libvirt slaves & update playbook

djgalloway:
Playbook tested on Zesty and CentOS slave.

Only thing untested is the change to build_utils.sh but looks pretty standard.


ktdreyer:
thanks for catching the python-jenkins upstream thing here!

PR #890, opened 2017-10-19
ansible: slave playbook overhaul (update and format)

djgalloway:


djgalloway:
`become: true` doesn't work in ansible v1.9.1 which is what prado ships.  Researching alternative methods.

djgalloway:
Relies on https://github.com/ceph/prado/pull/18

PR #896, opened 2017-10-27
ceph-ansible-nightly: add more nightly tests

guits:

### DIFF
--- a/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
+++ b/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
@@ -17,8 +17,12 @@
       - purge_cluster_non_container
       - purge_filestore_osds_container
       - purge_filestore_osds_non_container
+    ceph_ansible_branch:
+      - stable-2.2
+      - stable-3.0
+      - master
     jobs:
-        - 'ceph-ansible-nightly-{release}-{ansible_version}-{scenario}'
+      - 'ceph-ansible-nightly-{release}-{ansible_version}-{ceph_ansible_branch}-{scenario}'
 
 - project:
     name: ceph-ansible-nightly-luminous
@@ -51,15 +55,18 @@
       - purge_filestore_osds_non_container
       - purge_bluestore_osds_container
       - purge_bluestore_osds_non_container
+    ceph_ansible_branch:
+      - stable-3.0
+      - master
     jobs:
-        - 'ceph-ansible-nightly-{release}-{ansible_version}-{scenario}'
+        - 'ceph-ansible-nightly-{release}-{ansible_version}-{ceph_ansible_branch}-{scenario}'
 
 - job-template:
-    name: 'ceph-ansible-nightly-{release}-{ansible_version}-{scenario}'
+    name: 'ceph-ansible-nightly-{release}-{ansible_version}-{ceph_ansible_branch}-{scenario}'
     node: vagrant&&libvirt&&centos7
     concurrent: true
     defaults: global
-    display-name: 'ceph-ansible: Nightly [{release}-{ansible_version}-{scenario}]'
+    display-name: 'ceph-ansible: Nightly [{release}-{ansible_version}-{ceph_ansible_branch}-{scenario}]'
     quiet-period: 5
     block-downstream: false
     block-upstream: false
@@ -77,7 +84,7 @@
       - string:
           name: CEPH_ANSIBLE_BRANCH
           description: "The ceph-ansible branch (or tag) to test"
-          default: master
+          default: "{ceph_ansible_branch}"
 
     triggers:
       - timed: '@daily'


ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
```diff
@@ -16,8 +16,12 @@
       - purge_cluster_non_container
       - purge_filestore_osds_container
       - purge_filestore_osds_non_container
+    ceph_ansible_branch:
+      - v2.2.11
```
alfredodeza: this looks like a tag though, wouldn't you want to test the stable-2.2 branch here?

leseb: stable-2.2 currently points to 2.2.11 but will soon point to 2.2.12. rhcs 2.{3,4} have 2.2.11 so it's better to stick with the tag for this one.

alfredodeza: exactly, a tag doesn't move, so what is the idea here, to test this every night even though it doesn't change?

leseb: Yes, that's the idea, but this is temporary, this is just to validate that it works. Once we do, we will change to stable-2.2. This is something we need to demonstrate to @GregMeno.

ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
```diff
@@ -16,8 +16,12 @@
       - purge_cluster_non_container
       - purge_filestore_osds_container
       - purge_filestore_osds_non_container
+    ceph_ansible_branch:
+      - stable-2.2
```
leseb: please keep `v2.2.11` tag.

andrewschoen: to use a tag jenkins requires it to be in the format of ``tags/$tag``. In this case it'd be ``tags/v2.2.11``

alfredodeza: @leseb could you explain why you need to test a tag every night? A tag never changes, so the results would always be the same. If you want to verify that a given tag works, it could be done with a one-time run. There is such a job already in https://2.jenkins.ceph.com/job/ceph-ansible-scenario/

leseb: I understand your concern, but what if I want to test all the scenarios against one tag?
How should I do that?

andrewschoen: @leseb The current nightly jobs also take a ceph-ansible branch or tag as a parameter. For example, I just started this nightly against ``tags/v2.2.11``: https://2.jenkins.ceph.com/job/ceph-ansible-nightly-jewel-ansible2.3-centos7_cluster/78/console

The job that @alfredodeza linked to also allows you to specify which scenario you'd like to test as well as the ceph-ansible and ceph branch.

Once this PR merges it'll just create new jobs, each defaulting to the branch or tag defined here.

GregMeno:
It's my understanding that this PR is supposed to configure jenkins to test a setup where we ship 
ceph-ansible-3 with jewel. It seems that we're held up by agreement of what ceph-ansible 2 version to test against. @leseb Why are you asking for the tag instead of the branch? I'm confused why the stable-2.2 branch is not sufficient

leseb:
@GregMeno, my argument is the following: stable-2.2 currently points to 2.2.11 but will soon point to 2.2.12. rhcs 2.{3,4} have 2.2.11 so it's better to stick with the tag for this one.

andrewschoen:
@leseb we could merge this with ``stable-2.2`` so the nightlies are automatically kept up to date and then run them manually with ``tags/v2.2.11``. Would that work?

Do we need to be running against ``tags/v2.2.11`` every day or would it be a one time validation?

guits:
@leseb here is the diff between `v2.2.11` and `v2.2.12`
```
f067dc95fe57dacf7f5619ddc0bb7ed5b33a7308 (tag: v2.2.12) Merge pull request #1932 from fghaas/1902
b562b84e32794b7c9b48335e8313fe5bb429849f Merge pull request #2034 from fghaas/2033
2e111221c1b6ad261608e7b2173d32dd31d45753 infra: Fix ceph.conf creation when taking over existing cluster
fb9fa6c65f4f09e77d10e06c9ae7102415ba845c Merge pull request #1935 from fghaas/1907
23c42939cd756aeed13d983aef274aa2c84924af stable-2.2: Drop remaining ceph_nfs_rgw_* variables
b278d7cbcdf22ad0cc33b44fee0fe1d771c113b4 stable-2.2: drop nfs_obj_gw
3b01f8b41240dea05e26f8b6c01b72fb32de7978 stable-2.2: Drop RGW FSAL from ganesha.conf
38eb97e6b6fd9d55284ade68c6ff698880b992a5 Merge pull request #1936 from fghaas/1905
b6cb5f76fb6f40cd80476bac2e7d4ce6e58c4ce9 stable-2.2: add ceph_nfs_ceph_user to nfss.yml.sample
774a28d2a6f07de2976cf97ae77ef8fce5e9f073 stable-2.2: docs: drop "OSD Configuration"
cef1bcc90508bab649dddeaf16893159facfc1c6 stable-2.2: docs: Mention OSD scenario booleans
6fcecb0ccf162fa9f34bf556ba6fa0d00fcc24ef docs: add docs for dev contributions
5a6ec4d4f7cf5ab6faba8c3d2ac70a32f65c1bd7 docs: add docs for ceph_conf_overrides usage
fda37266d7d9fafc50dfef7e226dd43805322113 docs: add links to example videos
08390ef53a3ac96ef5c5955bc04b03351f234060 docs: docs for installation and basic config
c02fd8ec1e6e4e07f56943fcaedf814fbf7626ac Merge pull request #1912 from fghaas/1905
72e32ae0bf1f060aaabd88ddb9c8dd000c0c5d97 stable-2.2: Introduce ceph_nfs_ceph_user
04868bf882da39857a377ad832ea42f0ffe449b9 stable-2.2: Fix nfs-ganesha configuration template
f349b90bc8c8413751df32c2be1eeea3b7f38357 stable-2.2: Fix nfs-ganesha package installation
```

alfredodeza:
@leseb if you are testing against a tag, and that passes, it will pass today and will pass tomorrow and there will not be any changes ever. What are you trying to accomplish here?

If it fails, I guess you would want to try and add changes and see if those work? (if not, please explain here what would you intend to do ) But with a tag you wouldn't get the benefit of trying out any fixes.

We understand that the "stable-2.2 currently points to 2.2.11  but will soon point to 2.2.12",  because that is how that branch is intended to work: it will keep tracking the next 2.2.X version.

Are you trying a one-off test here by pointing to a tag? What would you want to do if there is a failure and a change is needed? Update the tag?

leseb:
@alfredodeza, I was trying to ease what we have to demonstrate with @guits. I asked you in a previous comment: 

> "I understand your concern, but what if I want to test all the scenarios against one tag?
> How should I do that?

It looks like we will have to go with the annoying manual option. Using the tag temporarily would have saved the burden, but it seems that no one is ready to accept that.
Thus I'm merging this.

alfredodeza:
Testing all scenarios against one tag, or more specifically, how to test N scenarios against a given tag: manually.

There is no support in our jobs to do something like that for all scenarios (or for a selection of scenarios). 

If that is something you may want to use frequently, you may want to consider creating a job that can accept those as parameters. Let just please not make a nightly test job do something that doesn't make sense just because there is no automatic way to run a job with different requirements.

PR #903, opened 2017-11-02
ansible: always update slave.jar

djgalloway:
Makes sure latest version of JNLP is being used.  Primarily useful when
re-running the playbook against a static slave.


ansible/examples/slave_static.yml
```diff
@@ -301,6 +301,7 @@
       template:
         src: "templates/systemd/jenkins.service.j2"
         dest: "/etc/systemd/system/jenkins.service"
+        force: yes
```
alfredodeza: I thought this was for slave.jar not the systemd unit. Does this file need to change if the slave.jar gets updated?

djgalloway: Whoops you right.

PR #909, opened 2017-11-10
ceph{-,dev-,dev-new-}build: install centos-release-scl repo

tchaikov:
enable/install centos-release-scl on rhel7/centos7, so we can use
newer gcc/toolchain offered by Software Collections (scl)'s devtoolset
see https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/

### DIFF
--- a/ceph-build/build/setup_rpm
+++ b/ceph-build/build/setup_rpm
@@ -39,6 +39,7 @@ CentOS|Fedora|SUSE*|RedHatEnterpriseServer)
                 $SUDO yum install -y yum-utils
                 ;;
         esac
+        $SUDO yum install -y centos-release-scl
         sed -e 's/@//g' < ceph.spec.in > $DIR/ceph.spec
         $SUDO yum-builddep -y $DIR/ceph.spec
         ;;
--- a/ceph-build/build/setup_rpm
+++ b/ceph-build/build/setup_rpm
@@ -29,24 +29,10 @@ tar xzf *.orig.tar.gz
 cd ceph-*
 pwd
 
-case $(lsb_release -si) in
-CentOS|Fedora|SUSE*|RedHatEnterpriseServer)
-        case $(lsb_release -si) in
-            SUSE*)
-                $SUDO zypper -y yum-utils
-                ;;
-            *)
-                $SUDO yum install -y yum-utils
-                ;;
-        esac
-        $SUDO yum install -y centos-release-scl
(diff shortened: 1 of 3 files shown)


tchaikov:
see also https://github.com/ceph/ceph-build/pull/909

tchaikov:
@alfredodeza is there any way we can test this change?

ceph-build/build/setup_rpm
```diff
@@ -39,6 +39,19 @@ CentOS|Fedora|SUSE*|RedHatEnterpriseServer)
...
                 ;;
         esac
+        major_ver=$(lsb_release -rs | cut -f1 -d.)
+        case $(lsb_release -si) in
+            CentOS)
+                if [ $major_ver = 7 ]; then
+                    $SUDO yum install centos-release-scl
+                fi
+                ;;
+            RedHatEnterpriseServer)
```
alfredodeza: we don't build for RHEL at all, don't think this is necessary. CentOS7 should be fine, unless you feel strongly about leaving it here for granularity?

There is also no need to specify `$major_ver` either because we don't have anything that is not CentOS7

Maybe I would suggest just commenting that when needed, RHEL can be enabled with `rhel-server-rhscl-7-rpms`

tchaikov: okay, i found that we are checking `RedHatEnterpriseServer` in the case . will drop these tests.

tchaikov:
@alfredodez fixed and repushed.

ktdreyer:
I really wish we'd talked more about this. It is not clear if this is getting security updates.

tchaikov:
@ktdreyer the reason why i chose devtoolset-7 is 1) the newer the merrier 2) i assume that its EOL will be later than that of devtoolset-6. if this is not true, and turns out to be a concern of you, i am happy to use devtoolset-6.

dmick:
So, sadly, this breaks the arm64 build, because there is no SCL for Centos 7 in aarch64.

(this must have been noted as a build failure.)

liewegas:
Crap... the arm builds fail intermittently so we probably ignored it.

In other news, we're thinking about moving to C++14 too, which should also require an updated toolchain.  So perhaps the next investigation to do here is what it needed to get the updated gcc etc for arm...

dmick:
Added http://tracker.ceph.com/issues/22301 to track this

PR #913, opened 2017-11-10
ceph-volume-ansible-prs: use subcommand to build the testing path

andrewschoen:
I'm not exactly sure of the reason but ``TEST_PATH`` was being rendered by jjb as a list. However, we can easily build the testing path by using the ``subcommand`` factor that we already have.

### DIFF
--- a/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
+++ b/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
@@ -9,8 +9,6 @@
     scenario:
       - create
       - prepare_activate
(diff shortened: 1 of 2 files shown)


ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
```diff
@@ -10,7 +10,7 @@
       - create
       - prepare_activate
     test_path:
```
alfredodeza: I think this needs to be:

    test_path: "src/ceph-volume/ceph_volume/tests/functional/lvm"

PR #914, opened 2017-11-12
setup_pbuilder: point gcc to gcc-7 on trusty

tchaikov:

### DIFF
--- a/ceph-dev-build/build/setup_pbuilder
+++ b/ceph-dev-build/build/setup_pbuilder
@@ -83,6 +83,41 @@ echo "BUILD_HOME=`mktemp -d`" >> ~/.pbuilderrc
 # in newer versions. This ticket solves the specific issue in 8.1.1 (which vendors urllib3):
 # https://github.com/shazow/urllib3/issues/567
 echo "USENETWORK=yes" >> ~/.pbuilderrc
+# point gcc,g++ to the newly installed ones
+hookdir=$HOME/.pbuilder/hook.d
+if [ "$DIST" = "trusty" ]; then
+    cat >> ~/.pbuilderrc <<EOF
+OTHERMIRROR="deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu $DIST main"
+OTHERMIRROR="$OTHERMIRROR | deb http://mirror.cs.uchicago.edu/ubuntu-toolchain-r $DIST main"
+OTHERMIRROR="$OTHERMIRROR | deb http://mirror.yandex.ru/mirrors/launchpad/ubuntu-toolchain-r $DIST main"
+ALLOWUNTRUSTED=yes
+EXTRAPACKAGES="g++-7"
+HOOKDIR=$hookdir
+EOF
+fi
+
+if [ ! -e $hookdir ]; then
+    mkdir -p $hookdir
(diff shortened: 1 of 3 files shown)


tchaikov:
i will give it a spin locally tomorrow.

tchaikov:
@alfredodeza i think we can pre-install gcc-7 on all centos / trusty jenkins slave for reducing the build time. if it makes sense to you also, i will be happy to update `ansible/slave.yml` also.

tchaikov:
also, i found that all three `setup_pbuilder`s are the same. is it possible to symlink the two of them to one? but i searched the symlinks in the source directory, none of them shows up.
```
find . -type f -ls
```

@alfredodeza is there any reason preventing us from doing so? if not, i'd like do so to dedup this.

ceph-dev-build/build/setup_pbuilder
```diff
@@ -83,6 +83,46 @@ echo "BUILD_HOME=`mktemp -d`" >> ~/.pbuilderrc
...
+OTHERMIRROR="deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu $DIST main"
+ALLOWUNTRUSTED=yes
+EXTRAPACKAGES="g++-7"
+HOOKDIR=$hookdir
+EOF
+fi
+
+if [ ! -e $hookdir ]; then
+    mkdir -p $hookdir
+    cat > $hookdir/E10update-gcc-alternatives <<EOF
```
tchaikov: it looks overkill for using `update-alternatives` in a pbuilder chroot tarball. but actually, i was thinking about reusing this in `slave.yml` . so it's less intrusive when being applied to the jenkins slave node.

alfredodeza:
@tchaikov any changes like this have a big impact on all of our builds. I know there is a lot of duplication, but the build script aren't getting worked on anymore, so it is difficult to try an make improvements that have a major impact.

I don't have any opinions on GCC-7, except maybe for how horrible it is sometimes to rely on PPAs hosted on launchpad. ceph-ansible has had many test failures when it can't retrieve nfs-ganesha from launchpad.

@djgalloway and @ktdreyer you might be interested in these changes

tchaikov:
if the basetgz is cached, and can be reused, we won't suffer that much if the PPA hosted by launchpad is not reachable?

alfredodeza:
@tchaikov caching doesn't mean a lot in this context because our builders are ephemeral. They usually aren't around for more than a couple of hours at the most

tchaikov:
@alfredodeza thanks for the reminder. and sorry, i just keep forgetting this.

tchaikov:
@alfredodeza i added more mirrors for ppa.launchpad.net/ubuntu-toolchain-r. HTH.

tchaikov:
tested locally. ceph builds fine using the pbuilder created with the hook even i change the warning gcc/libstdc++ checking in cmake to a fatal error.

tchaikov:
@liewegas @alfredodeza @djgalloway @ktdreyer anything i can do to make further progress? if you are good with removing support for trusty instead as discussed over ceph-{devel,user} ML. i will try to update ceph-build to do so.

tchaikov:
@alfredodeza thanks for your review. i am merging this then. will fix issues if any pops up.

PR #925, opened 2017-12-05
ceph{-,dev-,dev-new-}build: install centos-release-scl-rh repo for arm64

tchaikov:
the devtoolset-7 for aarch64 is not offered by centos-release-scl, but
at least centos-release-scl-rh has devtoolset-6.

see https://github.com/sclorg/centos-release-scl
see https://buildlogs.centos.org/centos/7/sclo/aarch64/rh/

Fixes: http://tracker.ceph.com/issues/22301

### DIFF
--- a/ceph-build/build/setup_rpm
+++ b/ceph-build/build/setup_rpm
@@ -30,7 +30,13 @@ cd ceph-*
 pwd
 
 $SUDO yum install -y yum-utils
-$SUDO yum install -y centos-release-scl
+if [ "$ARCH" = x86_64 ]; then
+    $SUDO yum install -y centos-release-scl
+elif [ "$ARCH" = arm64 ]; then
+    $SUDO yum install -y centos-release-scl-rh
+    $SUDO yum-config-manager --disable centos-sclo-rh
+    $SUDO yum-config-manager --enable centos-sclo-rh-testing
+fi
 sed -e 's/@//g' < ceph.spec.in > $DIR/ceph.spec
 $SUDO yum-builddep -y $DIR/ceph.spec
 
--- a/ceph-dev-build/build/setup_rpm
+++ b/ceph-dev-build/build/setup_rpm
@@ -30,7 +30,13 @@ cd ceph-*
 pwd
 
 $SUDO yum install -y yum-utils
-$SUDO yum install -y centos-release-scl
+if [ "$ARCH" = x86_64 ]; then
+    $SUDO yum install -y centos-release-scl
+elif [ "$ARCH" = arm64 ]; then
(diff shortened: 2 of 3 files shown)


tchaikov:
@alfredodeza @dmick ping?

dmick:
I agree the code does what's intended.  I don't know whether that's a reasonable thing to intend or not.  It worries me that we're moving so far outside the "supported tools" in a general sense, and I don't understand enough about the issue to know whether that's warranted.  Take that as the FUD it is.

tchaikov:
@dmick we are using centos for upstream build. strictly speaking, i don't think it belongs to "supported tools". and yes, the "testing" repo is even more scaring. but i think this is the best we can have at this moment, before CentOS community 1) builds SCL for aarch64 2) promote the SCL testing repo for aarch64 to the official mirror.

tchaikov:
@alfredodeza @dmick i've updated ceph.spec to ready aarch64, see the https://github.com/ceph/ceph/blob/master/ceph.spec.in#L130. it'd be ideal if we can have this change in ceph-build sooner. without it, the aarch64 is still broken on mimic-dev1 and master. otherwise we will need to wait on https://bugs.centos.org/view.php?id=14078&nbn=1 or reimage our CentOS builders with RHEL.

ktdreyer:
regarding "before CentOS community" (does some things)...

Who specifically is going to do these things?

It would be great to engage in the community directly so we ensure they get done. It's not beyond our control. We should take responsibility for new dependencies. The CentOS community is something we can influence and contribute to.

dmick:
@tchaikov: by "supported tools" I mean "the ones that are supported by distro vendors on the distros we care about", which still include CentOS and Ubuntu AFAIK, and certainly do unless we pretty seriously change our build system.

dmick:
I'm not going to stand in the way, and I certainly don't understand all the details; I'm just expressing nervousness and counseling conservative behavior.  As I said in chat, the folks who actually have to ship things are the ones directly affected, and if it results in "we can't release an upstream or downstream release", that would be bad.

tchaikov:
> we can't release an upstream or downstream release

i believe that i've taken that into consideration.

tchaikov:
> It would be great to engage in the community directly so we ensure they get done. It's not beyond our control. We should take responsibility for new dependencies. The CentOS community is something we can influence and contribute to.

@ktdreyer agreed. as long as we have enough bandwidth for this.

ktdreyer:
> @ktdreyer agreed. as long as we have enough bandwidth for this.

If things are competing for that, please let me know and we can figure it out together.

PR #928, opened 2017-12-06
setup_pbuilder: only use ppa on newer releases

tchaikov:
* only use GCC-7 offered by ppa repo on mimic and newer release.
  we are dropping the support of trusty in mimic and newer releases,
  but we still need to use the old toolchain shipped with trusty to
  build jewel and luminous. so, only use GCC-7 on trusty for building
  mimic and up. as it should be safe to use the new GCC ABI on the
  supported distros, like xenial.
* extract the ppa setup related code into build_utils.sh, so the
  setup_pbuilder scripts can reuse it.

### DIFF
--- a/ceph-build/build/setup_pbuilder
+++ b/ceph-build/build/setup_pbuilder
@@ -83,41 +83,7 @@ echo "BUILD_HOME=`mktemp -d`" >> ~/.pbuilderrc
 # in newer versions. This ticket solves the specific issue in 8.1.1 (which vendors urllib3):
 # https://github.com/shazow/urllib3/issues/567
 echo "USENETWORK=yes" >> ~/.pbuilderrc
-# point gcc,g++ to the newly installed ones
-hookdir=$HOME/.pbuilder/hook.d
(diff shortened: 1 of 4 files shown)


tchaikov:
> 1. build jewel and luminous on trusty with the old GCC shipped with the distro

at least, this one is addressed

> 2. do not build master, mimic-dev* branches on trusty

i will try to understand "parameter-factories" later in a rainy afternoon.

tchaikov:
http://qa-proxy.ceph.com/teuthology/kchai-2017-12-06_14:08:32-ceph-disk-wip-18589-luminous-kefu-distro-basic-vps/1939457/teuthology.log

```
2017-12-06T14:23:33.636 INFO:teuthology.orchestra.run.vpm163.stdout: ceph-fuse : Depends: libstdc++6 (>= 6) but 4.8.4-2ubuntu1~14.04.3 is to be installed
```

tchaikov:
@smithfarm currently, the luminous build does not install. this PR addresses this issue.

PR #935, opened 2017-12-09
ceph-dev{,-new}-trigger: pass DISTROS using a list

tchaikov:


alfredodeza:
@tchaikov can you help explain the "why" of these changes? I am not following what the objective is here

alfredodeza:
It seems that this change is needed because: using items as a list in the Jenkins Job YAML configuration file cannot be used as a string (with quotes), which otherwise will be consumed as a single item.

PR #939, opened 2017-12-13
script/build_utils.sh: import ppa key into apt's keyring

tchaikov:
to be immune from the MITM attack

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -457,8 +457,10 @@ echo "deb http://mirror.cs.uchicago.edu/ubuntu-toolchain-r $DIST main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
 echo deb http://mirror.yandex.ru/mirrors/launchpad/ubuntu-toolchain-r $DIST main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
+# import PPA's signing key into APT's keyring
+apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
 apt-get update
-apt-get install -y --allow-unauthenticated g++-7
+apt-get install -y g++-7
 EOF
 
         cat > $hookdir/E10update-gcc-alternatives <<EOF


scripts/build_utils.sh
```diff
@@ -457,8 +457,10 @@ echo "deb http://mirror.cs.uchicago.edu/ubuntu-toolchain-r $DIST main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
 echo deb http://mirror.yandex.ru/mirrors/launchpad/ubuntu-toolchain-r $DIST main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
+# import PPA's signing key into APT's keyring
+apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 1E9377A2BA9EF27F
```
tchaikov: @ktdreyer @liewegas please note, if `keyserver.ubuntu.com` is not accessible at that moment. the build is broken. that's why i added multiple mirrors for the ubuntu-toolchain-r.

ktdreyer: Who in the Ubuntu community should we contact when that happens?

tchaikov: TBH, i don't know. i just want to make sure that the reviewers understand the pros and cons. because i'd rather to avoid SPOF if possible.

ktdreyer: Ok, thanks.

If it becomes a recurring issue, we can store the entire pubkey in the job, like this:
```bash
cat << ENDOFKEY | sudo apt-key add -
-----BEGIN PGP PUBLIC KEY BLOCK-----
Version: SKS 1.1.6
Comment: Hostname: keyserver.ubuntu.com

mI0ESuBvRwEEAMi4cDba7xlKaaoXjO1n1HX8RKrkW+HEIl79nSOSJyvzysajs7zUow/OzCQp
9NswqrDmNuH1+lPTTRNAGtK8r2ouq2rnXT1mTl23dpgHZ9spseR73s4ZBGw/ag4bpU5dNUSt
vfmHhIjVCuiSpNn7cyy1JSSvSs3N2mxteKjXLBf7ABEBAAG0GkxhdW5jaHBhZCBUb29sY2hh
aW4gYnVpbGRziLYEEwECACAFAkrgb0cCGwMGCwkIBwMCBBUCCAMEFgIDAQIeAQIXgAAKCRAe
k3eiup7yfzGKA/4xzUqNACSlB+k+DxFFHqkwKa/ziFiAlkLQyyhm+iqz80htRZr7Ls/ZRYZl
0aSU56/hLe0V+TviJ1s8qdN2lamkKdXIAFfavA04nOnTzyIBJ82EAUT3Nh45skMxo4z4iZMN
msyaQpNl/m/lNtOLhR64v5ZybofB2EWkMxUzX8D/FQ==
=LcUQ
-----END PGP PUBLIC KEY BLOCK-----
ENDOFKEY
```

I think fetching it from the keyserver is fine for now.

tchaikov: thank you. will apply this change if this turns out to be a problem.

PR #942, opened 2017-12-22
script/build_utils.sh: do not download unmirrored arch

tchaikov:


tchaikov:
see also https://github.com/ceph/ceph/pull/19645

ktdreyer:
I've always been uncomfortable with the fact that we reuse the pbuilder cache tarballs at all, because it introduces a lot of persistent state into the build environments. It seems we have even more complexity now because we have to add or *clear* the new GCC in the pbuilder environment.

I think this may be a result of using an "E" prefix pbuilder hook script - any changes that run there persist in the pbuilder cache tarball, right? If we switched to a "D" prefix pbuilder hook, I think the changes would not be stored in the pbuilder cache. Then we could treat each pbuilder environment as clean/throwaway without having to clean it up ourselves.

Downstream (eg. with [rhcephpkg](https://github.com/red-hat-storage/rhcephpkg)) I've always used the "D" prefix hooks because I want pbuilder to start from a clean cache environment each time.

`pbuilder(8)` "--hookdir" section describes how each prefix works (eg. "D" vs "E").

tchaikov:
@ktdreyer thanks for the pointers. please see https://github.com/ceph/ceph-build/pull/942/commits/b42c75be6b906eb624b4d2db2bcfba7aa076617c#diff-fca3b107e8945cd175aba5640ef7360dR478 .

ktdreyer:
I've read the comments there, and I'm still failing to understand why this is so complex. It's quite possible it's my fault and I'm too dumb to understand what is going on in this code now. Sounds like I need to step through the shell commands and pbuilder invocations myself to understand what the root issue is with `build-essential`. I'm sorry I cannot review this as is because it's over my head.

tchaikov:
i will give the "D" hook a try this week-end and paste the error message if it helps with the review. IIRC, i was using "D" and ran into that issue. that's why i switched to "E".

tchaikov:
@ktdreyer i tested the 'D' hooks, it work! and i recalled that it was `--extrapackages` which fails. but after a second thought, i think it's *might* be a good thing to "introduce a lot of persistent state into the build environments". i understand that it's always expected to have a cleanroom for building packages. but downloading gcc-7 and its dependencies for every build takes extra resources and time. so probably we can cache them in the base.tgz? i don't feel strong either way. what do you think?

ktdreyer:
If it takes a long time to download, maybe we could maintain a separate mirror in OVH, since we're already relying on a static mirror list.

It would be great if the OVH build VMs were preconstructed to include warm pbuilder (and eventually yum/mock) caches.

yuriw:
@tchaikov FYI http://tracker.ceph.com/issues/22465#note-6

tchaikov:
@ktdreyer guess that's way beyond my devops-fu. anyway, does this change look better? i'd like to get it merged sooner. as the ceph-build master is broken now.

tchaikov:
@yuriw could you file a ticket? that's a different issue, i believe.

yuriw:
@tchaikov done http://tracker.ceph.com/issues/22558

ktdreyer:
Thanks, @tchaikov for looking into this and giving more detail. I think I understand a bit better now.

tchaikov:
i am merging this. i think it's for the good of our builds on jenkins at this moment.

PR #964, opened 2018-02-21
Discard builds for jobs

alfredodeza:
Enables discarding builds after 90 days for a few projects that were just keeping them for ever. 

Once these are in control, we can try and see what others can this be applied. I picked only the top 4 which were beyond 4,000 jobs
screen shot 2018-02-20 at 8 08 59 pm

### DIFF
--- a/bz-mill/config/definitions/bz-mill.yml
+++ b/bz-mill/config/definitions/bz-mill.yml
@@ -2,6 +2,10 @@
     name: bz-mill
     defaults: global
     display-name: 'bz-mill'
+    properties:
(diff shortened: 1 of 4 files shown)


alfredodeza:
Changed the commits a bit to just keep builds for 30 days. In chacra nodes we only keep builds up to 30 days so it doesn't make sense to keep them for longer.

Also, after applying these changes by hand, the almost 14,000 jobs in ceph-dev-new-build just dropped to a bit over 9,000 (with a 90 day limit)

PR #982, opened 2018-03-13
migrate from logrotate to build-discarder

alfredodeza:
build-discarder has been deprecated for a while and it is pervasive in
all jobs. Jenkins will stop supporting this as well as Jenkins Job
Builder

Job description that notes deprecation: https://docs.openstack.org/infra/jenkins-job-builder/definition.html#job

build-discarder documentation: https://docs.openstack.org/infra/jenkins-job-builder/properties.html#properties.build-discarder

### DIFF
--- a/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
+++ b/ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
@@ -105,11 +105,11 @@
     properties:
       - github:
           url: https://github.com/ceph/ceph-ansible
-    logrotate:
-      daysToKeep: 90
-      numToKeep: -1
-      artifactDaysToKeep: -1
-      artifactNumToKeep: -1
+      - build-discarder:
+          days-to-keep: 90
+          num-to-keep: -1
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
 
     parameters:
       - string:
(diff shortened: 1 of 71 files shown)


ceph-ansible-nightly/config/definitions/ceph-ansible-nightly.yml
```diff
@@ -133,6 +128,11 @@
     builders:
       - inject:
           properties-content: |
+      - build-discarder:
```
alfredodeza: this one is wrong

alfredodeza:
This is going to take a significant effort. Not only the changes for build-discarder are needed but also the ones for postbuildscript were a 'build-step' is required now

alfredodeza:
All the jobs that used `postbuildscript` and `shell` directives have been updated to use `build-steps` as well. The JJB is now finally passing.

dmick:
First paragraph really confused me.  It's logrotate that's deprecated, right?

alfredodeza:
@dmick sorry about that, you are right, I have updated the subject of the PR to reflect the correct one

djgalloway:
`jenkins-job-builder==2.0.0.0b2` doesn't like `build-step` but I confirm locally that 2.0.3 does.

https://jenkins.ceph.com/view/all/job/jenkins-job-builder/881/consoleFull

Looks like I pinned it back in https://github.com/ceph/ceph-build/commit/4f10012571b120e01854fae49be4d43f382787f1#diff-e62f3bf1d9b2d8f48c1c852d0c15c056 because the beta version worked with a config we needed.  I'll submit a PR to unpin JJB.

PR #983, opened 2018-03-14
ceph-container-build-push-imgs: use a bigger machine to build images

leseb:
We want to parallelise the build image process. Currently it takes 40min
to build 8 images, we hope to reduce this by 8 with this new flavor.


ErwanAliasr1:
Since commit https://github.com/ceph/ceph-container/commit/80f25427d0c7e86e4ab4519b6e7cda5d13d85ae6 , ceph-container is parallelized and can improve its build procedure by using the "make build.parallel". As we do have currently 7 flavors to build, I endorse this change to reduce the build time.

ErwanAliasr1:
This could be simplified once https://github.com/ceph/ceph-container/pull/962 will be merged.

PR #987, opened 2018-03-16
jenkins-job-builder: Unpin JJB version

djgalloway:
It was pinned to a beta release in https://github.com/ceph/ceph-build/commit/4f10012571b120e01854fae49be4d43f382787f1 so we could take advantage of an option that wasn't available in the stable version.

It's now in the stable version so we can pin to latest stable.

### DIFF
--- a/jenkins-job-builder/build/build
+++ b/jenkins-job-builder/build/build
@@ -9,7 +9,7 @@
 set -ex
 
 # the following two methods exist in scripts/build_utils.sh
-pkgs=( "jenkins-job-builder==2.0.0.0b2" )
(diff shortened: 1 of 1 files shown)


alfredodeza:
I would rather like pinning again to a stable version. We've gotten into trouble before by not caring on versions before. Is there a concern on pinning (to stable)?

djgalloway:
> I would rather like pinning again to a stable version. We've gotten into trouble before by not caring on versions before. Is there a concern on pinning (to stable)?

No, that makes sense.  I've repushed with it pinned to 2.0.3.

PR #988, opened 2018-03-16
sepia-fog-images: Add retry function and pause the queue

djgalloway:
I've observed deployment failures while teuthology jobs are running if
the OS image is replaced/removed while it's in use.  Pausing the queue
will allow Deploy tasks to finish so we can capture a new OS image
without interrupting running jobs.

### DIFF
--- a/sepia-fog-images/build/build
+++ b/sepia-fog-images/build/build
@@ -38,6 +38,16 @@ funPowerCycle () {
   fi
 }
 
+# There's a few loops that could hang indefinitely if a curl command fails.
+# This function takes two arguments: Current and Max number of retries.
+# It will fail the job if Current > Max retries.
+funRetry () {
+  if [ $1 -gt $2 ]; then
+    echo "Maximum retries exceeded.  Failing job."
+    exit 1
+  fi
+}
+
 # Clone or update teuthology
 if [ ! -d teuthology ]; then
   git clone https://github.com/ceph/teuthology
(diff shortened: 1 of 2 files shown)


sepia-fog-images/build/build
```diff
@@ -153,13 +153,34 @@ set -ex
...
+
+# If there are scheduled or active deploy tasks, pause the queue and let them finish.
+# Capturing a new OS image can interrupt active OS deployments.
+if [ $deploytasks -gt 0 ]; then
+  for type in $MACHINETYPES; do
+    # Only pause the queue for 20 minutes just in case anything goes wrong with the Jenkins job.
+    teuthology-queue --pause 1200 --machine_type $type
+  done
+  pausedqueue=true
+  while [ $deploytasks -gt 0 ]; do
```
alfredodeza: this will try for ever if the curl command fails. Would it make sense to try for a number of times (multiplied by the sleeps) ?

djgalloway: curl never fails :rage: 

Probably.  That's true for multiple places in the job though.  I'll add some retry maximums.

djgalloway: @alfredodeza Better?  https://github.com/ceph/ceph-build/pull/988/commits/855e1b570722d962a5059107c85da14c299726af

djgalloway:
DNM yet.  Debugging why job failed.

alfredodeza:
This looks OK. My BASH is not strong, but I am fine with this as-is. Feel free to merge once you are OK with it.

PR #991, opened 2018-03-23
ceph-{ansible,container}-prs: Fix skip-build-phrase regex

djgalloway:
JJB converts the string to lowercase so the \W was getting converted to
\w so the actual phrase wasn't being detected.  We can just use a space
instead of trying to match any character.

https://git.openstack.org/cgit/openstack-infra/jenkins-job-builder/tree/jenkins_jobs/modules/triggers.py#n1027

### DIFF
--- a/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
+++ b/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
@@ -124,7 +124,7 @@
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
-          skip-build-phrase: '^jenkins do not test.*|.*\[skip\W+ci\].*'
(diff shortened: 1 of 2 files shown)


dmick:
this sounds like a bug in jjb; has it been filed?

djgalloway:
@dmick looks intentional to me: https://git.openstack.org/cgit/openstack-infra/jenkins-job-builder/commit/?id=5c7a4e4d5c89aca57987d2304f603dd7520e4d47

dmick:
I don't see anything in the comments indicating that they intended to break REs; in fact it took a long time to find that 'lower'.  Am I missing a statement of intent about lowercasing somewhere?

ktdreyer:
it does look like an unintentional JJB bug

PR #997, opened 2018-04-01
scripts/build_utils.sh: use "download" instead of "--download"

tchaikov:
"pip install --download" was deprecated since v8.0.0, and was replaced
with "pip download". pip v10.0.0 removed support the "--download"
option. so we need to replace "install --dowload" with "download".


tchaikov:
this should address recent build failures on xenial. this pip problem impacts both pull request build and ceph-ci build.

tchaikov:
retest this please

PR #999, opened 2018-04-09
ceph-container-flake8: Using project's tox.ini

ErwanAliasr1:
When running the flake8, it's really important to reuse project's
tox.ini to insure the CI is testing the code with the same configuration
unless that brings to a situation where a local flake pass and the CI
fails.

### DIFF
--- a/ceph-container-flake8/build/build
+++ b/ceph-container-flake8/build/build
@@ -21,7 +21,7 @@ function check(){
     while read -r filename; do
         pushd "$(dirname "$filename")"
         file=$(basename "$filename")
-        sudo docker run --rm -v "$(pwd)"/"$file":/"$file" eeacms/flake8 /"$file"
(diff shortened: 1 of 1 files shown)


ceph-container-flake8/build/build
```diff
@@ -21,7 +21,7 @@ function check(){
     while read -r filename; do
         pushd "$(dirname "$filename")"
         file=$(basename "$filename")
-        sudo docker run --rm -v "$(pwd)"/"$file":/"$file" eeacms/flake8 /"$file"
+        sudo docker run --rm -v "$(pwd)"/tox.ini:/tox.ini "$(pwd)"/"$file":/"$file" eeacms/flake8 /"$file"
```
leseb: Not sure to understand why you expose tox.ini inside the container.
Also, you forgot to add another `-v`, one for each file/path you want to expose.

ErwanAliasr1: Good catch. Thx !
I updated the commit message for a better explanation. Hope that's much better.

PR #1001, opened 2018-04-09
container: add docker hub credential binding

ktdreyer:
docker-hub-leseb is defined as a credential on 2.jenkins.ceph.com

### DIFF
--- a/ceph-container-build-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
+++ b/ceph-container-build-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
@@ -39,3 +39,8 @@
       - inject-passwords:
           global: true
           mask-password-params: true
+      - credentials-binding:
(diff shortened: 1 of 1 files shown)


ceph-container-build-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
```diff
@@ -39,3 +39,8 @@
       - inject-passwords:
```
leseb: should we remove these lines too?

leseb:
Thanks for your help on this one @ktdreyer!

ceph-container-build-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
ktdreyer: That's a good idea. Let's do that in a separate PR

PR #1002, opened 2018-04-09
ceph-container: Add staging test

BlaineEXE:
Staging is an important part of the build process, and if the staging
integration test fails, it is very possible that some/all images may not
build as expected. Therefore, add the integration test to the PR
pre-flight tests.

I haven't submitted a Jenkins job before. I want to make sure that the build at least keeps the `stage.log` from the test. Do I need to configure this somewhere? Or does the `-1` entry make it keep all artifacts?

Signed-off-by: Blaine Gardner <blaine.gardner@suse.com>

### DIFF
--- a/ceph-container-stage-test/build/build
+++ b/ceph-container-stage-test/build/build
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+set -x
+
+cd "$WORKSPACE"/ceph-container/ || exit
+make test.staging
--- a/ceph-container-stage-test/config/JENKINS_URL
+++ b/ceph-container-stage-test/config/JENKINS_URL
@@ -1,0 +1,1 @@
+2.jenkins.ceph.com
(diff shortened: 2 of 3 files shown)


BlaineEXE:
I wasn't able to request reviewers, so: @leseb @ErwanAliasr1, is there anything else I should be doing here?

ErwanAliasr1:
Your PR is valid but as long as we don't have a pipeline we may miss one of its goal.

If the test.staging fails, it would be better not trying to build the whole thing and stop immediately.
To get this behavior a pipeling in the CI is needed but it doesn't seems trivial.

WDYT about adding this test in the early calls of tests/tox.sh and contrib/build-push-ceph-container-imgs.sh from ceph-container ?

Those scripts are already used by the CI and having a failure in it would stop the CI & prevent building useless stuff that will breaks after on.

BlaineEXE:
RE @ErwanAliasr1, that sounds like it would be effective.

PR #1016, opened 2018-04-19
cleanup: remove all top-level scripts

alfredodeza:
I doubt these are used today, if they are, lets see what breaks to fix/improve them. These have been around for years.


zmc:
so much red 😍

PR #1019, opened 2018-04-26
ceph-build: do not build mimic on old distros

tchaikov:
mimic and nautilus require C++17 to build, but following distro do
not ship new enough GCC compiler:

- ubuntu trusty
- debian jessie
- debian stretch

so we won't build mimic or newer ceph on these distros.

### DIFF
--- a/ceph-build/config/definitions/ceph-build.yml
+++ b/ceph-build/config/definitions/ceph-build.yml
@@ -49,35 +49,50 @@
             - ARCHS
 
     builders:
-      - shell: |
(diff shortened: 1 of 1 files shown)


alfredodeza:
@liewegas are you fine dropping Debian and Trusty for Mimic and beyond? we can always re-add them later I guess

liewegas:
Definitely fine dropping trusty.

I'd like to hear from the Debian maintainer(s) whether we can/should pull the gcc from testing to build for stretch and/or jessie.  We can drop them for now, but I suspect/hope we can add one or both of them back before mimic release...

PR #1023, opened 2018-05-03
ceph-build: use BRANCH instead of non-existing GIT_BRANCH

alfredodeza:
`GIT_BRANCH` didn't exist, so it wouldn't match and the build would be skipped for everything/anything.


alfredodeza:
This change has already been applied to produce mimic 13.1.0

alfredodeza:
@tchaikov I didn't ask you to review this directly, you will probably get to it in the morning.

PR #1029, opened 2018-05-14
rtslib-fb: reduce pyudev version requirements

dillaman:


ktdreyer:
Is this change going to rtslib-fb upstream?

PR #1033, opened 2018-05-28
ceph-ansible: remove fetch directory in teardown

guits:
Remove the fetch directory in teardown for ceph-ansible to avoid
keeping leftover from previous jobs.


scripts/build_utils.sh
```diff
@@ -760,6 +760,7 @@ teardown_vagrant_tests() {
         # collect all ceph logs from all test nodes
         collect_ceph_logs all
         vagrant destroy -f
+        stat ./fetch > /dev/null 2>&1 && rm -rf ./fetch
```
leseb: Why not simply `rm -rf ./fetch`?

guits: Indeed, too many sleepless night I guess :-|

PR #1035, opened 2018-05-30
build_utils.sh: add ubuntu-ports repo for arm64 build

tchaikov:
as it offers updated binutils which does not exist in the default repo.
and the updated binutils is depended by gcc-7. this change address
following failure like:

The following packages have unmet dependencies:
 gcc-7 : Depends: binutils (>= 2.26.1) but 2.26-8ubuntu2 is to be
installed

when installing g++-7.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -510,8 +510,8 @@ update-alternatives \
 update-alternatives --auto gcc
 
 # cmake uses the latter by default
-ln -nsf /usr/bin/gcc /usr/bin/x86_64-linux-gnu-gcc
-ln -nsf /usr/bin/g++ /usr/bin/x86_64-linux-gnu-g++
+ln -nsf /usr/bin/gcc /usr/bin/\$(arch)-linux-gnu-gcc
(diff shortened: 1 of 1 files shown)


theanalyst:
@tchaikov the link you posted shows a failure even after g++-7 was installed, do we need to explicitly set compilers for cmake as well?

scripts/build_utils.sh
```diff
@@ -510,8 +510,8 @@ update-alternatives \
 update-alternatives --auto gcc
 
 # cmake uses the latter by default
-ln -nsf /usr/bin/gcc /usr/bin/x86_64-linux-gnu-gcc
-ln -nsf /usr/bin/g++ /usr/bin/x86_64-linux-gnu-g++
+ln -nsf /usr/bin/gcc /usr/bin/\$(arch)-linux-gnu-gcc
```
alfredodeza: where is `$(arch)` coming from? I can't find where it is defined in this file

tchaikov: please note, `$(arch)` != `${arch}`.

PR #1036, opened 2018-05-30
ceph-ansible: add some tasks for debugging purpose

guits:
related to ceph/ceph-ansible/pull/2665 and ceph/ceph-ansible/pull/2652

We are seeing too many failures related to yum.
These additional tasks will help us to have more information when a yum
failure occurs.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -729,6 +729,18 @@ write_collect_logs_playbook() {
 - hosts: all
   become: yes
   tasks:
+    - name: get yum repoinfo result
+      command: yum repoinfo
+      failed_when: false
+
+    - name: get date result
+      command: date
+      failed_when: false
+
+    - name: get yum check-update result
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -729,6 +729,18 @@ write_collect_logs_playbook() {
 - hosts: all
   become: yes
   tasks:
+    - name: get yum repoinfo result
```
alfredodeza: these `yum` commands are now failing on every non-yum test. Although it is getting ignored because of the `failed_when: false`, it is distracting to see these run, fail, and get ignored on Debian-based tests

guits: @alfredodeza this is not supposed to stay, but we could add a when so these are run only on CentOS?

alfredodeza: Yeah, sure, that would help!

PR #1044, opened 2018-06-07
Implement ceph-ansible pipeline

djgalloway:


ceph-ansible-docs-prs/config/definitions/ceph-ansible-docs-prs.yml
```diff
@@ -28,7 +28,8 @@
           org-list:
             - ceph
           trigger-phrase: 'jenkins test docs'
-          only-trigger-phrase: false
+          # This is set so the job can be manually triggered or by the ceph-ansible-pipeline multijob
+          only-trigger-phrase: true
```
alfredodeza: this doesn't read right (probably my misunderstanding), but I thought `only-trigger-phrase: false` meant you can trigger with a comment in github *as well as other triggers*. Are you sure this is doing what the comment suggests?

djgalloway: `only-trigger-phrase: false` means the job runs for every PR that gets opened/repushed was my understanding.

djgalloway: Now I'm thinking I probably need `permit-all` set to false as well.

djgalloway: `permit-all` just allows any Github user to trigger the job.  It should be false anyway.

But to further clarify my response, `only-trigger-phrase: true` means only the trigger phrase can trigger the job **as far as the Github Pull Request Builder plugin is concerned**.  The job can still be triggered manually by another job.

djgalloway:
@leseb Sure, I can merge Sunday.

PR #1048, opened 2018-06-19
ceph-deploy-build: remove centos6

alfredodeza:


liewegas:
:+1:

Should trusty still be in there?

alfredodeza:
we just need to build in one really, these packages are arch agnostic and aren't tied to the distro at all, we actually include them in all the ceph repos for Debian-based builds (e.g. it will exist in debian-mimic regardless)

For correctness sake, I've updated to use xenial only (again, it doesn't really matter)

PR #1049, opened 2018-06-19
ceph-volume-nightly: remove master from simple, ceph-disk no longer available

alfredodeza:


alfredodeza:
@andrewschoen this will remove master for the simple sub-command, and adds mimic testing for both simple and master (which were missing)

PR #1051, opened 2018-06-20
ceph-ansible: move bluestore_osds_container in ceph-ansible-prs-auto

guits:
we are seeing this scenario failing because of the race condition,
moving it to a project which requires smithi nodes should help.


djgalloway:
What's the race condition?  Can you paste log output or details just so we have it documented?

PR #1053, opened 2018-06-25
ceph-ansible-pr-syntax-check: fix grep error

leseb:
If grep fails the script exits because of the set -e so let's use ||
true to workaround that issue.


guits:
thanks maggle

PR #1056, opened 2018-06-27
[WIP] Added Jenkins job for dashboard end-to-end tests

callithea:

The "ceph-dashboard-pull-requests.yml" is based on the [ceph-pull-requests.yml](https://github.com/ceph/ceph-build/blob/master/ceph-pull-requests/config/definitions/ceph-pull-requests.yml) file with some modifications.

Signed-off-by: Laura Paduano lpaduano@suse.com

### DIFF
--- a/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
+++ b/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
@@ -1,0 +1,69 @@
+- job:
+    name: ceph-dashboard-pull-requests
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: huge && (centos7 || trusty) && rebootable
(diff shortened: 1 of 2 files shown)


alfredodeza:
We don't normally use open pull requests for active development. Could you close this and re-open when you feel this is ready? Otherwise it is triggering tests that aren't really needed.

callithea:
Sure, I'll close it. @alfredodeza is there another way to test this beforehand?

I found [this section](https://github.com/ceph/ceph-build#testing-jjb-changes-by-hand-before-merging-to-master) in the README but it's about testing JJB changes so I wasn't sure if it would also make sense for a new Jenkins job?

alfredodeza:
There is no way to test this beforehand :(

I would recommend to get this as ready as you can, and then we can review and iterate over if you need additional fixes that were not detected in the review process

callithea:
Okay, thanks! :)

PR #1057, opened 2018-06-27
ceph-pull-requests-arm64: Normalize name to appease JJB

djgalloway:
Not sure why `!!python/unicode` was there originally but the current
version of JJB doesn't like it.


djgalloway:
```
<dmick_> Yeah, no idea.  Looks like an artifact of some auto dumper
```

ktdreyer:
maybe a Jenkins Job Wrecker bug :laughing:

dmick:
it might have been, actually (well, not a bug, I think that's valid yaml, but...)

joscollin:
There was no such problem with the build 2 days ago :) 
Anyway, thanks @djgalloway.

PR #1058, opened 2018-06-28
ceph-pull-requests: Improving ccache efficiency

ErwanAliasr1:
As of today, ccache is installed by run-make-check but the statistics
reports almost no cache it (<1%) and even worse, this is slowing down
the whole process. The needed time to setup the cache isn't compensated
by the reuse of the cache as every build is rebuilding the cache. This
is leading to a 10% loss of the compilation time.

So the question, is why does two consecutives builds doesn't reuse the
cache. Its usually a matter of variables introduces on each builds like
a sha1 or a __DATE__ or __TIME__ in a header.

As per
https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal, several
tools from the tool chain (like compilers or make tools) are able to use
theses variables to enforce a particular timestamp at build time.

This commit also add a "-D ENABLE_GIT_VERSION=OFF" to avoid having the
commit id in the ceph_ver.h header. Not having the git versioning inside
the CI job is not a by deal as the binaries will be destroyed after the
build procedure.

These changes avoid introducing small variations in the source code
which breaks the build cache (ccache). The expected result is to get a
80 or 90% cache hit between two very close builds.

Signed-off-by: Erwan Velu <erwan@redhat.com>

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -51,7 +51,7 @@
           wipe-workspace: true
 
     builders:
-      - shell: "export NPROC=$(nproc); timeout 7200 ./run-make-check.sh"
+      - shell: "export SOURCE_DATE_EPOCH=$(date +%D |date -f- +%s); export NPROC=$(nproc); timeout 7200 ./run-make-check.sh -D ENABLE_GIT_VERSION=OFF"
 
     publishers:
       - postbuildscript:


alfredodeza:
This looks OK to me. At some point we need to move that shell line into a file because it is getting harder to read.

ErwanAliasr1:
@tchaikov As per my local tests, its enough to publish it as the shell variable to get it consumed.

tchaikov:
@ErwanAliasr1 i am not sure who is consuming this en var.

ErwanAliasr1:
@tchaikov You are right. I wasn't able to find any. I removed it.

gregsfortytwo:
Had an irc discussion that referenced this PR, so I just want to be clear: this is just changing the yaml config file for the jenkins job running "make check", right?
We definitely need to keep the sha1 versions on for release builds, and almost certainly for teuthology test run ones as well!

ErwanAliasr1:
lgtm ?

tchaikov:
@alfredodeza JFYI, i also updated the jjb jobs of `ceph-pull-requests` and `ceph-pull-requests-arm64` with this setting.

dmick:
The reason this works, unstated here, is that gcc itself uses SOURCE_DATE_EPOCH to derive __DATE__ and __TIME__ macros.  As far as I can tell this does not apply to clang, who appear to have added a new flag "-ffixed-date-time=" instead.  See https://reviews.llvm.org/D23934

PR #1062, opened 2018-06-30
mgr/dashboard: Removes extra attributes from Cobertura plugin

a2batic:
Due to some build failures, this PR removes attributes for Cobertura that are not required right now.
Signed-off-by: Kanika Murarka <kmurarka@redhat.com>


dmick:
I don't know how to reasonably review this without someone documenting what the options mean; there's no way to tell if this will be better or worse

joscollin:
Closing because of inactivity. No specific requirements available, to make a decision to keep or remove those values.

PR #1067, opened 2018-07-03
Syntax head 2

leseb:


leseb:
It's coming from jenkins's injectedEnvVars. Amending.

alfredodeza:
This doesn't look quite right. We already have jobs that poke at the commits from a PR:

https://github.com/ceph/ceph-build/blob/master/ceph-pr-commits/build/test_commits.py

As you can see, that implementation tries a bunch of different things

PR #1069, opened 2018-07-03
ceph-ansible-pr-syntax-check: return 1 only on failure

leseb:
This was a leftover from an initial attempt with if combinaison. We now
return 1 only if the test line fails.

### DIFF
--- a/ceph-ansible-pr-syntax-check/build/build
+++ b/ceph-ansible-pr-syntax-check/build/build
@@ -50,8 +50,8 @@ function group_vars_check {
 function test_sign_off {
   git fetch
   # shellcheck disable=SC2154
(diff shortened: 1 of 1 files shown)


ceph-ansible-pr-syntax-check/build/build
```diff
@@ -50,8 +50,7 @@ function group_vars_check {
 function test_sign_off {
   git fetch
   # shellcheck disable=SC2154
-  test "$(git log --oneline --no-merges origin/"${ghprbTargetBranch}"..HEAD | wc -l)" -ne "$(git log --no-merges origin/"${ghprbTargetBranch}"..HEAD | grep -c Signed-off-by)" && echo "One or more commits is/are missing a Signed-off-by. Add it with 'git commit -s'."
-  return 1
+  test "$(git log --oneline --no-merges origin/"${ghprbTargetBranch}"..HEAD | wc -l)" -ne "$(git log --no-merges origin/"${ghprbTargetBranch}"..HEAD | grep -c Signed-off-by)" && echo "One or more commits is/are missing a Signed-off-by. Add it with 'git commit -s'." && return 1
```
ErwanAliasr1: Can you explicit where ghprbTargetBranch is coming from ?

leseb: Done.

PR #1075, opened 2018-07-06
Added Jenkins job definition and setup for ceph dashboard frontend (e2e tests)

callithea:

The "[ceph-dashboard-pull-requests.yml](https://github.com/ceph/ceph-build/blob/master/ceph-pull-requests/config/definitions/ceph-pull-requests.yml)" is based on the ceph-pull-requests.yml file with some modifications.

TODO: rebase/squash commits

``Signed-off-by: Laura Paduano lpaduano@suse.com``

### DIFF
--- a/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
+++ b/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
@@ -1,0 +1,68 @@
+- job:
+    name: ceph-dashboard-pull-requests
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: huge && (centos7 || trusty) && rebootable
+    display-name: 'ceph: dashboard Pull Requests'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          num-to-keep: 300
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: https://github.com/ceph/ceph/
+
+    parameters:
+      - string:
+          name: sha1
+          description: "A pull request ID, like 'origin/pr/72/head'"
+
+    triggers:
+      - github-pull-request:
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          white-list-labels:
+            - dashboard
+          trigger-phrase: 'jenkins test make check'
+          skip-build-phrase: '^jenkins do not test.*'
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "make check"
+          started-status: "running make check"
+          success-status: "make check succeeded"
+          failure-status: "make check failed"
+
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph.git
+          branches:
+            - ${sha1}
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
+          skip-tag: true
+          wipe-workspace: true
+
+    builders:
+      - shell: "export NPROC=$(nproc); timeout 7200 ./run-make-check.sh"
+      - shell: !include-raw ../../setup/setup
+      - shell: "timeout 7200  ./src/pybind/mgr/dashboard/run-frontend-e2e-tests.sh"
+
+    publishers:
+      - postbuildscript:
+          builders:
+            - role: SLAVE
+              build-on:
+                - FAILURE
+                - ABORTED
+              build-steps:
+- shell: "sudo reboot"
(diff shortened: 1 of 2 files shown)


alfredodeza:
These commits need to be rebased (there is a merge commit in the history), and squashed. We also try to prefix the commits with the name of the job we are changing/adding:

    ceph-dashboard-pull-requests: some commit message here

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -0,0 +1,72 @@
...
+          description: "A pull request ID, like 'origin/pr/72/head'"
+
+    triggers:
+      - github-pull-request:
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          white-list-labels:
+            - dashboard
+          trigger-phrase: 'jenkins test make check'
```
alfredodeza: this trigger phrase doesn't look right, this is not make check. I would make an ad-hoc phrase here

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -0,0 +1,72 @@
...
+          skip-build-phrase: '^jenkins do not test.*'
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "make check"
```
alfredodeza: none of these "make check" statuses look right

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -0,0 +1,72 @@
...
+          branches:
+            - ${sha1}
+          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          browser: auto
+          timeout: 20
+          skip-tag: true
+          wipe-workspace: true
+
+    builders:
+      - shell: "export NPROC=$(nproc); timeout 7200 ./run-make-check.sh"
```
alfredodeza: why is this running make check? there is already a make check job

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -0,0 +1,72 @@
...
+
+    publishers:
+      - postbuildscript:
+          builders:
+            - role: SLAVE
+              build-on:
+                - FAILURE
+                - ABORTED
+              build-steps:
+                - shell: "sudo reboot"
```
alfredodeza: why does it need to reboot?

callithea: I'll remove it if it's not necessary. Do I need the "publishers" section at all?

alfredodeza: I don't know? If you think this job needs to reboot: yes. I just can't tell why would it need to reboot

callithea: I was not sure if the reboot does anything special (like resetting/cleaning-up the system as preparation for the next build or so).
But I'll remove this section for now.

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -26,22 +26,25 @@
           description: "A pull request ID, like 'origin/pr/72/head'"
 
     triggers:
+      - reverse:
```
alfredodeza: I've never seen this before, could you explain a bit what is going on with this?

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -26,22 +26,25 @@
...
+            jobs: 'ceph-pull-requests'
+            result: 'success'
       - github-pull-request:
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
           white-list-labels:
             - dashboard
-          trigger-phrase: 'jenkins test make check'
+          trigger-phrase: 'jenkins test dashboard'
```
alfredodeza: this looks correct now!

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
callithea: Sure! 
(Found this in the documentation:
https://docs.openstack.org/infra/jenkins-job-builder/triggers.html?highlight=triggers#triggers.reverse)

So my thought here was if make check passes based on a (dashboard-related) PR we could trigger the ceph-dashboard-pull-requests Jenkins job.
But what I'm not sure/wondering about here is if this will now trigger it for every PR or if it will only trigger the Job if the ceph-pull-requests Job passed AND the dashboard label I've defined in line 36/37 is assigned to the PR.

alfredodeza: yeah, I think this will cause this to run for both almost all the time. I would just leave it as an ad-hoc call to trigger the tests. This is how we do it for ceph-volume: we don't need to run on every PR, just on a selected few

callithea:
See new PR #1086 (sorry about this :( )

PR #1077, opened 2018-07-11
Use ansible 2.5 with master and 2.4 with stable-3.x

djgalloway:
Again, surely there's a more elegant way to do this but this works and doesn't rely on any additional plugins.

I tried setting the job names like `ceph-ansible-prs-luminous-$ANSIBLE_VERSION-purge_bluestore_osds_non_container` so I wouldn't have to duplicate code but `$ANSIBLE_VERSION` doesn't get set globally if I run it during a step in the first `multijob`.

The new conditionals could be boiled down to `test [[ "$ghprbTargetBranch" =~ stable-3.[0-1] ]]` and `on-evaluation-failure` set accordingly but I think since this job yaml is already confusing, the larger functions make it somewhat clearer.


guits:
Closing in favor of #1082

PR #1078, opened 2018-07-12
build_utils: should not use "sudo" in pbuilder hook script

tchaikov:

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -548,7 +548,7 @@ EOF
         exit 1
     fi
 cat >> $hookdir/D05install-gcc-7 <<EOF
-cat << ENDOFKEY | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


alfredodeza:
Interesting, could you expand on why removing `sudo` is preferred? what is the side effect?

tchaikov:
@alfredodeza pbuilder is a clean-room environment, and `sudo` is not available in a base system in debian before it is installed. so we will end up with
```
/tmp/hooks/D05install-gcc-7: 5: /tmp/hooks/D05install-gcc-7: sudo: not found
```

see https://jenkins.ceph.com/job/ceph-dev-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=xenial,DIST=xenial,MACHINE_SIZE=huge/18111//consoleFull

PR #1086, opened 2018-07-23
ceph-dashboard-pull-requests: Jenkins job for ceph dashboard frontend (e2e) tests

callithea:

The "[ceph-dashboard-pull-requests.yml](https://github.com/ceph/ceph-build/blob/master/ceph-pull-requests/config/definitions/ceph-pull-requests.yml)" is based on the ceph-pull-requests.yml file with some modifications.

``Signed-off-by: Laura Paduano lpaduano@suse.com``

### DIFF
--- a/ceph-dashboard-pull-requests/setup/setup
+++ b/ceph-dashboard-pull-requests/setup/setup
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+if grep -q  debian /etc/*-release; then
+    echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list
+    wget https://dl.google.com/linux/linux_signing_key.pub
+    apt-key add linux_signing_key.pub
+    sudo apt-get update
+    sudo apt-get install -y google-chrome-stable
+
+elif grep -q rhel /etc/*-release; then
+    cat <<-EOF > /etc/yum.repos.d/google-chrome.repo
+		[google-chrome]
+		name=google-chrome
+		baseurl=http://dl.google.com/linux/chrome/rpm/stable/x86_64
+		enabled=1
+		gpgcheck=1
+		gpgkey=https://dl.google.com/linux/linux_signing_key.pub
+	EOF
+    yum install -y google-chrome-stable
+fi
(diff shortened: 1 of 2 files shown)


alfredodeza:
This looks OK to me, not sure if you need anything else here @joscollin ? otherwise feel free to merge

ceph-dashboard-pull-requests/setup/setup
```diff
@@ -0,0 +1,22 @@
...
+    cat << EOF > /etc/yum.repos.d/google-chrome.repo
+        [google-chrome]
+        name=google-chrome
+        baseurl=http://dl.google.com/linux/chrome/rpm/stable/x86_64
+        enabled=1
+        gpgcheck=1
+        gpgkey=https://dl.google.com/linux/linux_signing_key.pub
+EOF
+    yum install -y google-chrome-stable
+fi
```
joscollin: <jcollin> Why the setup only specify google-chrome? 
<laura__> Most of the time we tested the GUI with Google Chrome
<laura__> But we can try to run the tests with Firefox as well and if it works, we could add it to the Jenkins job as well
<laura__> But in the past we had problems to write the tests in a way so that Firefox and Chrome both work
<laura__> But I think it would be great if we could cover more browsers in the future
<jcollin> ok laura__ , let's go with chrome for this PR and we will add the others later.
<jcollin> laura__ : Only some alignment changes in setup.
<jcollin> 1. Extra line after if block
<jcollin> 2. align that EOF, to stay inside if block?
<jcollin> 3. Align the elif block properly.

callithea: I hope I addressed all your requested changes.

ceph-dashboard-pull-requests/setup/setup
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+if grep -q  debian /etc/*-release; then
+    echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" > /etc/apt/sources.list.d/google-chrome.list
```
ktdreyer: yikes, plain http

callithea: I've created PR #1135 to fix this

PR #1103, opened 2018-08-22
ceph-pull-requests-arm64: set jobs number according to free memory

tchaikov:
guess we are using more compile-time optimizations now, to compile
ceph source requires more memory. sometimes, a single cc1plus takes
more than 3GB memeory. that's why we are seeing more and more OOM
in our arm64 builders. following is a sample from omani09 -- a
arm64 builder compiling a ceph/ceph PR targeting master:
```
30474 jenkins+  20   0 2573092 2.388g  16220 R 100.0  3.8   9:11.04 cc1plus
34339 jenkins+  20   0 3279768 3.005g  16296 R 100.0  4.8   5:07.63 cc1plus
36382 jenkins+  20   0  322276 278932  12740 R 100.0  0.4   0:09.78 cc1plus
...
```
also, the performance of over-all compiling is also impacted by the I/O
subsystem. so lower the number of job could actually reduce the time
of the compiling processes to completing for the I/O queue of local
device. so we can use an conservative number for calc an upper bound of
job number for "make" instead using $(nproc). in this change,
$(free_memory_in_mega / 1800) is used as the upper limit of n_jobs.

### DIFF
--- a/ceph-build/build/build_deb
+++ b/ceph-build/build/build_deb
@@ -74,16 +74,17 @@ echo deb vers $bpvers
 echo building debs for $DIST
 
 CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS $(extra_cmake_args)"
+DEB_BUILD_OPTIONS="parallel=$(get_nr_build_jobs)"
 
 # pass only those env vars specifically noted
 sudo \
     CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS" \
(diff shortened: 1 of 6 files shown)


tchaikov:
@liewegas i recall that you were asking about this over `#ceph-devel` the other day.

dmick:
I guess it's as good a guess as any, but, should we also change the ceph-build job for the actual build?

tchaikov:
you mean the jenkins-job ? i updated it yesterday. now it works fine so far.

dmick:
they're both Jenkins jobs, right, I just mean rather than the pull-request job, also the actual "build the release" job

PR #1106, opened 2018-08-22
has not been touched since 2016 has not passed in recent memory

ChristinaMeno:


leseb:
In our to make the build pass we need to add the `jenkins-build` key on the galaxy repositories, any idea how to achieve that?

djgalloway:
> In our to make the build pass we need to add the `jenkins-build` key on the galaxy repositories, any idea how to achieve that?

@leseb I'm cleaning up ceph-build.git PRs.  Is this job still desired?  I can try to get it working again if so.

djgalloway:
None of these repos even exist.

https://github.com/ceph/ceph-ansible/blob/master/contrib/push-roles-to-ansible-galaxy.sh#L7

Can we delete this job?

PR #1116, opened 2018-09-06
ceph-*-build: evaluate num jobs using total mem not free mem

tchaikov:
there is chance that the system is under extremely high load when `vmstat`
is called. so the "free memory" can not reflect the available memroy
when building Ceph. so use "total memory" instead.


tchaikov:
in https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=bionic,DIST=bionic,MACHINE_SIZE=huge/14304//consoleFull

we have `make -j1`, which was way too slow..

PR #1120, opened 2018-09-11
ceph-ansible-pr-syntax-check: add forbidden keywords

rishabh-d-dave:
Since we are replacing 'static' and 'include' throughout the ceph-ansible codebase, it would be nice to update the test so that we can check if 'static' and 'include' are being used in new PRs.

### DIFF
--- a/ceph-ansible-pr-syntax-check/build/build
+++ b/ceph-ansible-pr-syntax-check/build/build
@@ -77,6 +77,37 @@ function test_ceph_release_in_ceph_default {
   echo "No '- ceph_release_num[ceph_release]' statements found in ceph-defaults role!" && return 0
 }
 
+function check_if_a_keyword_is_used() {
+  searchstring=$1
+
+  # Since we are looking for keywords, instances of $searchstring found
+  # in comments and strings are not qualified. Ansible keywords can only
+  # be preceded by bunch of spaces or with a hypen and a space in conjunction.
+  # So, let's only look for those instances.
+  #
+  # Besides, completely ignore the lines marked as deleted in diff.
+  searchstring_regex="^\+ *-? +$searchstring"
+  count=$(git diff master..HEAD | grep -cE "$searchstring_regex")
+  if [[ $count > 0 ]]; then
+    echo "\"$searchstring\" is amongst the forbidden keywords; found $count times"
+    return $count
+  fi
+
+  return 0
+}
+
+function check_for_forbidden_keywords() {
+  check_if_a_keyword_is_used "include:"; forbidden_keyword_count=$?
+  check_if_a_keyword_is_used "static:"; forbidden_keyword_count=$?
+
+  if [[ $forbidden_keyword_count != 0 ]]; then
+    return 1
+  fi
+
+  echo "Forbidden keyword(s) not found."
+  return 0
+}
+
 
 ########
 # MAIN #
@@ -87,3 +118,4 @@ syntax_check
 group_vars_check
 test_sign_off
 test_ceph_release_in_ceph_default
+check_for_forbidden_keywords


alfredodeza:
How could this detect something that is not part of the actual syntax, but a comment, or a string?

rishabh-d-dave:
> How could this detect something that is not part of the actual syntax, but a comment, or a string?

Yeah, I did not think of that. I can add some code to pick the particular line where these keywords occur followed by a check to see if those lines are, partly or completely, comment or string and if the keyword is a part of it. But this feels unwieldy.

alfredodeza:
The problem here is that this seems like a hard check, if it is catching a string or a comment, it will be impossible to merge because this will keep failing. Even worse: in the case of an *existing* comment that contains any of these words, that is being changed, it will always fail.

This has happened before with https://github.com/ceph/ceph-build/pull/1097 where it enforces lower case for everything. It kept failing when we were trying to submit a PR that removed upper case letters.

rishabh-d-dave:
> The problem here is that this seems like a hard check, if it is catching a string or a comment, it will be impossible to merge because this will keep failing. Even worse: in the case of an existing comment that contains any of these words, that is being changed, it will always fail.

Won't adding some code that will check and exclude the case were the keywords are a part of comment or string would help?

> This has happened before with #1097 where it enforces lower case for everything. It kept failing when we were trying to submit a PR that removed upper case letters.

That shouldn't be the case here. I am doing a `git grep` and not a `grep` on git commits/diffs. Perhaps that needs to be changed there.

rishabh-d-dave:
Also, do I need to add a check to my test function to know whether the "PR branch" is the current branch, or that is assured?

ceph-ansible-pr-syntax-check/build/build
```diff
@@ -82,6 +82,46 @@ function test_ceph_release_in_ceph_default {
...
+  if [ "$2" -lt 1 ]; then
+    exit -2
+  elif [ "$2" -eq 1 ]; then
+    pathspec="HEAD"
+  else
+    pathspec="HEAD~"$(($2 - 1))
+  fi
+
+  count=0
+  output=$(git grep -c "$searchstring" $pathspec | awk -F ":" '{print $3}')
```
rishabh-d-dave: `git grep` would search the entire codebase and not just the lines added/modified.

rishabh-d-dave:
> How is this code going to be able to tell what is a comment from a string?

Right now it won't be able to. I have written some code that could also work with these cases (as I said it [here](https://github.com/ceph/ceph-build/pull/1120#issuecomment-420354948)). I'll update this PR shortly.

rishabh-d-dave:
alfredodeza [wrote](https://github.com/ceph/ceph-build/pull/1120#issuecomment-420368026):
>>    How is this code going to be able to tell what is a comment from a string?
>
> rishabh-d-dave [wrote](https://github.com/ceph/ceph-build/pull/1120#issuecomment-420643791):
> Right now it won't be able to. I have written some code that could also work with these cases (as I said it here). I'll update this PR shortly.

Updated the PR. The code will search for keywords in `git diff`'s output  by passing a regex to grep that matches if and only if a sequence of zero or more spaces, optionally, a hyphen and another sequence of one or more spaces precedes the keyword. IMO, this should exclude instances in comments and strings.

And, the lines marked as deleted by `git diff` would be ignored so the case where deleted lines matches the pattern is excluded.  If I understood correctly, this was the case you were mentioning [here](https://github.com/ceph/ceph-build/pull/1120#issuecomment-420368026)

ceph-ansible-pr-syntax-check/build/build
```diff
@@ -82,6 +82,37 @@ function test_ceph_release_in_ceph_default {
...
+  count=$(git diff master..HEAD | grep -cE "$searchstring_regex")
+  if [[ $count > 0 ]]; then
+    echo "\"$searchstring\" is amongst the forbidden keywords; found $count times"
+    return $count
+  fi
+
+  return 0
+}
+
+function check_for_forbidden_keywords() {
```
leseb: Is this function necessary? Why don't we simply call `check_if_a_keyword_is_used` out of a function here https://github.com/ceph/ceph-build/pull/1120/files#diff-6cbf1ad8bb8686f24d82a5d4412b4311R127? If you find something forbidden in `check_if_a_keyword_is_used` then simply return `1` instead of `count` otherwise just keep it 0.

What do you think?

rishabh-d-dave: I've written two functions because I think `check_if_a_keyword_is_used` might come into use separately (since it's purpose is more generic). However, if we don't want that I can merge them into one. Should I do so?

rishabh-d-dave:
@leseb
Can you please review this PR? Relevant ceph-ansible PRs have been merged.

rishabh-d-dave:
@alfredodeza  @guits  @leseb 
Can we merge this PR, it's  complete in your opinion?

leseb:
@rishabh-d-dave these kinds of checks tend to break the CI more than anything else in the end. So we might spend more time on re-working them constantly and at the same time blocking the ceph-ansible CI. Soon we will be testing on 2.7 and we always keep a good pace with the Ansible releases. So if someone introduces a keyword that is not supported this will fail on the master PR anyway.

I'm in favor of closing this, @guits anything to add?

guits:
I'd say I'm -1 on merging this.
For instance, see https://github.com/ceph/ceph-build/pull/1144

It's generally complicated to have a such a check relying on regexp because it can report false positives at some point.

I'm not really trying to block this PR since I don't have any other option at the moment, just want to raise this concern.

rishabh-d-dave:
Maybe we can run this job but not let it block the given PR? It would only serve as a note to the PR's author, which would help the author adopt the preferred coding style which in turn improves the chances that our codebase is consistent.

leseb:
@rishabh-d-dave in fact if we don' fail, no one is going to look at the syntax sequence.

rishabh-d-dave:
At least reviewers would know without having to look at the code. I don't know if this possible but may the CI job or a bot (if we have one) can post a comment with a TODO list task to get rid of un-preferred keywords.

I am trying to emphasize on this because if we don't indicate about these keywords, someone will end up using them and "replace x keywords by y"  PRs won't serve their purpose well.  :)

guits:
if I'm not mistaking, ansible already warns users about usage of deprecated keywords.

guits:
@rishabh-d-dave the deprecation warning didn't show up in your example because you used `import_tasks`, therefore `static` is ignored, it means it would be more a cosmetic issue, I guess.

rishabh-d-dave:
Yeah, actually.

rishabh-d-dave:
If there is some way to indicate preferred ansible keywords on the PR it would be nice, otherwise we can proceed to close this PR.

leseb:
@rishabh-d-dave not possible, let's close this.

rishabh-d-dave:
yep, no problem.

PR #1121, opened 2018-09-12
Change build discard options for ceph and ceph-volume jobs

alfredodeza:
So that it isn't limited by the number, it is limited by days only. This
will allow builds to avoid 404'ing when the limit is reached, which is
unexpected if there are too many builds and not long enough time between
them (like 100 builds in one day, but limit is 20 before discarding)

### DIFF
--- a/ceph-pr-arm-trigger/config/definitions/ceph-pr-arm-trigger.yml
+++ b/ceph-pr-arm-trigger/config/definitions/ceph-pr-arm-trigger.yml
@@ -11,8 +11,7 @@
     block-upstream: false
     properties:
       - build-discarder:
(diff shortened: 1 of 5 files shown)


tchaikov:
@epuertat agreed. one way to DRY is to a template to extract the shared settings out of the job descriptions. and the jenkins job description can always use some refactory, but we are just not putting lots of resource/thoughts/love in it at this moment i guess..

alfredodeza:
@epuertat these are separate projects, so no way to try anything DRY. For more common projects, like @tchaikov said, we have templates. 

When the limits are reached, the logs/builds are fully removed, which is why you hit a 404.

PR #1131, opened 2018-09-19
ceph-dev-[new-]setup/build/build: -DWITH_CEPH_DEBUG_MUTEX=ON for notcmalloc

liewegas:
Enable the debug ceph::mutex in the notcmalloc build.  The normal build
will use the defaults, which (if it's a Release or RelWithDebug build)
will not include extra debug checks.

Signed-off-by: Sage Weil <sage@redhat.com>


tchaikov:
see also https://github.com/ceph/ceph/pull/24133

alfredodeza:
Is there no need to special case releases? Or is this supposed to work all the way back to Luminous?

tchaikov:
@alfredodeza yes, cmake will by annoyed if an `-D` variable is not used at all. but it will just spit a warning message.

probably a better way to go is to use `CMAKE_BUILD_TYPE=Debug` as an alias of `-DALLOCATOR=libc` plus `-DWITH_CEPH_DEBUG_MUTEX=ON` ?

just like GCC uses `-O` as a combination of a bunch of optimization options.

liewegas:
I was originally going go make this depend on Debug vs Release/RelWithWhatever but the notcmalloc builds aren't actually Debug builds... and I wasn't sure if the 'basic' builds were Release-ish builds.
Maybe they should be though?  That might be simpler!

liewegas:
There is no instance of CMAKE_BUILD_TYPE in ceph-build except for nfs-ganesha.  I didn't want to tangle with any magic in rpmbuild...  this i'm pretty confident will work.

PR #1139, opened 2018-09-29
ceph-dashboard-pull-requests: fix permission denied error

tchaikov:
as stdout redirect is performed by the shell, and shell is launched by
the "jenkins" user, so permission denied is expected when we are trying
to create a file under /etc/yum.repos.d/. in that case, we'll have
following failure:

/tmp/jenkins8899238112048417478.sh: line 14:
/etc/yum.repos.d/google-chrome.repo: Permission denied


tchaikov:
cc @callithea

PR #1140, opened 2018-10-01
add new ceph-container-build-ceph-base-push-imgs job

leseb:
This job runs on a daily basis and is responsible for building ceph/base
versionned images.

### DIFF
--- a/ceph-container-build-ceph-base-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
+++ b/ceph-container-build-ceph-base-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
@@ -1,0 +1,46 @@
+- job:
+    name: ceph-container-build-ceph-base-push-imgs
+    node: huge && trusty && x86_64
+    project-type: freestyle
+    defaults: global
+    display-name: 'ceph-container: build and push ceph base container images to Docker Hub'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 1
+          num-to-keep: 1
+          artifact-days-to-keep: 1
+          artifact-num-to-keep: 1
+      - github:
+          url: https://github.com/ceph/ceph-container
+
+    triggers:
+      - timed: '@daily'
+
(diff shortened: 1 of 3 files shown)


ceph-container-build-ceph-base-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
```diff
@@ -0,0 +1,46 @@
...
+    project-type: freestyle
+    defaults: global
+    display-name: 'ceph-container: build and push ceph base container images to Docker Hub'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: -1
```
alfredodeza: `-1` means this will be kept forever. Please don't keep logs and build artifacts forever, set a reasonable number in days or builds that you need and set them here

leseb: Right, thanks, let me adjust that.

ceph-container-build-ceph-base-push-imgs/config/definitions/ceph-container-build-push-imgs.yml
```diff
@@ -0,0 +1,46 @@
...
+    defaults: global
+    display-name: 'ceph-container: build and push ceph base container images to Docker Hub'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 1
+          num-to-keep: 1
```
alfredodeza: `days-to-keep` and `num-to-keep` can be mutually exclusive: whichever hits first will trigger the removal. If you want a whole day of logs, these settings will remove all except the last one. If that is fine to you, then these would make sense.

leseb: Yes, I don't even think we want to keep anything, that's just in case ;).

leseb:
@alfredodeza thanks for the review.

PR #1144, opened 2018-10-05
ceph-ansible-syntax: remove capital letter check

guits:
Some PRs aren't passing the CI while it should.

It's making us maintaining a list of path and pattern to exclude.
IMHO, this test isn't needed and brings more issues than it is supposed to
solve.


guits:
another example of why we should remove this check [here](https://2.jenkins.ceph.com/job/ceph-ansible-pr-syntax-check/3772/console)

PR #1147, opened 2018-10-09
ceph-ansible: fetch ceph config file on failure

guits:
Let's try to get ceph configuration file on failure, it can be useful at
some point.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -783,6 +783,13 @@ write_collect_logs_playbook() {
         flat: yes
       failed_when: false
       with_items: "{{ ceph_logs.stdout_lines }}"
+
+    - name: collect ceph configuration file
+      fetch:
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -783,6 +788,13 @@ write_collect_logs_playbook() {
         flat: yes
       failed_when: false
       with_items: "{{ ceph_logs.stdout_lines }}"
+
+    - name: collect ceph configuration file
+      fetch:
+        src: "{{ ceph_conf_key_directory }}/{{ cluster }}.conf"
+        dest: "{{ archive_path }}/{{ inventory_hostname }}/"
+      when: stat_ceph_conf.stat.exists
```
andrewschoen: You could use ``fail_on_missing: no`` instead of doing a stat and checking it here.

I think you'll want ``flat: yes`` as well or it's going to try to append the host name to the file.

guits: @andrewschoen thx

PR #1156, opened 2018-10-11
ceph-ansible: fix collect ceph config task

guits:
`ceph_conf_key_directory` doesn't exist in this playbook.
All our testing is done with '/etc/ceph', let's hardcode it here.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -786,10 +786,14 @@ write_collect_logs_playbook() {
 
     - name: collect ceph configuration file
       fetch:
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -786,7 +786,7 @@ write_collect_logs_playbook() {
 
     - name: collect ceph configuration file
       fetch:
-        src: "{{ ceph_conf_key_directory }}/{{ cluster }}.conf"
+        src: "/etc/ceph/{{ cluster }}.conf"
```
guits: there will be the same issue with `{{ cluster }}`

PR #1166, opened 2018-10-17
ceph-volume-scenario: lock tests to smithi nodes

andrewschoen:
We are currently having troubles running test on our OVH nodes. This
locks these tests to smithi nodes for now so we can do some testing for
12.2.9.


djgalloway:
I added 5 mira with tag `mira` so you could set it to `(smithi|mira)` now

PR #1168, opened 2018-10-18
ceph-dashboard-pull-requests: added 'white-list-labels'

callithea:
and changed 'only-trigger-phrase' to be able to automatically
start this Job for all dashboard PRs

https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.github-pull-request

### DIFF
--- a/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
+++ b/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
@@ -30,9 +30,9 @@
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
+          white-list-labels: "dashboard"
           trigger-phrase: 'jenkins test dashboard'
           skip-build-phrase: '^jenkins do not test.*'
-          only-trigger-phrase: true
           github-hooks: true
           permit-all: true
           auto-close-on-fail: false


ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -30,9 +30,10 @@
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
+          white-list-labels: "dashboard"
```
joscollin: Does this mean that if there is no `dashboard` label in the PR, the build won't trigger? Then this would be a narrowing down, if someone missed to label a dashboard PR.

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -30,9 +30,10 @@
...
           trigger-phrase: 'jenkins test dashboard'
           skip-build-phrase: '^jenkins do not test.*'
-          only-trigger-phrase: true
+          only-trigger-phrase: false
```
tchaikov: this is "false" by default, hence can be removed.

callithea: Okay, I can remove it - I left it there because I've seen it is set to ``false`` e.g. for the [ceph-pull-requests](https://github.com/ceph/ceph-build/blob/master/ceph-pull-requests/config/definitions/ceph-pull-requests.yml#L35) Jenkins config as well.

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
callithea: Yes, that would mean the Job won't be triggered for PRs without the ``dashboard`` label.
That's true, but usually we do assign the dashboard label to our PRs and I don't know how or what kind of other 'filter' I could use to only run this Job for dashboard-related PRs.

PR #1171, opened 2018-10-19
ceph-volume-ansible-prs use credentials binding plugin instead of global passwords

alfredodeza:
This might make it a bit more secure. I'm sure it will cause breakage, would love to see some other eyes on it.

### DIFF
--- a/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
+++ b/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
@@ -148,9 +148,11 @@
             - ../../build/build
 
     wrappers:
-      - inject-passwords:
-          global: true
-          mask-password-params: true
+      - credentials-binding:
+          - username-password-separated:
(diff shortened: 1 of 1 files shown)


djgalloway:
@andrewschoen I'm cleaning up ceph-build.git PRs.  Do you know where these credentials are being used?  I don't see them in the `build` job.

alfredodeza:
They are used with this https://github.com/alfredodeza/github-status, which is useful when triggering the job manually. Specifically, you are looking for these lines in PR jobs:

```text
GITHUB_STATUS_STATE="failure" $VENV/github-status create
```

djgalloway:
> They are used with this https://github.com/alfredodeza/github-status, which is useful when triggering the job manually. Specifically, you are looking for these lines in PR jobs:
> 
> ```
> GITHUB_STATUS_STATE="failure" $VENV/github-status create
> ```

(well look who it is)

Does `github-status` just read the credentials from the ENV?

PR #1177, opened 2018-10-25
ceph-ansible-pipeline: run all test against ceph@master

andrewschoen:
I think we should use the nightly jobs to test the stable releases, like luminous, and run the dev tests for pull requests. Often we need features that only exist in master and not in a release anyway.


guits:
We can't test ceph-ansible with ceph@master only, otherwise, all PRs in ceph-ansible will fail.
We need ceph/ceph-ansible#3187 before.

andrewschoen:
> We can't test ceph-ansible with ceph@master only, otherwise, all PRs in ceph-ansible will fail.
> We need [ceph/ceph-ansible#3187](https://github.com/ceph/ceph-ansible/pull/3187) before.

Let's just quit testing ceph-disk scenarios on PRs, then we could test against master. Do we care more about making sure ceph-disk is still working on ceph-ansible PRs or supporting new features in ceph? I'd prefer letting the nightly tests make sure ceph-disk is still working for ceph-ansible stable branches and test the latest ceph features on PRs.

PR #1180, opened 2018-10-29
ceph-ansible-pipeline: do not test ooo

leseb:
Since all the scenario use lvm now, the ooo scenario will always fail
when triggered from the pipeline. The job will remain in stable branches
though.


guits:
let's remove it from master until we can test a newer image, must sync with osp guys for that.

PR #1187, opened 2018-10-31
ceph-volume ansible jobs need python-netaddr installed in the controlling node

alfredodeza:
This should allow ceph-volume tests to work. PR https://github.com/ceph/ceph-ansible/pull/3199 made it so that any/all tests should have `python-netaddr` installed


alfredodeza:
This will probably need to be reverted, ceph-volume jobs need to have this installed via tox when running the tests, so will probably need to install from requirements.txt

PR #1207, opened 2018-11-16
ceph-ansible: add container_podman to the pipeline

leseb:
container_podman must run on every PR as it's the same scenario as
docker but using podman.


leseb:
Hold until https://github.com/ceph/ceph-ansible/pull/3308 gets merged.

PR #1212, opened 2018-11-30
build_utils: Manually clean up vagrant old domain IDs

djgalloway:


scripts/build_utils.sh
```diff
@@ -840,6 +840,12 @@ teardown_vagrant_tests() {
         sudo virsh net-destroy $network || true
         sudo virsh net-undefine $network || true
     done
+
+    # For when machines get stuck in state: preparing
+    # https://github.com/SUSE/pennyworth/wiki/Troubleshooting#missing-domain
+    for dir in $(sudo find $WORKSPACE | grep '.vagrant/machines'); do
+      rm -rf $dir/*
```
alfredodeza: if you add quotes here it will be safer (even though I don't recall seeing spaces in our paths)
```suggestion
      rm -rf "$dir/*"
```

PR #1221, opened 2018-12-19
ceph-*-build: skip unavailable repo when yum-builddep

tchaikov:
as centos-sclo-rh-source leads us to 404 at this moment. and we are not
using the source repo for building ceph. so we can just skip any
unavailable repo.

Fixes: http://tracker.ceph.com/issues/37707

### DIFF
--- a/ceph-build/build/setup_rpm
+++ b/ceph-build/build/setup_rpm
@@ -38,7 +38,7 @@ elif [ "$ARCH" = arm64 ]; then
     $SUDO yum-config-manager --enable centos-sclo-rh-testing
 fi
 sed -e 's/@//g' < ceph.spec.in > $DIR/ceph.spec
-$SUDO yum-builddep -y $DIR/ceph.spec
+$SUDO yum-builddep -y --setopt=*.skip_if_unavailable=true $DIR/ceph.spec
 
 BRANCH=`branch_slash_filter $BRANCH`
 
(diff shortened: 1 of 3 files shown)


tchaikov:
build passed. see https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=huge/16289/console

ktdreyer:
I think it would be a good idea to revert this once it is working again.

ktdreyer:
My concern with this change is that it will paper over other missing repositories that we expect to be enabled, and then the build could quietly change in weird ways we don't expect.

In this particular case of a broken source package repo, maybe we could've used `yum-builddep --disablerepo centos-sclo-rh-source` to target the exact one we want to disable.

The CentOS infrastructure meeting takes place weekly on Monday at 1400 UTC in #centos-devel . I'll bring [this bug](https://bugs.centos.org/view.php?id=15606) up there to get more eyes on it.

badone:
@ktdreyer surprised they haven't fixed that yet. I understand your argument and personally have no objection to backing this change out.

ktdreyer:
Thanks Brad. I've brought it up in the infra meeting today, and I updated [the bug](https://bugs.centos.org/view.php?id=15606) with the relevant part of the transcript. Brian is looking into it this week.

ktdreyer:
The CentOS infra admins resolved this issue, so we can revert this change, https://github.com/ceph/ceph-build/pull/1228

PR #940, opened 2017-12-15
script/build_utils.sh: don't fail if any mirror is not accessible

tchaikov:
Signed-off-by: Kefu Chai <tchaikov@gmail.com>

PR #1226, opened 2019-01-11
nfs-ganesha: minor apt-get update tweak for build_deb

tserlin:
There are occasional problems with failures in "apt-get update" because of the yandex.ru mirror. This just tweaks the command so we don't fail when the mirror is not available.

Similar to: https://github.com/ceph/ceph-build/pull/940/files

Signed-off-by: Thomas Serlin <tserlin@redhat.com>


tserlin:
@alimaredia I just wanted to get this PR in, but let's wait until https://github.com/ceph/ceph-build/pull/1224 is merged.

PR #1228, opened 2019-01-16
Revert "ceph-*-build: skip unavailable repo when yum-builddep"

ktdreyer:
The CentOS infra admins resolved the repository issue at https://bugs.centos.org/view.php?id=15606 .

Generally speaking, we want Yum to bail out early if there is a Yum repo problem, rather than silently skipping over failures and possibly introducing side effects to our builds.

This reverts commit 806f274710ebddb92541d6a1f82746beff530323.


tchaikov:
i pushed the change to jenkins . if everything goes fine, will merge it.

PR #1248, opened 2019-02-05
ceph-ansible-pr-syntax-check: get latest version of pip

guits:
ensure the latest version of pip is installed for this job to avoid the
following error:

```
ValueError: ('Expected version spec in', 'ansible~=2.7,<2.8', 'at', '~=2.7,<2.8')
```

### DIFF
--- a/ceph-ansible-pr-syntax-check/build/build
+++ b/ceph-ansible-pr-syntax-check/build/build
@@ -8,6 +8,7 @@ pkgs=( "ansible-lint" )
 # force virtualenv to be created earlier for this occurence, so we are not stuck because of ansible-lint dependencie,
 # which would enforce latest ansible version to be installed.
 virtualenv $TEMPVENV
+"$VENV"/pip install --upgrade pip
 "$VENV"/pip install -r <(grep ansible "$WORKSPACE"/ceph-ansible/tests/requirements.txt)
 install_python_packages "pkgs[@]"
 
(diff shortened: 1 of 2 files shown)


guits:
not sure whether we should simply ensure the latest version of pip is installed by adding something like this in a more "common" function in build_utils.sh instead of doing this at job configuration level (perhaps other jobs can't use the latest version...)
@djgalloway @alfredodeza @andrewschoen ?

djgalloway:
We pinned 10.0.0 intentionally in https://github.com/ceph/ceph-build/pull/1007 and prevented pip from updating itself in https://github.com/ceph/ceph-build/commit/ff01d2c5831fedeb8b87f3d51c28e357a4ea7374.

I don't remember why we had to pin pip to 10.0.0 though.

We stopped using `install_python_packages_no_binary` in https://github.com/ceph/ceph-build/commit/2092bf9179ff7284b7a83d54cb15026ad4b507e5

I'm adding all that ^ as an FYI.  I'd be okay with pinning pip to a newer version and seeing how things go.

guits:
@djgalloway 
> I'd be okay with pinning pip to a newer version and seeing how things go.
I'm merging this anyway for now because I need it to avoid some failure in ceph-ansible PRs.

meantime, I don't mind pushing a patch to pin pip to a newer version, do you want me to push this patch ?

PR #1269, opened 2019-03-29
nfs-ganesha: switch to using our own fork for now

jtlayton:
We are currently building a version of nfs-ganesha for nautilus that
lacks many needed patches. Some of those aren't suitable for inclusion
into V2.7-stable, but are already merged for V2.8 (which isn't released
yet).

Switch (at least for now) to building from a custom branch in the
ceph/nfs-ganesha fork.


ktdreyer:
When do we revert this Jeff?

PR #1276, opened 2019-04-03
ceph-ansible: Remove bluestore_lvm_osds scenario

dsavineau:
Testing bluestore is handled in the lvm_osds scenario. We don't need
a dedicated one for that.


ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
```diff
@@ -334,10 +331,8 @@
       - lvm_osds
       - lvm_batch
       - purge_lvm_osds
-      - bluestore_lvm_osds
```
guits: I think we still need this one

PR #1278, opened 2019-04-05
nfs-ganesha-stable: revert back to original repo

tserlin:
Now that nfs-ganesha 2.7.3 is released, we can go back to using the original repo instead of the temporary https://github.com/ceph/nfs-ganesha/tree/V2.7-ceph branch set up by Jeff Layton.


ktdreyer:
Can we delete that fork entirely to avoid confusion?

PR #1303, opened 2019-05-23
ceph-ansible: test podman on centos instead rhel

guits:
podman is available on atomic os, from upstream perspective it's better
to test against this image instead a rhel8 which was a beta image outdated.


guits:
related ceph-ansible PR: ceph/ceph-ansible#4031

PR #1304, opened 2019-05-28
ceph-dashboard-pull-requests: Change google-chrome-stable rpm installation

callithea:
There are currently some 404 errors when trying to access the official
Google download link, sometimes it seems to work to install the package
but then again there's an error message in the Jenkins log output saying
"No package google-chrome-stable available".
With this PR I'd like to test if this alternative approach is more reliable
for the time being in order to avoid job failures.

404 error: https://dl.google.com/linux/chrome/rpm/stable/x86_64


tchaikov:
@callithea i will test this change early tomorrow.

callithea:
> @callithea i will test this change early tomorrow.

Thank you, @tchaikov!

tchaikov:
@callithea see #1307

PR #1310, opened 2019-06-03
ceph-pull-requests: Don't run x86 make check on arm64

djgalloway:


djgalloway:
Nevermind.  The arm64 slaves were just labelled incorrectly after rebuilding Jenkins.

PR #1318, opened 2019-06-14
ceph-ansible: improve collect ceph config file and logs

guits:
let's collect these files in one task instead of two tasks.
By the way, it will avoid undesired failures regarding ceph config files
which don't exist (it first tried ceph.conf > test.conf >
mycluster.conf)

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -856,48 +856,29 @@ write_collect_logs_playbook() {
 - hosts: all
   become: yes
   tasks:
-    - name: find ceph logs
-      command: find /var/log/ceph -name "*.log"
-      register: ceph_logs
-      failed_when: false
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -856,30 +856,19 @@ write_collect_logs_playbook() {
 - hosts: all
   become: yes
   tasks:
-    - name: find ceph logs
-      command: find /var/log/ceph -name "*.log"
-      register: ceph_logs
+    - name: find ceph config file and logs
+      command: find /etc/ceph/ /var/log/ceph -name "*.conf" -o -name "*.log"
```
leseb: Use https://docs.ansible.com/ansible/latest/modules/find_module.html? :-)

PR #1319, opened 2019-06-17
 ceph-dev-new-setup: missing pieces in PR#1315

tchaikov:
these commits were dropped by accident while doing rebase


xiexingguo:
@tchaikov Ah, thank you for the quick fix!

PR #1321, opened 2019-06-19
experimental job for the container-building

dmick:

### DIFF
--- a/dmick-container-build/build/build
+++ b/dmick-container-build/build/build
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -e
+
+
+cd "$WORKSPACE"/ceph-container/ || exit
+bash -x contrib/build-push-ceph-container-imgs.sh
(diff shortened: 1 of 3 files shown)


leseb:
Can we also hook it up on the job that builds packages so that once we have the package we can build the container associated?

djgalloway:
@dmick we don't need to merge this to get the job into Jenkins.  We can push this manually using jjb from a workstation and tweak from there.  Let's work on this together this week.

@leseb How long does it usually take to build a container?  I'm wondering if we should do this concurrently instead of consecutively.  The downside of concurrent is we'll be waiting on system resources (slaves).  The downside of consecutive is I know devs are already waiting 2+ hours for branches they push to get built.  If we build containers *after* packages, they'll be waiting even longer.

leseb:
@djgalloway the problem is that in order to build the container you need the repo and the packages... So the container build must happen after repo/packages are available.

djgalloway:
Oh.  Duh.  :stuck_out_tongue_closed_eyes:

dmick:
Yes can definitely hook up.  This is very early experimenting with steps

PR #1328, opened 2019-06-28
scripts/build_utils.sh: only use GCC-9 for building crimson

tchaikov:
using GCC-9 for building mimic and other stable releases has following
consequences:

* it brings runtime dependencies of libstdc++ runtime shipped along with
  GCC-9
* GCC-9 is more picky, so some 3rd party libraries fail to build due to
  -Werror.

let's continue using GCC-8 on bionic when building non-crimson flavors


yuriw:
@tchaikov @badone we are green again, good job, guys!

PR #1329, opened 2019-07-02
Add ceph container nightly devel build job

dsavineau:
The ceph container images are only built on github event and also based
on stable ceph content.
We can also have nightly builds with ceph developement content from
shaman/chacra.

See https://github.com/ceph/ceph-container/pull/1411

### DIFF
--- a/ceph-container-build-push-imgs-devel-nightly/config/definitions/ceph-container-build-push-imgs-devel-nightly.yml
+++ b/ceph-container-build-push-imgs-devel-nightly/config/definitions/ceph-container-build-push-imgs-devel-nightly.yml
@@ -1,0 +1,49 @@
+- job:
+    name: ceph-container-build-push-imgs-devel-nightly
+    node: huge && trusty && x86_64
+    project-type: freestyle
+    defaults: global
+    display-name: 'ceph-container: Nightly build and push devel container images to Docker Hub'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: -1
+          num-to-keep: -1
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: https://github.com/ceph/ceph-container
+
+    triggers:
+      - timed: '@daily'
+
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph-container.git
+          branches:
+            - master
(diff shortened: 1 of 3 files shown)


ceph-container-build-push-imgs-devel-nightly/config/definitions/ceph-container-build-push-imgs-devel-nightly.yml
```diff
@@ -0,0 +1,46 @@
...
+            - ../../../scripts/build_utils.sh
+            - ../../build/build
+
+    wrappers:
+      - inject-passwords:
+          global: true
+          mask-password-params: true
+      - credentials-binding:
+          - username-password-separated:
+              credential-id: docker-hub-leseb
```
ktdreyer: Can we stop the proliferation of Sebastien's personal account?

dsavineau: In fact, since the jenkins reset, it's my account used behind the `docker-hub-leseb` id to avoid changing all jobs.
If you prefer to rename every `docker-hub-leseb` occurence by mine, I'm ok with it.
Or maybe you have a better solution for that ?

dsavineau:
> When you say the ceph containers are built on GitHub events, what does this mean? I'm asking in the context of [ceph/ceph-container#1405](https://github.com/ceph/ceph-container/issues/1405) . In general I want to be [event-driven](https://blog.ktdreyer.com/2019/01/nightly-builds-are-too-fast-and-too-slow.html) instead of clock-driven.

Well that's a little bit tricky.
Currently we only rebuild upstream containers when a PR is merged. So if nothing is merged during 1 month, we don't have any updates (ceph or security) for container images.
One of the CSI team needs was to be able to have more updated container images and based on unreleased content (not stable). That's why we need a daily trigger instead of github event.

ceph-container-build-push-imgs-devel-nightly/config/definitions/ceph-container-build-push-imgs-devel-nightly.yml
ktdreyer: Can we have a bot account for Docker Hub?

ktdreyer:
> Currently we only rebuild upstream containers when a PR is merged.
I see, and specifically you mean *when a ceph-container PR is merged*, not when a general PR of ceph itself is merged.

PR #1330, opened 2019-07-04
ceph-*build/build/setup_rpm: downgrade to gcc-c++-8.2.1

tchaikov:
Fixes: http://tracker.ceph.com/issues/40646

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -953,3 +953,12 @@ get_nr_build_jobs() {
     fi
     echo $n_build_jobs
 }
+
+maybe_downgrade_gcc() {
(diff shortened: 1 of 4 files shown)


scripts/build_utils.sh
```diff
@@ -953,3 +953,11 @@ get_nr_build_jobs() {
     fi
     echo $n_build_jobs
 }
+
+maybe_downgrade_gcc() {
```
ktdreyer: Would you mind adding a comment to https://bugzilla.redhat.com/show_bug.cgi?id=1726630? Without seeing the Git history here, it's quite hard to understand why we would ever do this.

tchaikov: sure. updated and repushed.

PR #1334, opened 2019-07-10
ceph-dev,ceph-dev-trigger: add "crimson" flavor to master, octopus branch builds

tchaikov:


ceph-dev-setup/build/build
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash -ex
+# -*- mode:sh; tab-width:4; sh-basic-offset:4; indent-tabs-mode:nil -*-
+# vim: softtabstop=4 shiftwidth=4 expandtab
```
alfredodeza: I love Vim, but do we really need to introduce editor-specific configuration in files? This can all be configured locally. Not everyone uses Vim either

tchaikov: i agree that not everyone uses VIM. and the previous line is for Emacs users =)

and yeah, i will remove them. the reason i added the editor variables was that i want to avoid mix of tab and spaces in future -- like what we have in `vstart.sh` now.

tchaikov: @alfredodeza updated and repushed

ceph-dev-setup/build/build
```diff
@@ -35,8 +35,9 @@ rm -rf release
 # because autogen+configure will check for dependencies, we are forced to install them
 # and ensure they are present in the current host
 if [ -x install-deps.sh ]; then
-  echo "Ensuring dependencies are installed"
-  ./install-deps.sh
+    echo "Ensuring dependencies are installed"
+    [ "${FLAVOR}" == "crimson" ] && WITH_SEASTAR=1
+    WITH_SEASTAR=$WITH_SEASTAR ./install-deps.sh
```
alfredodeza: is it fine if `$WITH_SEASTAR` is undefined?

ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
```diff
@@ -80,6 +80,34 @@
...
+                    FORCE=True
+                    DISTROS=bionic xenial centos7
+                - project: 'ceph-dev'
+                  predefined-parameters: |
+                    BRANCH=${GIT_BRANCH}
+                    FORCE=True
+                    DISTROS=centos7
+                    FLAVOR=notcmalloc
+                - project: 'ceph-dev'
+                  predefined-parameters: |
```
alfredodeza: this is building crimson always in addition to the other 4 builds right? (for a total of 5)

ceph-dev-setup/build/build
tchaikov: it's fine.

in `install-deps.sh`, we have:
```sh
[ $WITH_SEASTAR ] && with_seastar=true || with_seastar=false
```

ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
tchaikov: yeah, see https://shaman.ceph.com/builds/ceph/wip-kefu-testing-2019-07-12-1253/59756e32c61c433f7d62ae4dd1c40085f7c8f630/

djgalloway:
@tchaikov As of today, do we have crimson builds for all the correct jobs and distros?  Can this PR be closed?

tchaikov:
@djgalloway we still don't have teuthology-based test for crimson. and i expect that in future, we will need to build crimson flavor on master in addition to the `wip-*` branches. but before that, there is no need to have it. so i am closing this PR.

actually, i didn't expect that we build crimson on **both** ubuntu focal and centos 8. as we will likely deploy crimson solely using cephadm for teuthology-based test, so centos 8 would suffice by then.

before we have teuthology-based tests for crimson, i think there is no need to build crimson on multiple distros if our purpose is not to make sure crimson does build on these two distros.

the reason i enabled ubuntu build for crimson was to ensure that crimson builds with the PRs under test.

@ideepika what do you think?

ideepika:
@tchaikov  I assumed earlier we were using it for crimson perf testing, since that is not the case... sounds good, could save us some resources. :+1:

PR #1343, opened 2019-07-12
ceph-volume: allow passing GITHUB_TOKEN as a password to manually set github status

alfredodeza:
Before, we stored the GITHUB_TOKEN in Jenkins which was super duper bad, now we only pass it in when we need it, and it remains obscured in logs and parameters. 

This is useful when requesting the tests manually, so Jenkins has a way to set the status without being triggered from Github.

### DIFF
--- a/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
+++ b/ceph-volume-ansible-prs/config/definitions/ceph-volume-pr.yml
@@ -102,6 +102,9 @@
       - string:
           name: GITHUB_SHA
           description: "The tip (last commit) in the PR, a sha1 like 7d787849556788961155534039886aedfcdb2a88 (if set, will report status to Github)"
+      - password:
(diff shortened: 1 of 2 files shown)


ktdreyer:
Are there any docs about this?

alfredodeza:
The ceph-volume jobs use https://github.com/alfredodeza/github-status which in turn contacts Github to set the status. That tool is documented although not thoroughly, it explains the environment variables needed to contact Github to set a desired status. 

The ceph-volume jobs use this function https://github.com/ceph/ceph-build/blob/master/scripts/build_utils.sh#L809 to ensure proper vars are up and then it calls the tool:

```
GITHUB_STATUS_STATE="success" $VENV/github-status create
```

Are you looking forward to more documentation on usage in a Job in the ceph-build repository? Or more documentation in the github-status tool? (maybe both?)

PR #1344, opened 2019-07-12
Debian buster support for releases

ThoreKr:
Are there additional changes required in order to bring ceph builds to buster?


ldidry:
Up? Debian Buster has been released on july 2019, it’s been more than 6 months and it would be really appreciable to be able to upgrade to Buster and then upgrade Ceph.

winacid:
What is the problem here?

PR #1346, opened 2019-07-23
nfs-ganesha: Disable LizardFS FSAL

dsavineau:
LizardFS FSAL has been added in [1] and it's enabled by default. This
is currently failing in the CI.

CMake Error at CMakeLists.txt:595 (message):
  STRICT_PACKAGE: Cannot find LizardFS client lib.  Disabling lizardfs fsal

This patch configures the USE_FSAL_LIZARDFS option to OFF.

[1] https://github.com/nfs-ganesha/nfs-ganesha/commit/884ec97

### DIFF
--- a/nfs-ganesha/build/build_rpm
+++ b/nfs-ganesha/build/build_rpm
@@ -94,7 +94,7 @@ mkdir build
 cd build
 
 # generate .spec file, edit .spec file for correct versions of libs and make source tarball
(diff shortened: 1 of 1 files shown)


guits:
@dang @jtlayton do you mind ? :slightly_smiling_face:

jtlayton:
Would be nice if the nfs-ganesha cmake could just auto-disable it when the library isn't present, but this is a suitable workaround for now.

guits:
allowing myself to merge since it was approved (+ we are facing some issues because of this)

PR #1355, opened 2019-08-01
ceph-dev-build: fix deb and rpm validate

kshtsk:
validate_deb - not redhat does not mean it is deb distro,
there can be other packaging systems
validate_rpm - we do not want to proceed for zypper based,
we only want to proceed for redhat like repo

### DIFF
--- a/ceph-dev-build/build/validate_deb
+++ b/ceph-dev-build/build/validate_deb
@@ -2,6 +2,12 @@
 set -ex
 
 # Only do actual work when we are a DEB distro
-if test -f /etc/redhat-release ; then
-    exit 0
-fi
+( source /etc/os-release
+  case $ID in
+    debian|ubuntu)
+      exit 0
+      ;;
+    *)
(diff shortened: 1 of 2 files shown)


alfredodeza:
The only packages we build are for CentOS and Ubuntu, we don't build for zypper, devuan, virtuozzo or anything like that.

kshtsk:
> The only packages we build are for CentOS and Ubuntu, we don't build for zypper, devuan, virtuozzo or anything like that.

@alfredodeza This is what I'm working on, this patch is required for adding openSUSE.

kshtsk:
@alfredodeza do you suggest to remove devuan and virtuozzo from the case in validate scripts?

alfredodeza:
@kshtsk these scripts are for releasing ceph and creating development builds that are hosted in our systems and we don't have any way (or plan) to support openSUSE or any of the other distros this PR is adding.

PR #1365, opened 2019-08-14
osc build scripts

kshtsk:


kshtsk:
should be merged after #1361

PR #1367, opened 2019-08-18
ceph*build/build/build_deb: upload *.ddeb packages

tchaikov:
ubuntu names `*dbgsym*` packages .ddeb in hope to put them into a
separated repo. while debian just use the plain .deb extension for them.
so, to include the dbgsym packages created by ubuntu, we need to upload
the .ddeb packages as well.

### DIFF
--- a/ceph-build/build/build_deb
+++ b/ceph-build/build/build_deb
@@ -95,7 +95,7 @@ echo lintian --allow-root $releasedir/$cephver/*$bpvers*.deb
 
 if [ "$THROWAWAY" = false ] ; then
     # push binaries to chacra
-    find release/$vers/ | egrep "*\.(changes|deb|dsc|gz)$" | egrep -v "(Packages|Sources|Contents)" | $VENV/chacractl binary ${chacra_flags} create ${chacra_endpoint}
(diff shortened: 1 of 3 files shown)


tchaikov:
i think we will need https://github.com/ionos-enterprise/reprepro for indexing `*.ddeb` packages. this repo also has mutli-version patches applied.

alfredodeza:
The change of tooling will require some effort that I am not sure how to best go about it. In the past we did researched several tools for creating the repos but we ended up settling with what was used before as we wanted to deliver the new build system as soon as possible.

tchaikov:
@alfredodeza if we are fine with trying out and then using the reprepro with ddeb and multiple-version patches. i think we could

1. merge https://github.com/ceph/chacra/pull/267 and deploy the updated server side of chacra if necessary
2. build reprepro package for the distro where existing reprepro instance is deployed.
3. backup the .db and .version files under `--dbdir` first.
4. install the new reprepro package.
5. push https://github.com/ceph/ceph/pull/29720 to ceph-ci again.
6. push a random branch without https://github.com/ceph/ceph/pull/29720 to ceph-ci
7. check the repo, if the repo does not contain ddeb files or even worse.. add an `options` file under reprepro's confdir in `reprepro_confdir()` to add multiple `verbose` options in it. and repeat step 5, 6 and 7
8. as the last resort fallback to the vanilla version with the original copy of `dbdir`.

i can help on any/all of them if we are good with the plan in my day time as it might have less impact as presumably, we have less traffic in APAC timezones.

PR #1374, opened 2019-08-22
ceph-iscsi-stable: Check for specific package files

djgalloway:
The -f `-*` got expanded too many times if there are multiple matches which broke the `if` statement.

### DIFF
--- a/ceph-iscsi-stable/build/build_rpm
+++ b/ceph-iscsi-stable/build/build_rpm
@@ -65,13 +65,13 @@ chacra_endpoint="ceph-iscsi/${REPO_MAJOR_VERSION}/${GIT_COMMIT}/${DISTRO}/${RELE
 chacra_repo_endpoint="${chacra_endpoint}/flavors/default"
 
 # check to make sure ceph-iscsi-config package built
-if [ ! -f $WORKSPACE/dist/RPMS/ceph-iscsi-*.rpm ]; then
+if [ ! -f $WORKSPACE/dist/RPMS/ceph-iscsi-${CEPH_ISCSI_BRANCH}-1.el${RELEASE}.noarch.rpm ]; then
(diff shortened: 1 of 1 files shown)


ceph-iscsi-stable/build/build_rpm
```diff
@@ -65,13 +65,13 @@ chacra_endpoint="ceph-iscsi/${REPO_MAJOR_VERSION}/${GIT_COMMIT}/${DISTRO}/${RELE
 chacra_repo_endpoint="${chacra_endpoint}/flavors/default"
 
 # check to make sure ceph-iscsi-config package built
-if [ ! -f $WORKSPACE/dist/RPMS/ceph-iscsi-*.rpm ]; then
+if [ ! -f $WORKSPACE/dist/RPMS/ceph-iscsi-${CEPH_ISCSI_BRANCH}-${CEPH_ISCSI_BRANCH}.el${RELEASE}.noarch.rpm ]; then
```
dillaman: There shouldn't be a duplicated `CEPH_ISCSI_BRANCH` in the release. The 3.2 release has RPMs like `ceph-iscsi-3.2-1.el7.noarch.rpm`

djgalloway: Oops, you're right.  Where does the `-1` come from?

https://1.chacra.ceph.com/r/ceph-iscsi/3/ec80bb1cc8b288b40bc35e03e4ba4c60daced938/centos/7/flavors/default/noarch/

dillaman: It comes from here [1] since that's the RH standard way of denoting the first build after a release.

[1] https://github.com/ceph/ceph-build/blob/master/ceph-iscsi-stable/build/build_rpm#L24

PR #1375, opened 2019-08-22
ceph-iscsi-stable: revert to production chacra instance

dillaman:


alfredodeza:
Not sure why the change, but this looks fine to me

djgalloway:
> Not sure why the change, but this looks fine to me

I think @dillaman was debugging why `chacractl` was failing in the past so changed the chacra instance (The root cause was missing chacra credentials).  Since this is for stable/permanent releases, the packages need to be shipped to the permanent chacra instance.

PR #1377, opened 2019-08-23
ceph-volume: cleanup workspace before running tests

alfredodeza:
The repos were never getting cleaned up causing `git fetch` to fail. This is already tested and making ceph-volume tests pass again.

/cc @jan--f


djgalloway:
Why `workspace-cleanup` vs `wipe-workspace: true` under the `git`?

alfredodeza:
@djgalloway great catch, I was able to tell that the Gti SCM section had it turned off, so I made the change to enable it and after testing, this would work fine.

As a side note, someone may one to use the `workspace-cleanup` because it has granular controls for deletion, like patters on what to delete exactly, and external command to use for deletion.

Issue #1390, opened 2019-09-06
ceph-ansible-pr-syntax-check does not install all requirements


hjensas:
When installing requirements only ansible is installed because a grep is used on the requirements file.

`"$VENV"/pip install -r <(grep ansible "$WORKSPACE"/ceph-ansible/tests/requirements.txt)`

This causes failure in pull request https://github.com/ceph/ceph-ansible/pull/4339 which introduces a plug-in that require netaddr.

PR #1391, opened 2019-09-06
pr-syntax-check - install all requirements

hjensas:
This change removes the grep for 'ansible' when isntalling
requirements in ceph-ansible-pr-syntax-check. The grep
cause onlye the ansible requirement to be installed, ignoring
the other packages. This causes an issue in PR[1] which
introduce a filter plugin that depend on the netaddr package.

[1] https://github.com/ceph/ceph-ansible/pull/4339

Closes: #1390

Signed-off-by: Harald Jensås <hjensas@redhat.com>

### DIFF
--- a/ceph-ansible-pr-syntax-check/build/build
+++ b/ceph-ansible-pr-syntax-check/build/build
@@ -8,7 +8,7 @@ set -e
 # which would enforce latest ansible version to be installed.
 virtualenv $TEMPVENV
 "$VENV"/pip install --upgrade pip
-"$VENV"/pip install -r <(grep ansible "$WORKSPACE"/ceph-ansible/tests/requirements.txt)
+"$VENV"/pip install -r "$WORKSPACE"/ceph-ansible/requirements.txt
 "$VENV"/pip install ansible-lint
 
 


ceph-ansible-pr-syntax-check/build/build
```diff
@@ -8,7 +8,7 @@ set -e
 # which would enforce latest ansible version to be installed.
 virtualenv $TEMPVENV
 "$VENV"/pip install --upgrade pip
-"$VENV"/pip install -r <(grep ansible "$WORKSPACE"/ceph-ansible/tests/requirements.txt)
+"$VENV"/pip install -r "$WORKSPACE"/ceph-ansible/tests/requirements.txt
```
dsavineau: In fact, I don't really know why we are installing the tests/requirements.txt there when we can only install the basic requirements (which includes netaddr) [1]

Could you do that change ?

[1] https://github.com/ceph/ceph-ansible/blob/master/requirements.txt

hjensas: Done.

PR #1392, opened 2019-09-07
*/scm: only pull the PR under test

tchaikov:
see
https://github.com/jenkinsci/ghprb-plugin/blob/master/README.md#creating-a-job

no need to pull all PR references of the remote repo. it'd be faster
just pull the references of a single PR. actually, what we need is but

- refs/pull/${pr}/head
- refs/pull/30231/merge

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #1393, opened 2019-09-10
ceph-perf-pull-requests: add ceph-perf-pull-requests

tchaikov:


ceph-ansible-pr-syntax-check/config/definitions/ceph-ansible-pr-syntax-check.yml
```diff
@@ -5,7 +5,7 @@
           url: https://github.com/ceph/ceph-ansible.git
           branches:
             - ${sha1}
-          refspec: +refs/pull/*:refs/remotes/origin/pr/*
+          refspec: +refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*
```
alfredodeza: looks like this PR has feature creep. Are you trying to introduce these enhancements aside from adding the Ceph Performance tests? or are they related?

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -0,0 +1,130 @@
...
+    builders:
+      - shell: |
+          ${{WORKSPACE}}/cbt/compare.py -a ${{WORKSPACE}}/cbt-results/{pr} -b ${{WORKSPACE}}/cbt-results/{master} -v
+
+- job:
+    name: ceph-perf-pull-requests
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: huge && (centos7 || trusty) && rebootable
```
alfredodeza: performance tests need specific machines, these machines require a label so that these tests can identify them. I recently added one manually. Some coordination needs to happen so that this works with labels that are expected.

ceph-ansible-pr-syntax-check/config/definitions/ceph-ansible-pr-syntax-check.yml
tchaikov: @alfredodeza thanks for reviewing this PR. a subset of this changeset was posted as #1392.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
tchaikov: @alfredodeza yeah, i am still struggling with the glue code to drive cbt. and later will try to use github API to send a table to summarize the test result instead of just a single string.

are you talking about incerta08? i was trying to build ceph on RHEL8, but seems we are not quite there yet.

PR #1397, opened 2019-09-18
clean up shaman repos left behind after a build

alfredodeza:
Fixes: https://tracker.ceph.com/issues/41877


alfredodeza:
@djgalloway

PR #1402, opened 2019-10-02
kernel: Report build statuses to shaman

djgalloway:


djgalloway:
OK this is ready now.  I was having trouble getting `DISTRO` and `DISTRO_VERSION` set but I got it.

PR #1413, opened 2019-10-15
ceph-ansible-prs: re-add shrink/rgw_multisite jobs

dsavineau:
Since the pipeline refactoring we're not able to trigger manually the
shrink and rgw_multisite jobs. The shrink jobs were only configured for
ubuntu and rgw_multisite was missing.

### DIFF
--- a/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
+++ b/ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
@@ -41,6 +41,26 @@
         - 'ceph-ansible-prs-auto'
 
 - project:
+    name: ceph-ansible-prs-common-trigger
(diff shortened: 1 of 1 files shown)


ceph-ansible-prs/config/definitions/ceph-ansible-prs.yml
```diff
@@ -40,6 +40,26 @@
     jobs:
         - 'ceph-ansible-prs-auto'
 
+- project:
+    name: ceph-ansible-prs-common-trigger
+    slave_labels: 'vagrant && libvirt && (smithi || centos7)'
```
guits: in fact, I believe we could change this to `vagrant && libvirt && centos7` but probably a dedicated commit/pr for that would make sense.

PR #1415, opened 2019-10-17
ansible: Update playbooks for CentOS8 compatibility

djgalloway:

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -126,7 +126,7 @@
       when: ansible_os_family == "RedHat"
 
     - name: Install RPM requirements
-      yum:
+      package:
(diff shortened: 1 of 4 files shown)


ansible/examples/slave.yml
```diff
@@ -126,7 +126,7 @@
       when: ansible_os_family == "RedHat"
 
     - name: Install RPM requirements
-      yum:
+      package:
```
alfredodeza: is this change because the python-yum library doesn't exist? Or is it because it is now `dnf` and not `yum` ?

djgalloway: It exists but is not installed and the default is `dnf`.  So it's package-manager agnostic.

PR #1420, opened 2019-10-23
Add jenkins-build user to mock group during slave setup

djgalloway:
Instead of during each individual job


alfredodeza:
This is great. Good work here. I usually prefer more automation so that scripts are reduced. Reducing scripts is always a win!

djgalloway:
Yes, I thought I got them all though?

PR #1424, opened 2019-10-24
ceph-pr-render-docs: Clean up old doc renders

djgalloway:
Since these don't get stored on the larger volume attached to docs.ceph.com, they can fill up the root disk without being cleaned up.

### DIFF
--- a/ceph-pr-render-docs/build/build
+++ b/ceph-pr-render-docs/build/build
@@ -15,7 +15,6 @@ if [ -z "$ghprbPullId" ]; then
     PR_ID="manual"
 fi
 
-
 ./admin/build-doc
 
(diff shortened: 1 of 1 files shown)


ceph-pr-render-docs/build/build
```diff
@@ -15,6 +15,8 @@ if [ -z "$ghprbPullId" ]; then
     PR_ID="manual"
 fi
 
+# Cleanup docs rendered 90+ days ago
+find /var/ceph-prs/ -mindepth 1 -maxdepth 1 -mtime +90 -exec rm -rvf {} \;
```
alfredodeza: if for whatever reason this fails, the `set -ex` at the beginning of this file will cause the job to fail before building the docs. Can this unset the `e` flag? (I think this is done with `+e`) or placed at the very end?

PR #1426, opened 2019-10-29
ceph-*-pull-requests*: use ghprbPullId to define branches

tchaikov:
* add "ghprbPullId" param, simpler to use than "sha1"
* instead of using "sha1", use "ghprbPullId" to define branches to build,
  more consistent this way. as we pull the refs using
  `+refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*`
  and after the change, the ref to build is specified using
  `origin/pr/${ghprbPullId}/merge`. also, by removing "sha1",
  we can specify a single parameter when rebuilding/building a branch
  on jenkins instead of two parameters.


tchaikov:
tested at https://jenkins.ceph.com/job/ceph-perf-pull-requests/57

PR #1431, opened 2019-11-06
ceph-volume-pr: Rename job to ceph-volume-pull-requests

toabctl:
That's the naming schema that is documented in README.rst and used for
other jobs.
So to be consistent and follow our own rules, rename the job.


toabctl:
no response. closing

PR #1432, opened 2019-11-06
scripts: add bionic-updates to pbuilderrc for gcc 7.4.0

alfredodeza:
This should ensure that gcc 7.4.0 gets picked up, avoiding the problem of using 7.3.0 which is causing segfaults

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -482,6 +482,12 @@ setup_pbuilder() {
     echo "$components" > ~/.pbuilderrc
     echo "$debootstrapopts" >> ~/.pbuilderrc
 
+    # gcc 7.4.0 will come from bionic-updates, those need to be added as well
+    if [ $DIST = "bionic" ]; then
+        other_mirror='OTHERMIRROR="deb $mirror $DIST-updates $components"'
+        echo "$other_mirror" >> ~/.pbuilderrc
+    fi
+
     if [ -n "$use_gcc" ]; then
         # Newer pbuilder versions set $HOME to /nonexistent which breaks all kinds of
         # things that rely on a proper (writable) path. Setting this to the system user's $HOME is not enough


alfredodeza:
If this looks OK I am happy to push it out to try it live, so that we can rollback easily vs. just merging as-is

djgalloway:
It seems like maybe we're writing these configurations to an rc file but also passing flags on the CLI: https://github.com/ceph/ceph-build/blob/56a9d159d471884273f4125417867cc20cda8c11/scripts/build_utils.sh#L511

I think we also need to pass `--othermirror "$othermirror"` and figure out why we're doing both in a different PR.

alfredodeza:
The documentation for pbuilder states that it is "either or":

> Using the 'othermirror' option
>
> It is possible to specify more mirrors than the one specified for MIRRORSITE. There's two ways to do this, specifiying the option '--othermirror "<sources.list deb line>"' when running pbuilder, or using the variable 'OTHERMIRROR="<sources.list deb line>"' 

https://wiki.ubuntu.com/PbuilderHowto

djgalloway:
> The documentation for pbuilder states that it is "either or":
> 
> > Using the 'othermirror' option
> > It is possible to specify more mirrors than the one specified for MIRRORSITE. There's two ways to do this, specifiying the option '--othermirror "<sources.list deb line>"' when running pbuilder, or using the variable 'OTHERMIRROR="<sources.list deb line>"'
> 
> https://wiki.ubuntu.com/PbuilderHowto

Sure, I guess my question is which takes precedent?  I'm guessing if an option is passed on the CLI, the equivalent option in `.pbuilderrc` is ignored.  Is that not the case if there's an option in the RC file that's **not** passed on the CLI?  What's the harm in doing both and we can standardize later?

alfredodeza:
@djgalloway I believe that the current state of this function where things are specified in both places has happened because options where added quickly, without having a good understanding of the process. 

I can follow up with a cleanup after we can verify this works. If it does not work, then we can revise to use the CLI. I could not find any pbuilder doc that would mention what priority would be used, but since we are not specifying contradicting configuration, I am inclined to use this as-is

djgalloway:
I don't understand the resistance to be certain `OTHERMIRROR` gets passed to pbuilder.

Should also probably reference https://tracker.ceph.com/issues/42596 in the commit message.

Other than that, I appreciate you figuring this out.  I was in over my head.

alfredodeza:
I've added the ticket in the commit message (wasn't aware one existed)

djgalloway:
```
dgalloway@smithi069:~$ cat .pbuilderrc 
COMPONENTS="main universe"
MIRROR="deb mirror.from.rc"

dgalloway@smithi069:~$ sudo pbuilder dumpconfig --mirror "deb mirror.from.cli" --othermirror "deb othermirror.from.cli" | grep -i "MIRROR.*=.*"
W: No local /etc/mailname to copy, relying on /var/cache/pbuilder/build/20367/etc/mailname to be correct
W: --override-config is not set; not updating apt.conf Read the manpage for details.
BASH_ARGV=([0]="deb othermirror.from.cli" [1]="--othermirror" [2]="deb mirror.from.cli" [3]="--mirror" [4]="dumpconfig")
MIRROR='deb mirror.from.rc'
MIRRORSITE='deb mirror.from.cli'
OTHERMIRROR='deb othermirror.from.cli'

dgalloway@smithi069:~$ echo 'OTHERMIRROR="deb othermirror.from.rc"' >> .pbuilderrc

dgalloway@smithi069:~$ sudo pbuilder dumpconfig --mirror "deb mirror.from.cli" --othermirror "deb othermirror.from.cli" | grep -i "MIRROR.*=.*"
W: No local /etc/mailname to copy, relying on /var/cache/pbuilder/build/20435/etc/mailname to be correct
W: --override-config is not set; not updating apt.conf Read the manpage for details.
BASH_ARGV=([0]="deb othermirror.from.cli" [1]="--othermirror" [2]="deb mirror.from.cli" [3]="--mirror" [4]="dumpconfig")
MIRROR='deb mirror.from.rc'
MIRRORSITE='deb mirror.from.cli'
OTHERMIRROR='deb othermirror.from.cli'

dgalloway@smithi069:~$ sudo pbuilder dumpconfig --mirror "deb mirror.from.cli" | grep -i "MIRROR.*=.*"
W: No local /etc/mailname to copy, relying on /var/cache/pbuilder/build/20503/etc/mailname to be correct
W: --override-config is not set; not updating apt.conf Read the manpage for details.
BASH_ARGV=([0]="deb mirror.from.cli" [1]="--mirror" [2]="dumpconfig")
MIRROR='deb mirror.from.rc'
MIRRORSITE='deb mirror.from.cli'
OTHERMIRROR='deb othermirror.from.rc'
```

Um.  So according to this, with `MIRROR`, even if it's passed on the CLI, the RC file takes precedent.  But with `OTHERMIRROR`, if it's defined in both places, CLI takes precedent.  If it's ONLY defined in the RC file and not passed on CLI, the RC option is still included (which was the only scenario I wondered about).

ktdreyer:
We probably need to be using the "-updates" repo for all distros, not just bionic, right?

PR #1434, opened 2019-11-11
sepia-fog-images: Wait longer for lab to quiet down

djgalloway:
30min wasn't enough for jobs to stop scheduling Deploy tasks.

### DIFF
--- a/sepia-fog-images/build/build
+++ b/sepia-fog-images/build/build
@@ -189,8 +189,8 @@ if [ "$PAUSEQUEUE" == "true" ]; then
   # Capturing a new OS image can interrupt active OS deployments.
   if [ $deploytasks -gt 0 ]; then
     for type in $MACHINETYPES; do
-      # Only pause the queue for 20 minutes just in case anything goes wrong with the Jenkins job.
-      teuthology-queue --pause 1200 --machine_type $type
+      # Only pause the queue for 1hr just in case anything goes wrong with the Jenkins job.
+      teuthology-queue --pause 3600 --machine_type $type
     done
     pausedqueue=true
     currentretries=0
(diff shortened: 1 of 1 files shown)


sepia-fog-images/build/build
```diff
@@ -189,8 +189,8 @@ if [ "$PAUSEQUEUE" == "true" ]; then
   # Capturing a new OS image can interrupt active OS deployments.
   if [ $deploytasks -gt 0 ]; then
     for type in $MACHINETYPES; do
-      # Only pause the queue for 20 minutes just in case anything goes wrong with the Jenkins job.
-      teuthology-queue --pause 1200 --machine_type $type
+      # Only pause the queue for 1hr just in case anything goes wrong with the Jenkins job.
+      teuthology-queue --pause 3600 --machine_type $type
```
alfredodeza: These look like magical numbers (picking one that looks good enough), is there a way to programmatically asses how long this needs to pause?

djgalloway: teuthology jobs can take anywhere from like 20min to 12 hours.  So not really.

alfredodeza: So there is no way to ask teuthology if there is a job that is still doing work? I understand that the time can be variable, all the more interest to properly calculate this vs. updating the number

djgalloway: teuthology jobs can still be running.  That's fine.  The goal here is:

1. Pause the queue
2. Allow any active FOG deploy tasks to complete
3. Allow any teuthology jobs in Waiting or Running status to queue FOG deploy tasks and let them complete
4. Disallow any new teuthology jobs from starting so that no more FOG deploy tasks can be scheduled until the job is done

There is no way to query running/waiting jobs to see if they've done the FOG deploy stuff yet.

PR #1437, opened 2019-11-14
quay-pruner/build/prune-quay.py: handle new "short tags"

dmick:
There are two new tags proposed for each image: "branchname" and "full
sha1".  Branchname is currently being created.  When pruning image tags,
we need to prune the associated short tags as well.

This also fixes a current bug where, because branchname tags are being
created, the current unaware prune script deletes them because it
treats "can't parse tagname" as equivalent to "can't find in shaman".


dmick:
@liewegas without this the next run will delete all the branchname and sha1 tags

PR #1444, opened 2019-12-02
ceph-ansible-docs: build doc for stable-4.0

dsavineau:
Let's build the upstream documentation for stable-4.0

Closes: https://github.com/ceph/ceph-ansible/issues/4794


alfredodeza:
Is it worth to do this even if the docs aren't updated? The last change in the docs directory was a fix for spelling and that happened in July. Before that it was just a bump on versions supported.

PR #1451, opened 2019-12-12
ceph-dev-build: Use correct OBS repo octopus on openSUSE Leap

toabctl:
We want to build against openSUSE Leap 15.2 for octopus. So use the
correct repo.

Signed-off-by: Thomas Bechtold <tbechtold@suse.com>


kshtsk:
@djgalloway can we merge it?

PR #1469, opened 2019-12-18
ceph-pull-requests: Do NOT run on pbuilder hosts

djgalloway:
pbuilder hosts are running Xenial.  We still want to be able to build Bionic packages on Xenial but we do not want make check to run on *all* slaves with the "bionic" label.

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -3,7 +3,8 @@
     project-type: freestyle
     defaults: global
     concurrent: true
-    node: huge && bionic && rebootable && x86_64
+    # We want make check to only run on Bionic b/c it has python2 and python3 and b/c all builds should build on Bionic
+    node: huge && bionic && rebootable && x86_64 && !xenial && !trusty
     display-name: 'ceph: Pull Requests'
     quiet-period: 5
     block-downstream: false


ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -3,7 +3,8 @@
     project-type: freestyle
     defaults: global
     concurrent: true
-    node: huge && bionic && rebootable && x86_64
+    # We want make check to only run on Bionic b/c it has python2 and python3 and b/c all builds should build on Bionic
+    node: huge && bionic && rebootable && x86_64 && !xenial && !trusty
```
alfredodeza: if this says `bionic && rebootable` wouldn't it be enough to avoid the `!xenial && !trusty` ?

djgalloway: No because a `xenial_trusty_pbuilder_huge` slave, for example, has the labels `[u'trusty-pbuilder', u'trusty', u'amd64', u'x86_64', u'xenial', u'jessie', u'stretch', u'huge', u'rebootable', u'bionic']`

djgalloway: We still want to be able to build Bionic Ceph packages on Xenial slaves using pbuilder however.

PR #1472, opened 2019-12-19
ceph-dev: Build containers from ceph.git branches

djgalloway:

### DIFF
--- a/ceph-dev/config/definitions/ceph-dev.yml
+++ b/ceph-dev/config/definitions/ceph-dev.yml
@@ -59,6 +59,26 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           default: "default"
           description: "Type of Ceph build, choices are: notcmalloc, default (i.e. with tcmalloc). Defaults to: 'default'"
 
+      - bool:
+          name: CI_CONTAINER
+          description: 'Build container with development release of Ceph?'
+          default: true
+
+      - string:
+          name: CONTAINER_BRANCH
+          description: "For CI_CONTAINER: Branch of ceph-container to use"
+          default: master
+
+      - string:
+          name: CONTAINER_REPO_HOSTNAME
+          description: "For CI_CONTAINER: Name of container repo server (i.e. 'quay.io')"
+          default: "quay.io"
+
+      - string:
+          name: CONTAINER_REPO_ORGANIZATION
+          description: "For CI_CONTAINER: Name of container repo organization (i.e. 'ceph-ci')"
+          default: "ceph-ci"
+
     builders:
       - multijob:
           name: 'ceph dev setup phase'
--- a/ceph-dev-build/build/build_rpm
+++ b/ceph-dev-build/build/build_rpm
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -ex
 
+# set to "true" or "false" so that both string comparisons
+# and 'if $CI_CONTAINER' work as expected.  Conventions vary across the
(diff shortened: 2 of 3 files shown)


ceph-dev/config/definitions/ceph-dev.yml
```diff
@@ -59,6 +59,26 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           default: "default"
           description: "Type of Ceph build, choices are: notcmalloc, default (i.e. with tcmalloc). Defaults to: 'default'"
 
+      - string:
```
alfredodeza: why isn't this a boolean instead?

djgalloway: `description: 'Build container with development release of Ceph.  Note: this must be "false" or "true" so that it can execute a command or satisfy a string comparison'`

alfredodeza: I am aware that the intention here was to make it easier to compare it later as a string in BASH. I am asking here because Jenkins canl set this to a real boolean which is used in other jobs. For example, in ceph-tag we have the following:

```bash
if [ "$TAG" = false ] ; then
    echo "Assuming tagging process has succeeded before because TAG was set to false"
    exit 0
fi
```

And `TAG` is created as a boolean paramater in the YAML file:

```yaml
      - bool:
          name: TAG
          description: "When this is checked, Jenkins will remove the previous private tag and recreate it again, changing the control files and committing again. When this is unchecked, Jenkins will not do any commit or tag operations. If you've already created the private tag separately, then leave this unchecked.
Defaults to checked."
          default: true
```

Using a string that looks like a boolean but isn't a real boolean so that scripts can compare it to the contents of the string is going to be misleading if others are trying to update or modify these scripts at a later point.

djgalloway: I don't disagree with you.  But presumably you already argued this with @dmick and ended up being okay with it: https://github.com/ceph/ceph-build/pull/1378

I just copied/pasted from the ceph-dev-new job.

djgalloway: @alfredodeza I changed it to a bool and tested it.

true: https://jenkins.ceph.com/job/ceph-dev-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=huge/34151/

false: https://jenkins.ceph.com/job/ceph-dev-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos7,DIST=centos7,MACHINE_SIZE=huge/34152/

alfredodeza:
These changes look good! Feel free to merge

dmick:
Uhhhh, hold up; I don't think we can just change from string to boolean here without also changing ceph-container's invocation/scripts, which I why I didn't.  Let's discuss, it's been too long since I looked all this over

dmick:
OK, after digging *way* too deeply into shell behavior, I discover that the operative thing here is that a Jenkins boolean parameter *isn't* actually a "real" boolean, but rather just a constrained string; a Jenkins boolean parameter allows only the string values "true" and "false", corresponding to "checkbox ticked" and "checkbox blank", respectively, which is just what we want, confusion about comparison context in the shell aside.

I remove my objection to changing to a boolean, and suggest we do it for all the jobs that use CI_CONTAINER.

PR #1475, opened 2019-12-20
ceph-dev*-build: do not unpack source tar ball

tchaikov:
the uncompressed tarball is not used, moreover, the build fails on
bionic when `dpkg-source -x` the source package. because we've been
relying on a bug on xenial. see
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826334;msg=31

and on bionic, the bug was fixed. so `dpkg-source -x` started to
fail.


djgalloway:
Is this tested on a `PRETTY_NAME='Ubuntu 16.04.4 LTS'` slave too?

PR #1476, opened 2019-12-24
ceph-dev: Add crimson support

djgalloway:
We weren't previously making crimson builds for ceph.git branches.  I didn't know that when updating the conditionals based on the distro matrix @liewegas provided so `crimson` got added in https://github.com/ceph/ceph-build/pull/1459.  So since it seems like we want them now, support needs to be added.

### DIFF
--- a/ceph-dev-setup/build/build
+++ b/ceph-dev-setup/build/build
@@ -82,6 +82,42 @@ else
     echo "forcing."
 fi
 
+# because autogen+configure will check for dependencies, we are forced to install them
+# and ensure they are present in the current host
+if [ -x install-deps.sh ]; then
+    echo "Ensuring dependencies are installed"
+    [ "${FLAVOR}" == "crimson" ] && WITH_SEASTAR=1
+    WITH_SEASTAR=$WITH_SEASTAR ./install-deps.sh
+fi
+
+# Flavor Builds support
+
+#cryptopp is not default for CMake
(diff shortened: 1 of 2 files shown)


djgalloway:
We weren't previously making crimson builds for ceph.git branches.  I didn't know that when updating the conditionals based on the distro matrix @liewegas provided so `crimson` got added in https://github.com/ceph/ceph-build/pull/1459

tchaikov:
@djgalloway thanks for adding crimson support in ceph-dev! i think ceph-dev is for building branches like "master", "nautilus", etc. but i think crimson is not yet mature enough for running rados qa suite or a subset of it. i'd like keep this PR open until we have better coverage of the OSD feature set.

djgalloway:
> @djgalloway thanks for adding crimson support in ceph-dev! i think ceph-dev is for building branches like "master", "nautilus", etc. but i think crimson is not yet mature enough for running rados qa suite or a subset of it. i'd like keep this PR open until we have better coverage of the OSD feature set.

That's fine with me.  I was just going by the matrix @liewegas put on the etherpad.

djgalloway:
I think we need this merged now?  From https://jenkins.ceph.com/job/ceph-dev-trigger/31139/console

```
Triggering projects: ceph-dev
ERROR: Build step failed with exception
java.lang.IllegalArgumentException: Illegal choice for parameter FLAVOR: crimson
```

tchaikov:
@djgalloway i just posted #1524 to address this.

PR #1478, opened 2020-01-09
kernel: add centos8 builds

idryomov:


idryomov:
A test job with a centos8 kernel has passed:

http://pulpito.ceph.com/dis-2020-01-09_19:55:11-krbd-master-wip-buildtest-basic-smithi/

PR #1485, opened 2020-01-16
ceph-pull-requests-arm64: only build on centos8 and bionic

tchaikov:
since we've labeled the slaves correctly. and centos8 is able to build
ceph on arm64. the only pending issue is
https://bugzilla.redhat.com/show_bug.cgi?id=1673990 . but we can fix it
manually.

### DIFF
--- a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
+++ b/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
@@ -9,7 +9,7 @@
     concurrent: true
     disabled: false
     name: ceph-pull-requests-arm64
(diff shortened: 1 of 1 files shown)


djgalloway:
This will limit this job to *only* CentOS8.  We also want the Ubuntu slaves to run this job.  So we only want the `arm64` label.

djgalloway:
```
<kefu> dgalloway how about:  arm64 && (centos8 || bionic) && !xenial && !trusty ?
<kefu> as centos7 is missing the build dependencies on arm64.
<dgalloway> well there are no trusty arm64 builders
<dgalloway> or xenial anymore
<kefu> nice!
<dgalloway> so 'arm64 && !centos7'
```

PR #1488, opened 2020-01-22
ceph-*-setup: do not run install-deps.sh

tchaikov:
this is a follow-up change of #1487. since we don't use autotools for
preparing dist tarball, there is no need to run install-deps.sh for
installing the autotools.


djgalloway:
:eyes:

PR #1497, opened 2020-02-05
ceph-dashboard-* jobs: Install Prometheus and Alertmanager

callithea:
In order to test the alert configuration within the dashboard
(e2e tests) we need to have Prometheus and Alertmanager running
on the system. Added a new script which takes care of the installation
and running Prometheus and Alertmanager in the background.
Added the script only to the dashboard e2e Jenkins job (for the time being),
the processes will be killed at the end of the job.


callithea:
Edit: Killing the processes should not be required since the workspace dir will be wiped at the end of the job.

PR #1498, opened 2020-02-07
Disabling ceph-pull-requests-arm64 until we have packages and reliable hw

djgalloway:

### DIFF
--- a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
+++ b/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
@@ -67,4 +67,6 @@
         failure-status: "make check failed"
         white-list-target-branches:
         - master
+        white-list-labels:
+          - arm64
     wrappers: []


ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -7,7 +7,7 @@
           - ../../../scripts/build_utils.sh
           - ../../build/build
     concurrent: true
-    disabled: false
+    disabled: true
```
tchaikov: can we instead add
```yaml
white-list-labels:
            - arm
```

liewegas: what is the difference?

my goal is to make the little dot next a PR green (instead of red every time, requiring you to open the PR page to see whether it was the arm test or something important that failed)

tchaikov: the difference is that this job is disabled by default. and it's triggered only if "arm64" label is added.

liewegas: :+1:

djgalloway: Fixed

PR #1501, opened 2020-02-12
Add nightly support for ceph-volume centos8 tests

jan--f:
and remove centos7 and xenial from master and nautilus runs.

### DIFF
--- a/ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
+++ b/ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
@@ -3,6 +3,7 @@
     distro:
       - xenial
       - centos7
+      - centos8
     objectstore:
       - bluestore
       - filestore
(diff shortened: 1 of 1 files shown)


ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
```diff
@@ -16,6 +17,19 @@
       - mimic
       - luminous
       - nautilus
+    exclude:
+      - ceph-branch: master
+        distro: centos7
+      - ceph-branch: nautilus
+        distro: centos7
```
dsavineau: I think you want to exclude centos 8 for nautilus and not centos 7 right ?

ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
```diff
@@ -59,6 +74,19 @@
...
```
dsavineau: ditto

ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
```diff
@@ -80,6 +109,19 @@
       - nautilus
       - mimic
       - luminous
+    exclude:
+      - ceph-branch: master
+        distro: centos7
+      - ceph-branch: nautilus
+        distro: centos7
```
dsavineau: ditto

ceph-volume-nightly/config/definitions/ceph-volume-nightly.yml
jan--f: Hmm I intended to skip centos7 originally, since the nautilus nightlies were build with ceph-ansible master. But I have since changed that to ceph-ansible 4.0.

That hasn't fixed the nightlies however. Looking into it.

jan--f: so you're right, nautilus will be tested with centos7. I have a few fixes that get the nautilus tets to run again.

PR #1508, opened 2020-02-25
ceph-pull-requests: Behave like arm64 job

djgalloway:
Debugging unit test failures was difficult on x86 jobs because the slave got rebooted right away.

### DIFF
--- a/ceph-pull-requests/build/build
+++ b/ceph-pull-requests/build/build
@@ -2,4 +2,4 @@
 export NPROC=$(nproc)
 timeout 3h ./run-make-check.sh
 sleep 5
-ps -ef | grep ceph || true
+ps -ef | grep -v jnlp | grep ceph || true
--- a/ceph-pull-requests/build/build
+++ b/ceph-pull-requests/build/build
@@ -1,0 +1,5 @@
+#!/bin/bash -ex
+export NPROC=$(nproc)
+timeout 3h ./run-make-check.sh
+sleep 5
+ps -ef | grep ceph || true
--- a/ceph-pull-requests-arm64/build/build
+++ b/ceph-pull-requests-arm64/build/build
@@ -5,4 +5,4 @@ export CHECK_MAKEOPTS="-j${n_test_jobs}"
 export BUILD_MAKEOPTS="-j${n_build_jobs}"
 timeout 3h ./run-make-check.sh
 sleep 5
-ps -ef | grep ceph || true
+ps -ef | grep -v jnlp | grep ceph || true
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -56,17 +56,12 @@
           wipe-workspace: true
 
     builders:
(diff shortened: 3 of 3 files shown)


ceph-pull-requests/build/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -ex
+export NPROC=$(nproc)
+timeout 3h ./run-make-check.sh
+sleep 5
+ps -ef | grep ceph || true
```
dillaman: Changes lgtm -- just wondering what the purpose of this line is for if you know off-hand? It seems like it would just show any accidentally running Ceph daemons but it wouldn't clean-up from such a failure? I wonder if that was why the reboot was put in originally?

djgalloway: I'm guessing your assumption is correct.  I don't know enough about the Ceph building process to suggest a better alternative though.  Does a rogue daemon ever get left?

Now that I go back and look at the job config, it is configured to only run on ephemeral slaves so it was probably easiest to just reboot to make sure rogue daemons got killed.  Perhaps we could `kill -9` any PIDs that show up in `ps` output?

@tchaikov I'm curious what you'd suggest here.

dillaman: I guess presuming that we have non-ephemeral Jenkins builders, I suppose it shouldn't make a big difference since we weren't cleaning up any potential left-over daemons before.

dillaman: Looks like it was added in PR #284 to fix an issue w/ left-over processes causing build failures [1]. Can we throw a different exit code if any ceph processes remain and have the builder only reboot if said exit code is detected? 

[1] https://tracker.ceph.com/issues/14172

djgalloway: So I was thinking of just having PIDs with "ceph" in the process name killed.  I checked what this might look like on an existing slave and discovered that if this `ps -ef` command was actually being run, it would expose the API key used for slave agents to connect to Jenkins.  But that's not happening (thank goodness).

```
dgalloway@slave-ubuntu01:~$ ps -ef | grep ceph
jenkins+  1019     1  1  2019 ?        3-01:08:29 /usr/bin/java -jar /home/jenkins-build/slave.jar -jnlpUrl https://jenkins.ceph.com/computer/172.21.1.35+slave-ubuntu01/slave-agent.jnlp -jnlpCredentials OOPS:NOPE
```

But this means that even in the `ceph-pull-requests-arm64` job, the job quits immediately after failure and the `ps` command never gets run.  So we're back to not knowing what the difference is between the arm64 job and the x86 job.

Are any of the available ceph-pull-requests-arm64 job failures good examples I can use to look at the differences between the two?  https://jenkins.ceph.com/job/ceph-pull-requests-arm64/

dillaman: > So we're back to not knowing what the difference is between the arm64 job and the x86 job.

The reboot is still a difference, no?

djgalloway: Yeah, it is.  I can update and we can merge as-is to see if it helps.  Can't hurt anything assuming daemons aren't getting leftover.

PR #1509, opened 2020-02-25
osc: Build octopus on openSUSE Leap 15.2

toabctl:
We want to test with the upcoming openSUSE release (15.2) instead of
15.1.

Signed-off-by: Thomas Bechtold <tbechtold@suse.com>


toabctl:
@kshtsk could you have a look please? Is that enough to get packages build on Leap 15.2 ?

PR #1516, opened 2020-03-05
Test teuthology PRs on Bionic & install python3-dev

djgalloway:


djgalloway:
Whew.  Okay @tchaikov I don't know how adding that one task created such a long chain reaction but it somehow broke the playbook on Xenial.  Maybe if it's run locally on the ephemeral slave (like it is with prado), it works but from my workstation, I had to set the `ansible_python_interpreter` right before the Jenkins registration task.  This is good to go now though.

djgalloway:
@kshtsk FYI

kshtsk:
@djgalloway thanks

PR #1527, opened 2020-03-16
ceph-dev-trigger: Fix regex for master conditional step

djgalloway:


dmick:
Why does this fix it?  What is the name of the branch we have to match?

djgalloway:
> Why does this fix it? What is the name of the branch we have to match?

The actual `GIT_BRANCH` is `origin/master`.  So `.*master.*` matches but `master` didn't.

PR #1543, opened 2020-04-01
ceph-perf-pull-requests: install cbt dependencies for el8

tchaikov:
since all "performance" slaves are now el8, we need to update the
dependencies for cbt accordingly


tchaikov:
this should address the failure of https://jenkins.ceph.com/job/ceph-perf-pull-requests/699/console

PR #1554, opened 2020-04-21
DNM: Allow installing different pip versions

djgalloway:
And have ceph-ansible-prs job use the latest.

### DIFF
--- a/ceph-ansible-prs/build/build
+++ b/ceph-ansible-prs/build/build
@@ -2,7 +2,7 @@
 
 # the following two methods exist in scripts/build_utils.sh
 pkgs=( "tox" )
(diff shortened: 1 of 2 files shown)


djgalloway:
Don't merge this yet.  @tchaikov and I are investigating thinking this might be a more widespread issue.

Up until today, the `easy_install --upgrade pip` was installing 20.0.2.  Now it's installing 20.1b1 which has broken the CI for all builds except centos7.  Kefu noticed the only difference is centos7 installs pip via tarball while all others use wheels.

djgalloway:
https://github.com/ceph/ceph-build/pull/1556

PR #1557, opened 2020-04-24
Add support for Ubuntu Focal Fossa globally

djgalloway:

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -158,7 +158,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=bionic
+                    DISTROS=focal
                     FLAVOR=crimson
 
     wrappers:
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -104,7 +104,7 @@
                     DISTROS=centos7
                     FLAVOR=notcmalloc
       # build octopus on:
-      # default: bionic centos7 centos8 leap15
+      # default: focal bionic centos7 centos8 leap15
       # notcmalloc: centos8
       - conditional-step:
           condition-kind: regex-match
@@ -121,7 +121,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=bionic centos7 centos8 leap15
+                    DISTROS=focal bionic centos7 centos8 leap15
                 - project: 'ceph-dev-new'
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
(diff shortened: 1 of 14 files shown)


scripts/build_utils.sh
```diff
@@ -592,6 +597,8 @@ use_ppa() {
                         use_ppa=false
                     fi
                     ;;
+                focal)
+                    use_ppa=true;;
```
djgalloway, line 601: @tchaikov before this merges, I was hoping you could take a peek at all the changes in this file.  I'm not sure what GCC version should be used or what ppa is, etc.

djgalloway:
Also, do we went to have crimson builds for 20.04 AND 18.04?  @jdurgin mentioned we might want to add CentOS8 crimson builds.

scripts/build_utils.sh
tchaikov: > @tchaikov before this merges, I was hoping you could take a peek at all the changes in this file. I'm not sure what GCC version should be used or what ppa is, etc.

ubuntu focal comes with GCC-9.3, which would suffice for building ceph and crimson.

> 
> I manually pushed the job config changes to the `ceph-dev` job and built Octopus for Ubuntu 20.04 (Focal Fossa). Looks like maybe GCC 10 was used?

```
+ use_ppa=true
...
-- The CXX compiler identification is GNU 9.3.0
-- The C compiler identification is GNU 9.3.0
```

some gcc-10 runtime was downloaded. because the ubuntu-toolchain-r PPA was added somehow.

tchaikov:
> Also, do we went to have crimson builds for 18.04 AND 20.04?

currently we have crimson flavor built on bionic, but since ubuntu focal ships with GCC-9, presumably, it'd be faster to prepare the dependencies. so i'd suggest s/bionic/focal/

> @jdurgin mentioned we might want to add CentOS8 crimson builds.

yeah. but i think we can postpone it until one of us starts working on cephadm for crimson support.

scripts/build_utils.sh
```diff
@@ -710,6 +717,8 @@ setup_pbuilder_for_old_gcc() {
             old=5;;
         bionic)
             old=8;;
+        focal)
+            old=8;;
```
tchaikov, line 721:
```suggestion
            old=9;;
```

scripts/build_utils.sh
```diff
@@ -722,7 +731,7 @@ setup_pbuilder_for_ppa() {
         rm -rf $hookdir
         mkdir -p $hookdir
         local gcc_ver=7
-        if [ "$DIST" = "bionic" ]; then
+        if [ "$DIST" = "bionic" ] || [ "$DIST" = "focal" ]; then
```
tchaikov, line 734: could drop this change. as focal offers gcc-9 already.

scripts/build_utils.sh
tchaikov, line 601: drop this change, as focal comes with gcc-9 which is good enough for us at this moment.

djgalloway:
Okay @tchaikov.  Thanks for your review!  I think this is ready to go now.

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -158,7 +158,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=bionic
+                    DISTROS=focal
```
tchaikov, line 161: ack!

PR #1558, opened 2020-04-27
ceph-dashboard-prs: Change order of install dep script

callithea:
Not sure if it is related to the way the previous two lines
of dependency installation is formatted, but the install line
for xvfb was never executed in Jenkins.
But it does not really make sense to me since the line `sudo rm /etc/apt/sources.list.d/google-chrome.list` that was previously right after the splitted `sudo apt-get install -y [...]` line was always executed.
Could it be that the job was never updated with the changes of PR #1552?

### DIFF
--- a/scripts/dashboard/install-e2e-test-deps.sh
+++ b/scripts/dashboard/install-e2e-test-deps.sh
@@ -11,9 +11,9 @@ if grep -q  debian /etc/*-release; then
     curl -fsSL https://dl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
     sudo apt-get update
     sudo apt-get install -y google-chrome-stable
+    sudo apt-get install -y xvfb
     sudo apt-get install -y python-requests python3-requests python-openssl python3-openssl python-jinja2 python3-jinja2 \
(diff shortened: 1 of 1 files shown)


callithea:
Tested the changes with https://jenkins.ceph.com/job/ceph-dashboard-pull-requests/3987/console (see line `+ sudo apt-get install -y xvfb` in the log), previously e.g. here: https://jenkins.ceph.com/job/ceph-dashboard-pull-requests/3984/console `xvfb` was not installed.

tchaikov:
hi @callithea , i don't understand why this change addresses the failures. to make sure we get it fixed, could you pin me once you see it again?

callithea:
> hi @callithea , i don't understand why this change addresses the failures. to make sure we get it fixed, could you pin me once you see it again?

Hi @tchaikov it was just for testing purposes (because I had the feeling that the job was never updated with the changes of https://github.com/ceph/ceph-build/pull/1552), so I just moved the line and manually pushed it to Jenkins - and all of a sudden the pkg was installed, while it previously was simply not installed when the script was executed in Jenkins.

PR #1560, opened 2020-04-28
support crimson flavor for centos 8

ideepika:
supports crimson flavor for centos 8    
    add crimson flavor build for octopus and master

See Also: https://github.com/ceph/ceph-build/pull/1334/

### DIFF
--- a/ceph-dev-new-build/build/build_rpm
+++ b/ceph-dev-new-build/build/build_rpm
@@ -187,7 +187,8 @@ fi
 
 # XXX perhaps use job parameters instead of literals; then
 # later stages can also use them to compare etc.
-if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && $FLAVOR == "default" ]] ; then
+# build container image that supports building crimson-osd
+if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && $FLAVOR != "notcmalloc" ]] ; then
     loop=0
     ready=false
     while ((loop < 15)); do
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -158,7 +158,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal
+                    DISTROS=focal centos8
                     FLAVOR=crimson
 
     wrappers:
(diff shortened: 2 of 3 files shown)


ideepika:
hey @djgalloway! I wanted to test package builds for this PR. does it require special privilege to test a branch other than master for ceph-builds?

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -128,6 +128,38 @@
                     FORCE=True
                     DISTROS=centos8
                     FLAVOR=notcmalloc
+      # We need only centos8 for crimson ceph container for now
```
djgalloway, line 131: This isn't necessary.  WIP branches are detected in this last conditional step: https://github.com/ceph/ceph-build/blob/master/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml#L131-L162

You only need to add `centos8` after `focal` here: https://github.com/ceph/ceph-build/blob/master/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml#L161

ceph-dev-new-build/build/build_rpm
```diff
@@ -187,7 +187,8 @@ fi
 
 # XXX perhaps use job parameters instead of literals; then
 # later stages can also use them to compare etc.
-if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && $FLAVOR == "default" ]] ; then
+# build container image that supports building crimson-osd
+if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && $FLAVOR == "crimson" ]] ; then
```
djgalloway, line 191: We still want containers for `default` and `centos8`, right?  If so, you should just change this to:

```
if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && $FLAVOR != "notcmalloc" ]] ; then
```

ideepika: I was concerned do we want to have both container builds everytime, but since containers cache there penultimate builds, this won't cost much I suppose. Updated!

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
ideepika: ack!

ideepika:
@tchaikov @djgalloway i don't see crimson centos8 flavor(for wip branches) build in recent shaman builds, everything seems fine to me, can you take another look?

djgalloway:
I determined it was because the job that updates our Jenkins jobs broke a couple weeks ago.  I fixed it and the job configs are up to date now so we'll start getting CentOS8 crimson builds for wip branches again.

ideepika:
Aah I see, thanks for the update!

PR #1568, opened 2020-05-13
kernel: update KConfig to 5.7.0-rc5-ish

jtlayton:
Much has changed in v3.19, and we're hitting some issues with missing
options that are needed to do netfiltering correctly. Update to a much
more recent config.

The RPM one is based on Fedora 32 (kernel-5.6.11-300.fc32.x86_64)
and the debian one from Ubuntu Focal Fossa
(linux-source-5.4.0_5.4.0-31.35_all.deb). I then ran it through
make olddefconfig on the latest ceph-client/testing branch.


jtlayton:
Note that I haven't actually tested this, but I'm pretty sure we're going to need to update these configs in some fashion.

lxbsz:
Yeah, this is introducing many other new options which we will never use here, such as only for the INPUT_{JOYSTICK/TABLET/TOUCHSCREEN/MISC} four options, they will introduce almost 200 followed ones.

Maybe we can base the RHEL 8's /boot/config-4.18.0-147.el8.x86_64 , then check and enable some others those are new and necessary ? Checked the Netfilter options they look okay and work for me.

lxbsz:
There has one related tricker: https://tracker.ceph.com/issues/45531.

jtlayton:
Alternately, we could take the existing configs and just turn on the options we need. I am just wondering what else we're missing, given that we have missed a whole pile of netfilter option changes.

@idryomov When you updated the configs last time, how did you go about doing it? Should we take a different approach here?

idryomov:
I would go with manually editing in the netfilter stuff into the existing configs.  We've done this for ceph security xattrs, tcmu and other stuff.

jtlayton:
Ok, let's drop this PR then, and maybe @lxbsz can just update the existing configs.

lxbsz:
Okay, let me enable the entfilter ones with manually editing them.

PR #1579, opened 2020-06-02
ceph-*-build: remove focal for crimson flavor

ideepika:
we have now shifted to using centos 8 as base for building crimson
instead of using focal.


ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -158,7 +158,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal centos8
+                    DISTROS= centos8
```
tchaikov, line 161: could you add a prefix for the title of the commit message? and please remove the space after "=".

ideepika: ack.

PR #1582, opened 2020-06-03
ceph-deploy-pull-requests: s/installed/present/

tchaikov:
see
https://docs.ansible.com/ansible/latest/modules/apt_module.html#parameters

the desired package state should be "present" instead of "installed".

### DIFF
--- a/ceph-deploy-pull-requests/setup/playbooks/tasks/ubuntu.yml
+++ b/ceph-deploy-pull-requests/setup/playbooks/tasks/ubuntu.yml
@@ -3,7 +3,7 @@
     action: apt update_cache=yes
 
   - name: install python requirements
-    action: apt pkg={{ item }} state=installed
+    action: apt pkg={{ item }}
     with_items:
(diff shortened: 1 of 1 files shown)


ceph-deploy-pull-requests/setup/playbooks/tasks/ubuntu.yml
```diff
@@ -3,7 +3,7 @@
     action: apt update_cache=yes
 
   - name: install python requirements
-    action: apt pkg={{ item }} state=installed
+    action: apt pkg={{ item }} state=present
```
rzarzynski, line 6: Do we need to define the parameter at all? [The default is actually `present`](https://docs.ansible.com/ansible/latest/modules/apt_module.html#parameter-state).

tchaikov: @rzarzynski thanks! updated and repushed.

PR #1589, opened 2020-06-16
nfs-ganesha: enable CentOS 8 builds, other fixes

tserlin:
Enable CentOS 8 builds of the "next" branch of nfs-ganesha.

We also needed other fixes to the build script to account for both el8/el7 builds, as well as handling for the newer version of "mock" which now uses templates.


tserlin:
@alimaredia Can you throw another set of eyes on this?

I actually manually added the changes here, and we have successful builds, so this has been tested:
https://jenkins.ceph.com/job/nfs-ganesha/1295/

However, my manual changes will be reverted the next time JJB runs.

PR #1591, opened 2020-06-16
nfs-ganesha-stable: Enable CentOS 8 builds of the stable branches of nfs-ganesha

tserlin:
We also needed other fixes to the build script to account for both el8/el7 builds, as well as handling for the newer version of "mock" which now uses templates.


tserlin:
This is similar to my PR at https://github.com/ceph/ceph-build/pull/1589

PR #1602, opened 2020-06-24
ceph-volume: don't run simple func tests against PRs

rishabh-d-dave:
Since "simple" functional tests are meant to test that ceph-volume
successfully takes over ceph-disk deployed OSDs.


guits:
no pull request description? no commit message? 🙂 

@rishabh-d-dave Is that because these jobs don't make sense as of nautilus or something like this?

rishabh-d-dave:
@guits Right, the context is not apparent. I've updated the commit message as well as the PR's description now.

guits:
so we don't test it at all? even on luminous|nautilus branches?

rishabh-d-dave:
Perhaps we only need it for nautilus...
@jan--f @andrewschoen Any thoughts?

PR #1603, opened 2020-06-25
ceph-dev-new-build: fail, if building containers fail

sebastian-philipp:
This avoids:

* unnecessary breakages in the master branch
* unnecessary pulpito runs due to failing containers

Let's fail early, instead of later in the pulpilto runs.

See

* https://github.com/ceph/ceph/pull/35767
* http://pulpito.ceph.com/swagner-2020-06-25_11:37:21-rados:cephadm-wip-swagner-testing-2020-06-25-1124-distro-basic-smithi/

Signed-off-by: Sebastian Wagner <sebastian.wagner@suse.com>

### DIFF
--- a/ceph-dev-build/build/build_rpm
+++ b/ceph-dev-build/build/build_rpm
@@ -203,8 +203,7 @@ if [[ "$CI_CONTAINER" == true && $DISTRO == "centos" && $FLAVOR == "default" ]] 
     fi
 
     cd $WORKSPACE/ceph-container
-    # avoid failing the build if build-push fails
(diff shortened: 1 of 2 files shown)


kshtsk:
How is this related to pulpito? Or you meant just teuthology.

sebastian-philipp:
```
[16:40:11] <dgalloway> SebastianW, the problem with failing the build of the container build fails is technically we do have a functional repo.  but shaman will believe the build failed entirely so if a teuthology job that just installs packages (instead of using containers), it will ask shaman if there's a repo ready and shaman will say no even though there is
```

djgalloway:
If we're going to do this, is also needs to be applied to the `ceph-dev-build` job.

sebastian-philipp:
> If we're going to do this, is also needs to be applied to the `ceph-dev-build` job.

done

PR #1606, opened 2020-06-30
ceph-dev*build: Tell shaman the build passed before container build

djgalloway:
Another follow-up to https://github.com/ceph/ceph-build/pull/1603.  If the build passed and a repo got created, we can tell shaman.

If we want shaman to have information about failed/passed container builds, then that should be coded into shaman as an additional field.


sebastian-philipp:
... which makes shaman unusable for to know when to start teuthology runs. Do you think you can give our group (@varshar16 @mgfritch , @jschmid1 , @yuriw , Kefu and @ideepika )  way to know when the container builds worked?

PR #1607, opened 2020-07-02
ceph-dev-*-build: reset CI_CONTAINER when building nautilus

tchaikov:

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -1072,3 +1072,12 @@ get_nr_build_jobs() {
     fi
     echo $n_build_jobs
 }
+
+maybe_reset_ci_container() {
+    # ceph-container includes ceph-mgr-cephadm by default, but nautilus does
+    # not ship this package
+    if [[ "$CI_CONTAINER" && "$BRANCH" =~ nautilus ]]; then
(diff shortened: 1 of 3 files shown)


scripts/build_utils.sh
```diff
@@ -1072,3 +1072,12 @@ get_nr_build_jobs() {
     fi
     echo $n_build_jobs
 }
+
+maybe_reset_ci_container() {
+    # ceph-container includes ceph-mgr-cephadm by default, but nautilus does
+    # not ship this package
```
sebastian-philipp, line 1078: which sounds strange. isn't it the job of ceph-container to decide which ceph version ships which packages? anyway. this PR lgtm.

tchaikov: > which sounds strange. isn't it the job of ceph-container to decide which ceph version ships which packages? 

right. but nautilus does not have ceph-mgr-cephadm at all. ceph-container cannot ship the package which does not even exist in upstream.

tchaikov:
https://shaman.ceph.com/builds/ceph/nautilus/120f482407ab4070126cb7ffc2f4f1b37517b4a0/

PR #1611, opened 2020-07-09
ceph-release-rpm: Drop centos6

djgalloway:


ceph-release-rpm/config/definitions/ceph-release-rpm.yml
```diff
@@ -9,7 +9,7 @@
     parameters:
       - string:
           name: RELEASE
-          default: firefly
+          default: pacific
```
tchaikov, line 12: so, after the change, the chacra URL will be `ceph-release/pacific/HEAD/centos/8/noarch` for non-test centos8 builds. makes sense to me.

PR #1618, opened 2020-07-15
ceph-perf-pull-requests: pass job id and job url to check

tchaikov:
so we can check the details of check run


tchaikov:
for a sample, see https://github.com/ceph/ceph/pull/28554/checks

PR #1637, opened 2020-08-12
ceph*setup: Give ceph*build jobs permission to copy artifacts

djgalloway:
This is a new setting in the newest CopyArtifact plugin.

https://stackoverflow.com/a/32084538/6225562


djgalloway:
This is an emergency fix and I've tested it so merging.

PR #1647, opened 2020-08-24
ceph-*pull-requests*: cancel job on PR update

epuertat:
This should save quite a few resources, as right now if a PR is updated several times in a short lapse of time (typical case when the submitted realizes that they missed some files/changes...), the already started jobs will continue until finishing.

Signed-off-by: Ernesto Puerta <epuertat@redhat.com>


callithea (approved):
lgtm, thanks!

tchaikov (approved):

PR #1651, opened 2020-08-27
bz-mill: Scrap job

djgalloway:
This job gets stuck sometimes and with no timeout, it holds up the docs Jenkins agent from running any other docs jobs.


tchaikov:
thanks. i've been killing this job more than 5 times.

PR #1654, opened 2020-09-04
kernel: drop centos7 builds and arm64 cruft

idryomov:


idryomov:
https://jenkins.ceph.com/job/kernel/731/

PR #1658, opened 2020-09-15
ansible: Pass API key to slave.jar as file

djgalloway:
vstart_runner.py runs 'ps ww' for the jenkins-build user.  This exposes the API key by showing the java.jar process in Jenkins logs.  If we pass the credentials as a file, there's less of a chance of it getting exposed.


epuertat (approved):
Good catch! :+1:

PR #1662, opened 2020-09-20
ceph-dev{-new}: add build names

dmick:
Name builds by their build number as well as branch, distros, and flavor,
for easier viewing on the Jenkins UI.  Uses a new-to-us jenkins plugin.


ktdreyer:
Linking the JJB doc here for posterity: https://docs.openstack.org/infra/jenkins-job-builder/wrappers.html#wrappers.build-name

PR #1666, opened 2020-09-25
ceph-dev{,-new}: remove duplicated copyartifact actions

tchaikov:
they are already included by corresponding ceph-dev-build and
ceph-dev-new-build jobs.

### DIFF
--- a/ceph-dev-new/config/definitions/ceph-dev-new.yml
+++ b/ceph-dev-new/config/definitions/ceph-dev-new.yml
@@ -88,12 +88,6 @@ If this is checked, then the binaries will be built and pushed to chacra even if
             - name: ceph-dev-new-setup
               current-parameters: true
               exposed-scm: false
-      - copyartifact:
-          project: ceph-dev-new-setup
-          filter: dist/sha1
-          which-build: multijob-build
-      - inject:
-          properties-file: ${WORKSPACE}/dist/sha1
       - multijob:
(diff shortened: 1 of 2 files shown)


tchaikov:
- before this change: https://jenkins.ceph.com/job/ceph-dev-new/46892/
- after this change: https://jenkins.ceph.com/job/ceph-dev-new/46895/

djgalloway:
I'm deferring to @dmick for this one.

dmick:
Two separate copyartifact calls arenexactly the intent.

The point of the original PR was to add a second copyartifact to ceph-dev and ceph-dev-new, the parent jobs.  Copy takes a source and dest, and the source is -setup.  The original dest was -build, which meant -build could access sha1, but not the parent jobs.  I wanted the parent jobs to have access, so I added them as second dests.  This meant adding the build steps to copy and then inject the sha1 to -dev and -dev-new.

But one also must add permission to -setup for each job that can copyartifact, by name.  While the original perm used the bare job name -build, and all my testing showed that adding a comma and the parent job name allowed the new second copy, for some reason it wasn't enough for ceph-dev in particular.  Adding a / to make the job name "absolute" fixed the problem.

dmick (changes requested):
We don't want this change.

dmick:
I see that ceph-dev is still failing, and ceph-dev-new is still working.  I'm continuing to try to understand why.

PR #1667, opened 2020-09-25
Fix intermittent error with last copyartifacts change

dmick:
Apparently, sometimes, Jenkins can't find a toplevel job by "name", and
one must specify an "absolute path" to the job by prefixing "/".  This
was causing ceph-dev builds to fail to find ceph-dev-setup, but changing
to /ceph-dev-setup fixed the failure-to-find.  It didn't affect ceph-dev-new.
I've no idea why it affected one and not the other.


dmick:
@tchaikov

PR #1668, opened 2020-09-28
ceph: enable multibranch pipeline

epuertat:
* [Jenkins Job Builder Multibranch Pipeline](https://docs.openstack.org/infra/jenkins-job-builder/project_workflow_multibranch.html)

This change only enables the `Jenkinsfile` detection. For this to work, a `Jenkinsfile` needs to be added to `ceph/ceph`: https://github.com/ceph/ceph/pull/37425

So, it basically means that this change alone doesn't work. `Jenkinsfile` needs to propagate across branches & PRs to start seeing it (basically as soon as it hits a major branch, `master`/`octopus`/etc...).

## TODO
- [ ] Install [Jenkins Timeout Plugin](https://plugins.jenkins.io/build-timeout/) for [Timeout config](https://docs.openstack.org/infra/jenkins-job-builder/wrappers.html?highlight=timeout#wrappers.timeout). Timeouts could also be [set at different levels in `Jenkinsfile`s](https://www.jenkins.io/doc/book/pipeline/syntax/), but a max. global timeout should be set as a safeguard. 

#### BlueOcean

image

#### Classic

Unlike regular jobs, Multibranch rather looks like a folder of jobs:

image

#### BlueOcean

image

#### Classic

image

Signed-off-by: Ernesto Puerta <epuertat@redhat.com>

### DIFF
--- a/ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
+++ b/ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
@@ -1,0 +1,18 @@
+---
+- job:
+    name: ceph-multibranch-pipeline
+    project-type: multibranch
+    number-to-keep: 300
+    days-to-keep: 30
+    scm:
+      - github:
+          repo: ceph
+          repo-owner: ceph
+          ssh-checkout:
+            credentials: 'jenkins-build'
+          credentials-id: 8cffdeb4-283c-4d96-a190-05d5645bcc2f
+          clean:
+            before: true
+          shallow-clone: true
+          do-not-fetch-tags: true
+          discover-pr-forks-trust: permission


ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
```diff
@@ -0,0 +1,16 @@
...
+- job:
+    name: ceph-multibranch-pipeline
+    project-type: multibranch
+    number-to-keep: 600
+    days-to-keep: 30
+    scm:
+      - github:
+          repo: ceph
+          repo-owner: ceph
+          credentials-id: 8cffdeb4-283c-4d96-a190-05d5645bcc2f
```
tspmelo, line 11: is it OK to have this here?

djgalloway: Yes.  It's just a UUID Jenkins uses to identify which encrypted credentials to use.

ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
```diff
@@ -0,0 +1,16 @@
+---
+- job:
+    name: ceph-multibranch-pipeline
+    project-type: multibranch
+    number-to-keep: 600
```
djgalloway, line 5: Will this job contain the console logs for the entirety of `make check` and `API tests` jobs?  If so, 600 is far too many.

ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
epuertat: I think this is the 'name'/id of the credentials. Perhaps @djgalloway knows a better way of putting this?

djgalloway: > I think this is the 'name'/id of the credentials. Perhaps @djgalloway knows a better way of putting this?

It's fine :+1:

ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
```diff
@@ -0,0 +1,16 @@
...
+    days-to-keep: 30
```
callithea, line 6: Should we also reduce `days-to-keep` to 15 (also set in other jobs) or leave it like it is?

ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
epuertat: No problem. I'll reduce it to 300 (but keep the 60 day, so it's limited by amount rather than time).

ceph-multibranch-pipeline/config/definitions/ceph-multibranch-pipeline.yml
epuertat: Probably the number-to-keep will always override the days-to-keep setting, but I'd personally keep as many days as possible. There's this case where a PR is opened, all checks are run by then, and after 20-30 days of discussion (but no new commits/force-pushes) it becomes merged. If that causes an issue after merging, there's no logs to find out what happened. An alternative would be to publish comments in the PR with the test results...

toabctl (commented):
Nice to use pipelines in the future!

PR #1671, opened 2020-09-30
ceph-pr-commits: Don't require signed commits for doc/ dir

djgalloway:

### DIFF
--- a/ceph-pr-commits/build/build
+++ b/ceph-pr-commits/build/build
@@ -1,9 +1,25 @@
 #!/bin/bash
 
+# Don't require signed commits if only docs changed.
+# Modified from ceph-docs/build/build.
+# I tried using the excluded-regions parameter for the ghprb plugin but since
+# this job/check is required, it hung with 'Expected — Waiting for status to be reported'
+cd "$WORKSPACE/ceph"
+if git rev-parse --verify ${GIT_COMMIT}^2; then
+     files="$(git diff --name-only ${GIT_COMMIT}^1...${GIT_COMMIT}^2)"
+     echo -e "changed files:\n$files"
+     if [ $(echo "$files" | grep -v '^doc/' | wc -l) -gt 0 ]; then
+         echo "Not a docs only change.  Will proceed with signed commit check."
+         cd -
+     else
+         echo "Only the doc/ dir changed.  No need to check for signed commits."
+         exit 0
+     fi
+fi
+
 # the following two methods exist in scripts/build_utils.sh
 pkgs=( "pytest" )
 install_python_packages "pkgs[@]"
-
 
 cd "$WORKSPACE"
 timeout 600 $VENV/py.test -vs --junitxml="$WORKSPACE/report.xml" "$WORKSPACE/ceph-build/ceph-pr-commits/build/test_commits.py"


jdurgin (approved):
Thanks!

tchaikov:
@djgalloway @jdurgin 

this change breaks the ceph-pr-doc check for doc only changes. 
```
+ '[' 0 -gt 0 ']'
+ echo 'Only the doc/ dir changed.  No need to check for signed commits.'
Only the doc/ dir changed.  No need to check for signed commits.
+ exit 0
Recording test results
ERROR: Step ?Publish JUnit test result report? failed: Test reports were found but none of them are new. Did leafNodes run? 
For example, /home/jenkins-build/build/workspace/ceph-pr-commits/report.xml is 1 hr 9 min old

[Checks API] No suitable checks publisher found.
```

see https://jenkins.ceph.com/job/ceph-pr-commits/33731/console

tchaikov:
i reverted this change and deployed the reverted version to address the false alarms of the doc-only PRs.

djgalloway:
> i reverted this change and deployed the reverted version to address the false alarms of the doc-only PRs.

Did you do this manually?

tchaikov:
i did it manually.

ceph-pr-commits/build/build
```diff
@@ -1,9 +1,25 @@
 #!/bin/bash
 
+# Don't require signed commits if only docs changed.
+# Modified from ceph-docs/build/build.
+# I tried using the excluded-regions parameter for the ghprb plugin but since
+# this job/check is required, it hung with 'Expected — Waiting for status to be reported'
+cd "$WORKSPACE/ceph"
+if git rev-parse --verify ${GIT_COMMIT}^2; then
```
tchaikov, line 8:
```
+ pushd .
~/build/workspace/ceph-pr-commits ~/build/workspace/ceph-pr-commits
+ cd /home/jenkins-build/build/workspace/ceph-pr-commits/ceph
+ git rev-parse --verify 'e9dd427c6dec2b6282d4e2643f79df1eb639f256^2'
fatal: Needed a single revision
```

@djgalloway @jdurgin i spotted this failure in https://jenkins.ceph.com/job/ceph-pr-commits/33940/console .

any idea?

djgalloway: Yeah, I could probably use @dmick's expertise here.  When I was researching the best ways to just "list what files changed in a PR," `git rev-parse ${GIT_COMMIT}^2` was what came up.  Maybe we can get rid of `--verify`?

```
       --verify
           Verify that exactly one parameter is provided, and that it can be turned into a raw 20-byte SHA-1 that can be used to access the object database. If so, emit it to the standard output; otherwise, error out.

           If you want to make sure that the output actually names an object in your object database and/or can be used as a specific type of object you require, you can add the ^{type} peeling operator to the parameter. For example, git rev-parse "$VAR^{commit}"
           will make sure $VAR names an existing object that is a commit-ish (i.e. a commit, or an annotated tag that points at a commit). To make sure that $VAR names an existing object of any type, git rev-parse "$VAR^{object}" can be used.
```

(That job failed because of a legit commit problem though, right?)

tchaikov: @djgalloway any update? there are more and more test failures caused by this issue. see also #36544

PR #1679, opened 2020-10-06
ceph-pr-commits: Stop using junit plugin

djgalloway:
It's not really useful for this mostly inconsequential job and it's causing job failures.


tchaikov:
#1680

PR #1681, opened 2020-10-06
jjb: Unpin jjb version

djgalloway:
Pinning it keeps biting us more often than fixing/updating job yamls would.


epuertat (approved):
Thanks!

ktdreyer:
cheering for unpinning

PR #1686, opened 2020-10-16
ceph-pr-commits: use origin/master as the base commit

tchaikov:
instead of using `${sha1}^1...${sha1}^2` use
`origin/master...origin/pr/${ghprbPullId}/head` for getting the range of
commits introduced by a PR. as per `man git-diff`

> "git diff A...B" is equivalent to "git diff $(git-merge-base A B) B".

and `origin/master` and `origin/pr/${ghprbPullId}/head` are fetched by
the Git plugin, so it's safe to use them.

### DIFF
--- a/ceph-pr-commits/build/build
+++ b/ceph-pr-commits/build/build
@@ -6,16 +6,14 @@
 # this job/check is required, it hung with 'Expected — Waiting for status to be reported'
 pushd .
 cd "$WORKSPACE/ceph"
-if git rev-parse --verify ${sha1}^2; then
-     files="$(git diff --name-only ${sha1}^1...${sha1}^2)"
-     echo -e "changed files:\n$files"
-     if [ $(echo "$files" | grep -v '^doc/' | wc -l) -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


tchaikov:
```
+ cd /home/jenkins-build/build/workspace/ceph-pr-commits/ceph
++ git diff --name-only origin/master...origin/pr/37687/head
+ files='src/crimson/common/errorator.h
src/test/crimson/CMakeLists.txt
src/test/crimson/test_errorator.cc'
+ echo -e 'changed files:\nsrc/crimson/common/errorator.h
src/test/crimson/CMakeLists.txt
src/test/crimson/test_errorator.cc'
changed files:
src/crimson/common/errorator.h
src/test/crimson/CMakeLists.txt
src/test/crimson/test_errorator.cc
```

djgalloway:
What if the target branch isn't `master` though?

djgalloway:
I think we should use `$ghprbTargetBranch` instead of `origin/master`

tchaikov:
> ghprbTargetBranch

@djgalloway fixed and repushed.

PR #1693, opened 2020-10-27
ceph-container-flake8: add selinux flag to volume

dsavineau:
Without the selinux flag on the container volume, the flake8 command
fails with:

```console
E902 IOError: [Errno 13] Permission denied: '/envglobals.py'
```


dsavineau:
This is required for sloppy python job to passe on https://github.com/ceph/ceph-container/pull/1706

PR #1696, opened 2020-11-11
build_utils.sh: add centos8 storage sig repo

yuvalif:
to support luarocks package on ceph

Signed-off-by: Yuval Lifshitz <ylifshit@redhat.com>


yuvalif:
not needed anymore. since luarocks was moved to the lab-extras repo which his already set

PR #1697, opened 2020-11-11
ceph-dev*build: Pull centos images from our registry mirror

djgalloway:
Since docker started rate limiting on Nov2, we've often seen errors that we hit our `pull` limit.  This should greatly reduce the number of times we're going to docker.io.  Only ephemeral builders would pull from docker.io once this is merged and those builders' IP changes when they're spun up so we don't really need to worry about maxing out that IP's `pull` limit.

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -24,6 +24,7 @@
     - grant_sudo: true
     - osc_user: 'username'
     - osc_pass: 'password'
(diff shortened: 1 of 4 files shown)


djgalloway:
This change relies on https://github.com/ceph/ceph-container/pull/1752

djgalloway:
Closing in favor of https://github.com/ceph/ceph-build/pull/1698

PR #1698, opened 2020-11-12
Use internal docker registry mirror

djgalloway:


dmick:
While I believe this probably works, it's distasteful to add a self-signed cert to the root cert trust db.  I'm not sure of the best alternative, and maybe we could do this temporarily until we find a better way.

dmick:
I assume you've seen this:  https://docs.docker.com/registry/insecure/#use-self-signed-certificates

djgalloway:
> While I believe this probably works, it's distasteful to add a self-signed cert to the root cert trust db. I'm not sure of the best alternative, and maybe we could do this temporarily until we find a better way.

Guess I needed to sleep on it to get the proper way working.  It does now.

scripts/build_utils.sh
```diff
@@ -1287,3 +1287,12 @@ EOF
                  ${build_area}/SPECS/ceph-release.spec
     fi
 }
+
+# dockerhub started aggressively rate limiting in November 2020. We can use an internal docker mirror if the builder is in the upstream lab.
+use_internal_container_registry() {
+  if ping -c 1 docker-mirror.front.sepia.ceph.com; then
```
dmick, line 1293: It might be more reliable and a better test of an available service to actually http-transfer something, like with curl, and maybe a timeout.  I think ping can fail for silly reasons sometimes.

djgalloway: Good point.  Updated.

dmick (approved):
LGTM.

PR #1701, opened 2020-11-12
ceph-dev-new-trigger: disable leap15 for nautilus

kshtsk:
We do not need for now opensuse-15.1 builds, since
support of it is limited.

The current version of Leap is 15.2 and the 15.1
is going to be EOL, thus we are not going to spend
resources on it.

We can disable it back if required in future.

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -79,7 +79,7 @@
                     DISTROS=centos7
                     FLAVOR=notcmalloc
       # build nautilus on:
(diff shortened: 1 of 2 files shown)


smithfarm (approved):

djgalloway:
Here too: https://github.com/ceph/ceph-build/blob/master/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml#L108

kshtsk:
> Here too: https://github.com/ceph/ceph-build/blob/master/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml#L108

aye-aye

kshtsk:
@djgalloway ping

PR #1707, opened 2020-11-16
ceph-pr,api-pr: skip when not needed 2/2

epuertat:
...now with proper syntax.

Proof:
`jenkins-jobs --ignore-cache --conf ~/.jenkins_jobs.jenkins.ceph.com.ini update ceph-pull-requests/config/definitions/ceph-pull-requests.yml`

image

Let's see if the regex patterns work as expected!

Fixes: https://github.com/ceph/ceph-build/pull/1705
Signed-off-by: Ernesto Puerta <epuertat@redhat.com>

PR #1708, opened 2020-11-17
ceph-pr,api-pr: Revert" skip when not needed"

tchaikov:
This reverts commit ae9a52ba2faa0b7faf0b9e0c6d3fd289cdd576ea.

### DIFF
--- a/ceph-pr-api/config/definitions/ceph-pr-api.yml
+++ b/ceph-pr-api/config/definitions/ceph-pr-api.yml
@@ -48,25 +48,6 @@
           started-status: "running API tests"
           success-status: "ceph API tests succeeded"
           failure-status: "ceph API tests failed"
-          exclude-region:
-            ^.github/
-            ^admin/
-            ^ceph-menv
-            ^doc/
-            ^etc/
-            ^examples/
-            ^fusetrace/
-            ^keys/
-            ^man/
-            ^mirroring/
-            ^selinux/
-            ^share/
-            ^sudoers.d/
-            ^systemd/
-            ^udev/
-            \.rst$
-            \.txt$
-            \.md$
 
     scm:
       - git:
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -45,26 +45,6 @@
           started-status: "running make check"
           success-status: "make check succeeded"
           failure-status: "make check failed"
-          exclude-region:
-            ^.github/
-            ^admin/
-            ^ceph-menv
-            ^doc/
-            ^etc/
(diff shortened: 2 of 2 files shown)


callithea (approved):

epuertat (approved):
Thank you @tchaikov! It's a pity that this didn't work :(

djgalloway:
@epuertat did you double check that this got templated out correctly in the Jenkins Web UI?  The Help for the plugin says "Each exclusion uses regular expression pattern matching, and must be separated by a new line." which I see you wrote here in the yaml.  I'm just curious if you needed to

```
do: |
  this
```

epuertat:
@djgalloway, yup. I fixed some syntax issues from the 1st PR in the 2nd one (https://github.com/ceph/ceph-build/pull/1707), which resulted in the following UI config (I also tried with just `doc/`):

image

...but even though, it didn't catch a basic case (doc only change):

- https://github.com/ceph/ceph/pull/38115
- https://github.com/ceph/ceph/pull/38116
- https://github.com/ceph/ceph/pull/38117

I wish you detect any omission over there, but after Googling a bit, it seems that the issue has to do with the way the Jenkins `git-client` plugins provides the changeset list with a shallow clone (I also tried with a regular clone and with an incremental checkout, but no better luck).

djgalloway:
Did you try `^doc/.*` ?

epuertat:
> Did you try `^doc/.*` ?

No, I tried with both `^doc/` (a [Java regexp](https://www.freeformatter.com/java-regex-tester.html) should still match `doc/etc`) and just `doc`... Do you think that `^doc/.*` will fix this?

djgalloway:
> > Did you try `^doc/.*` ?
> 
> Do you think that `^doc/.*` will fix this?

I think it's possible.  All of the examples they provide (except the single file `.gitignore`) have wildcards in them.  I wonder if the plugin is being super strict and looking for **only** the doc dir itself being changed (even though that's not a thing that would happen).

Here's the code: https://github.com/jenkinsci/ghprb-plugin/pull/293/files

epuertat:
You were right! I tried with the following patterns and some of them worked:

image

The funny thing is that, the check doesn't run but Github still considers that as required, so the situation is rather weird:

image

I guess the only way to overcome this is that a single/required Jenkins pipeline check is required and that goes through a variable number of stages (sub-checks)....

djgalloway:
> The funny thing is that, the check doesn't run but Github still considers that as required, so the situation is rather weird:

Ah, that's right.  I ran into this with the [ceph-pr-commits job](https://github.com/ceph/ceph-build/pull/1671).  In that instance, the job still runs but immediately "passes" If it's a docs only change.  That's not gonna work with the ghprb plugin.

PR #1709, opened 2020-11-17
ceph*build: Login to docker.io if needed

djgalloway:
We ran into an issue today where the internal mirror was temporarily unavailable.  The job fell back to docker.io and despite the fact that we're not using it all that much, we still got rate limited.  I suspect other tenants in the Community Cage are still pulling from docker.io anonymously.

This PR will have us login (with a higher rate-limit).


dmick (approved):
It's possible this will engender some surprise in ceph-container to find that login has been addressed elsewhere, but division of labor between ceph-build and ceph-container is already muddy, and this is probably as clean as it can be made.

PR #1717, opened 2020-12-03
ansible: Disable daily apt transaction on Ephemeral slaves

djgalloway:

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -330,6 +330,24 @@
             line: '127.0.1.1 ceph-builders'
           failed_when: false
           when: ansible_os_family == "Suse"
+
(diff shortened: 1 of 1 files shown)


smithfarm (approved):
having test slaves update themselves automatically on a daily basis seems like a recipe for trouble, so yes. if this change disables that, I'm for it.

smithfarm:
@djgalloway Thanks for taking care.

djgalloway:
> having test slaves update themselves automatically on a daily basis seems like a recipe for trouble, so yes. if this change disables that, I'm for it.

Yep, that's what was happening in the jobs you pinged me about.  The slave would come up, connect to Jenkins, then `unattended-upgrades` or the systemd timer would start an `apt` transaction which interrupted the Jenkins job.  This PR stops that.

tchaikov:
@djgalloway thanks for looking into this. i had been wondering who's using dpkg / apt while the jenkins slave was running test..

PR #1719, opened 2020-12-07
quay-pruner.py: find orphaned SHA1-based tags and delete

dmick:
There's also some small cleanups/debug helps.  I've run this a bunch in dryrun mode and once for real, and it seems like it's doing what's desired, but I sure wouldn't mind some review.


dmick:
@dsavineau @djgalloway please take a look and suggest to anyone else who might be a good reviewer

djgalloway:
This is a pretty inconsequential job.  I'm fine with merging and cleaning up after-the-fact if any is necessary.

PR #1722, opened 2021-01-06
ceph-dev-new: add Windows build

petrutlucian94:
ceph-dev-new: add Windows build

We're going to do a Windows build using Mingw as part of the
ceph-dev-new job.

"windows" is added as a valid distro. The "validate_mingw" script
will ensure that we're running on top of a Linux distro that the
Windows build script supports (currently just Ubuntu and Suse
distros).

### DIFF
--- a/ceph-dev-build/build/validate_rpm
+++ b/ceph-dev-build/build/validate_rpm
@@ -11,3 +11,7 @@ set -ex
       exit 1
       ;;
   esac) || exit 0
+
+if [ "${DIST}" == "windows" ]; then
+  exit 0
+fi
--- a/ceph-dev-new-build/build/validate_mingw
+++ b/ceph-dev-new-build/build/validate_mingw
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+# At the moment, we only support cross compiling Windows binaries using
+# Ubuntu or SUSE distros.
+( source /etc/os-release
+  case $ID in
+    opensuse*|suse|sles)
+      exit 0
+      ;;
+    ubuntu)
+      exit 0
+      ;;
+    *)
+      exit 1
+      ;;
+  esac) || exit 0
+
+if [ "${DIST}" != "windows" ]; then
+  exit 0
+fi
--- a/ceph-dev-new-build/config/definitions/ceph-dev-new-build.yml
+++ b/ceph-dev-new-build/config/definitions/ceph-dev-new-build.yml
@@ -54,6 +54,7 @@
             - centos7
             - centos8
             - leap15
+            - windows
       - axis:
           type: dynamic
           name: DIST
@@ -105,6 +106,13 @@
             - ../../../scripts/build_utils.sh
             - ../../build/setup_osc
             - ../../build/build_osc
+      # mingw build scripts (targeting Windows)
+      - shell:
+          !include-raw:
+            - ../../build/validate_mingw
+            - ../../../scripts/build_utils.sh
+            - ../../build/setup_mingw
+            - ../../build/build_mingw
 
     publishers:
       - postbuildscript:
--- a/ceph-dev-build/build/validate_deb
+++ b/ceph-dev-build/build/validate_deb
@@ -11,3 +11,7 @@ set -ex
       exit 1
       ;;
   esac) || exit 0
+
+if [ "${DIST}" == "windows" ]; then
+  exit 0
+fi
--- a/ceph-dev-build/build/validate_osc
+++ b/ceph-dev-build/build/validate_osc
@@ -11,3 +11,7 @@ set -ex
       exit 1
       ;;
   esac) || exit 0
+
+if [ "${DIST}" == "windows" ]; then
+  exit 0
+fi
--- a/ceph-dev-new-build/build/build_mingw
+++ b/ceph-dev-new-build/build/build_mingw
@@ -1,0 +1,39 @@
+#!/bin/bash
+set -ex
+
+timeout 3h CMAKE_BUILD_TYPE=Release BUILD_ZIP=1 CLEAN_BUILD=1 ./win32_build.sh
+
+if [ "$THROWAWAY" = false ]; then
+    # push binaries to chacra
+    chacra_binary="$VENV/chacractl binary"
+    chacra_create="$chacra_binary create"
+    [ "$FORCE" = true ] && chacra_binary="$chacra_binary --force"
+
+    find build -name "*.zip" |
+        $chacra_create ${chacra_endpoint}/source/flavors/${FLAVOR}
+
+    # write json file with build info
+    cat > $WORKSPACE/repo-extra.json << EOF
+{
+    "version":"$vers",
+    "package_manager_version":"",
+    "build_url":"$BUILD_URL",
+    "root_build_cause":"$ROOT_BUILD_CAUSE",
+    "node_name":"$NODE_NAME",
+    "job_name":"$JOB_NAME"
+}
+EOF
+    # post the json to repo-extra json to chacra
+    curl -X POST \
+         -H "Content-Type:application/json" \
+         --data "@$WORKSPACE/repo-extra.json" \
+         -u $CHACRACTL_USER:$CHACRACTL_KEY \
+         ${chacra_url}repos/${chacra_repo_endpoint}/extra/
+    # start repo creation
+    $VENV/chacractl repo update ${chacra_repo_endpoint}
+
+    echo Check the status of the repo at: https://shaman.ceph.com/api/repos/${chacra_repo_endpoint}/
+fi
+
+# update shaman with the completed build status
+update_build_status "completed" "ceph" $NORMAL_DISTRO $NORMAL_DISTRO_VERSION $NORMAL_ARCH
--- a/ceph-dev-new-build/build/setup_mingw
+++ b/ceph-dev-new-build/build/setup_mingw
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+set -ex
+HOST=$(hostname --short)
+echo "Building on $(hostname)"
+echo "  DIST=${DIST}"
+echo "  BPTAG=${BPTAG}"
+echo "  KEYID=${KEYID}"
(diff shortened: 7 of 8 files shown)


jan--f:
This is related to the windows related changes recently added. https://github.com/ceph/ceph/pulls?q=is%3Apr+is%3Aopen+label%3Awin32+

djgalloway:
We don't have any plans to build or test on Windows in our CI as far as I know.

petrutlucian94:
> We don't have any plans to build or test on Windows in our CI as far as I know.

Well, this is exactly what Jason suggested here: https://github.com/ceph/ceph/pull/34859#discussion_r419747584

We are just cross compiling Ceph for Windows using Mingw on Linux. We need at least some minimal testing in order to ensure that at the build still works. In fact, it was already broken by a few recent commits.

djgalloway:
@dillaman this would double (at least?) the amount of time required for `make check` to finish on every PR.  Seems like something that should at least go through CLT.

petrutlucian94:
> @dillaman this would double (at least?) the amount of time required for `make check` to finish on every PR.

Not necessarily. We're only building a fraction of the Ceph project, the client side. It's true that ATM it takes a bit longer than it should due to the fact that we're picking make targets in order to avoid complicating the cmake scripts: https://github.com/ceph/ceph/blob/master/win32_build.sh#L161-L177.  Using ninja instead will help, we dropped it for now since a few components (zlib related IIRC) could not be built using ninja.

Another idea would be to run the jobs in parallel, maybe as separate jobs.

dillaman:
For clarity, I wasn't suggesting it for the per-PR job, I was suggesting it for the standard "ceph-ci" build job where we already build 6 different releases per branch.

petrutlucian94:
> For clarity, I wasn't suggesting it for the per-PR job, I was suggesting it for the standard "ceph-ci" build job where we already build 6 different releases per branch.

I may have misunderstood. Sorry, this is quite new to me. So I assume that a ceph-ci job would be triggered on demand by the QA team against certain branches. Would that happen to all PRs before getting merged?

petrutlucian94:
@dillaman awesome, I'll move the windows build there. Thanks!

dillaman:
@djgalloway Any objections to merging this? It would be great to get the Windows builder. We can merge this as a test and then tweak the trigger to enable the new distro for pacific and master in a later PR once it's verified to be working.

dillaman:
@petrutlucian94 Can you please rebase this?

dillaman (approved):
lgtm

ceph-dev-new-build/config/definitions/ceph-dev-new-build.yml
```diff
@@ -105,6 +106,13 @@
             - ../../../scripts/build_utils.sh
             - ../../build/setup_osc
             - ../../build/build_osc
+      # mingw build scripts (targeting Windows)
+      - shell:
+          !include-raw:
+            - ../../build/validate_mingw
```
dillaman, line 112: Actually, if I am understanding the jenkins build process correctly, these scripts are executed on all builders for all distro types. Normally, the `validate_XYZ` scripts will `exit 1` to indicate that it shouldn't build on this host. Therefore, it seems like on an Ubuntu host, for example, it will first run the `build_deb` script above and then come down here to the `build_mingw` script. Would the `validate_XYZ` scripts need to be tweaked to also ensure that when `DIST == windows`, it's not building DEBs and when `DIST != windows` it's not building Windows here?

ceph-build/build/build_mingw
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
dillaman, line 1: Should these files be in the `ceph-dev-new-build` directory instead?

djgalloway: Yes.  The `ceph-build` directory is for upstream releases.

ceph-dev-new-build/config/definitions/ceph-dev-new-build.yml
djgalloway: Yep, the `DIST` variable should also be evaluated.

djgalloway: `validate_deb` should also be modified

ceph-build/build/setup_mingw
```diff
@@ -0,0 +1,62 @@
...
+    SUDO=sudo
+fi
+export LC_ALL=C # the following is vulnerable to i18n
+
+BRANCH=`branch_slash_filter $BRANCH`
+
+cd $WORKSPACE
+
+vers=$(cat ./dist/version)
+raw_version=$(echo $vers | cut -d '-' -f 1)
```
dillaman, line 26: Nit: you might be better off just taking the major so that you can compare that it's >= 16 -- otherwise this check will fail on the master branch since it's now quincy.

ceph-dev-new-build/config/definitions/ceph-dev-new-build.yml
dillaman: @djgalloway I modified and force-pushed. Look good to you?

petrutlucian94:
@dillaman thanks a lot for updating this PR! sorry for the late reply, I'm on GMT+2.

ceph-dev-build/build/validate_rpm
```diff
@@ -11,3 +11,7 @@ set -ex
       exit 1
       ;;
   esac) || exit 0
+
+if [ "${DIST}" == "windows" ]; then
+  exit 0
```
petrutlucian94, line 16: I think it's supposed to return 0 for the targeted job type and 1 otherwise. Wondering if we should return 1 and maybe move this check a bit higher.

ceph-dev-new-build/build/validate_mingw
```diff
@@ -0,0 +1,21 @@
...
+    ubuntu)
+      exit 0
+      ;;
+    *)
+      exit 1
+      ;;
+  esac) || exit 0
+
+if [ "${DIST}" != "windows" ]; then
+  exit 0
```
petrutlucian94, line 20: I think we're supposed to return 1 for other DIST values. I can update those checks but I'll wait for confirmation.

ceph-dev-build/build/validate_rpm
djgalloway: That `if` statement is never going to run because the one above it will either `exit 1` or `0` before it gets a chance.

May I suggest

1. Create a function in `scripts/build_utils.sh`.  Maybe call it `is_windows()` and put that `if` statement in it to return `true` or `false`.  Or whatever a bashier way is.
2. Run the `is_windows` function before the `exit` in each Ubuntu and Windows case.  (e.g., [here](https://github.com/ceph/ceph-build/blob/master/ceph-dev-build/build/validate_deb#L7))

djgalloway: Example `validate_deb`

```
#!/bin/bash
set -ex

# Only do actual work when we are a DEB distro
( source /etc/os-release
  case $ID in
    debian|ubuntu)
      if is_windows; then
        exit 1
      fi
      exit 0
      ;;
    *)
      exit 1
      ;;
  esac) || exit 0
  ```

Then

```
is_windows() {
  if [ "${DIST}" == "windows" ]; then
    return
  fi
  false
}
```

ceph-dev-new-build/build/validate_mingw
dillaman: If you exit w/ `1`, it will fail the jenkins builder when in actuality all we really want to do is stop the processing of the shell script (where the validate/build/etc) scripts are merged together and run as a single unit.

djgalloway: Yeah, actually, this is correct.  So maybe my [approach](https://github.com/ceph/ceph-build/pull/1722#discussion_r582896553) is also wrong still.

ceph-dev-build/build/validate_rpm
dillaman: Actually, the script will only exit if that `case` statement exits w/ return code `1`. The code is very confusing at initial glance but it's relying on the fact that the `case` statement within the parathesis is run as a separate script, so the `exit` calls only affect that sub-script. Then if it exits with 0 (i.e. continue), we don't hit the `||` case and continue. If it exits with 1 (i.e. stop the build), it hits the `||` and exits the build cleanly.

ceph-dev-new-build/build/validate_mingw
dillaman: Right, the `validate_XYZ` scripts should exit (cleanly) if we are on the wrong DEB vs RPM but then also exit cleanly if we don't match on the `windows` check.

ceph-dev-build/build/validate_rpm
dillaman: > Create a function in scripts/build_utils.sh. Maybe call it is_windows() and put that if statement in it to return true or false. Or whatever a bashier way is.

Right now the `validate_XYZ` scripts are invoked before the `build_utils.sh` are pulled in. I figured a repeated three line `if` statement was better than totally restructuring at this point.

djgalloway: Fair enough.  Sounds like we need to rethink the validation scripts anyway

dillaman:
@djgalloway How would we go about getting Bionic builder nodes labeled w/ `windows` so that Jenkins can use them to build? I kicked off a manual build for windows and it's stuck waiting for a node to become available with the labels `‘gigantic&&x86_64&&windows` [1]

[1] https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=windows,DIST=windows,MACHINE_SIZE=gigantic/

djgalloway:
> @djgalloway How would we go about getting Bionic builder nodes labeled w/ `windows` so that Jenkins can use them to build? 

I just added a `windows` label to all the braggi running Bionic.

dillaman:
@djgalloway Thanks -- I'll kick off a test in a bit.

PR #1732, opened 2021-01-27
disable leap15 for master and pacific

kshtsk:
SUSE is not going to support ceph except of octopus anymore

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -147,7 +147,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8 leap15
+                    DISTROS=focal bionic centos8
                 - project: 'ceph-dev-new'
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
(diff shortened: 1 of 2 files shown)


smithfarm (approved):

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -177,7 +177,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8 leap15
```
tchaikov, line 180: this also matches with "luminous\|mimic\|nautilus\|octopus\" in addition to pacific. are you sure we don't need to build these branches for leap?

ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
```diff
@@ -189,7 +189,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8 leap15
```
tchaikov, line 192: ack.

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
kshtsk: @tchaikov according to condition it matches master and anything else but not `luminous|mimic|nautilus|octopus|pacific`.
```
          condition-command: |
            echo "${GIT_BRANCH}" | grep -v '\(luminous\|mimic\|nautilus\|octopus\|pacific\)'
```

tchaikov: ahh, right! I missed `-v`.

tchaikov (approved):

PR #1734, opened 2021-02-02
Fix prune-quay.py's deletion of busy (well-known) refs

dmick:


dmick:
Marked DNM to allow a full test once a busy ref is rebuilt

dmick:
@liewegas I think this is the reason well-known refs were disappearing from quay.  I want to test it once there have been at least two rebuilds of a well-known tag.

dmick:
It seems to be doing the right thing with master now in my tests, and the logging is easier to read and understand.  

The key thing was that while there are three tags per ref (maybe 4, if -crimson is created), when the same busy ref is repushed, the ref tag gets overwritten, while the other two remain.  So master v1 is gone, replaced by master v2, but the master v1 sha1 and ref-shortsha1-el tags remain, and should be pruned.  Just not the new (v2) 'master' tag.  (or octopus, or pacific, or wip-sage-testing, or whatever.)

dmick:
This looks right, and I've just been tweaking the logging.

I've disabled the Jenkins job; if you merge this, please reenable the job; @djgalloway will know how

djgalloway (approved):

PR #1736, opened 2021-02-03
build: notcmalloc builds enable WITH_BOOST_VALGRIND

cbodley:
this flag silences a false-positive from valgrind in boost::context:
```
<error>
  <unique>0x0</unique>
  <tid>124</tid>
  <kind>InvalidRead</kind>
  <what>Invalid read of size 8</what>
  <stack>
    <frame>
      <ip>0x52FF7E7</ip>
      <obj>/usr/lib64/libradosgw.so.2.0.0</obj>
      <fn>spawn::detail::continuation_context::resume()</fn>
    </frame>
  </stack>
  <auxwhat>Address 0x57f65728 is in a rw- anonymous segment</auxwhat>
</error>
```

Fixes: https://tracker.ceph.com/issues/48963

### DIFF
--- a/ceph-dev-new-setup/build/build
+++ b/ceph-dev-new-setup/build/build
@@ -70,7 +70,7 @@ case "${FLAVOR}" in
     notcmalloc)
         echo "Detected notcmalloc flavor: will use flag: -DALLOCATOR=libc"
         CEPH_EXTRA_RPMBUILD_ARGS="--without tcmalloc"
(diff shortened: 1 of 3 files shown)


tchaikov (approved):

cbodley:
thank you @tchaikov!

cbodley:
fyi, i'm seeing radosgw segfaults from these builds. hopefully it doesn't affect other targets

https://github.com/ceph/ceph/pull/39263#issuecomment-772724899

cbodley:
ok, the segfault is just an issue with the radosgw's cmake. the radosgw target didn't get the `BOOST_USE_VALGRIND` compile definition attached to the boost::context target. see https://github.com/ceph/ceph/pull/39279

PR #1743, opened 2021-02-16
Switch to building ceph.git branches twice daily

djgalloway:


dmick (approved):
LGTM.

ktdreyer:
Can we rename this to "timed" instead of "nightly" since it's happening multiple times in 24 hours?

PR #1745, opened 2021-02-18
ansible: Install tmpreaper/tmpwatch and clean /tmp daily

djgalloway:

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -55,6 +55,8 @@
           - doxygen
           - ditaa
           - ant
+          - tmpreaper
+        tmp_cleaner_name: tmpreaper
       when: ansible_os_family == "Debian"
 
     # Libvirt DEBs (Bionic and older)
@@ -145,6 +147,8 @@
           - rpmdevtools
           - openssl-devel
           - libffi-devel
+          - tmpwatch
+        tmp_cleaner_name: tmpwatch
       when: ansible_os_family == "RedHat"
 
     # Libvirt RPMs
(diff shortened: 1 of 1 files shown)


ansible/examples/slave.yml
```diff
@@ -627,6 +631,29 @@
             line: '%dist .el{{ ansible_distribution_major_version }}'
       when: ansible_os_family == "RedHat"
 
+    ## tmpwatch/tmpreaper TASKS
+    - name: tmpwatch/tmpreaper Tasks
+      block:
+        # In case we're running 'ansible-playbook --tags tmp'
```
dmick, line 637: This is the thing that installs the package, tag or not, right?

ansible/examples/slave.yml
```diff
@@ -627,6 +631,29 @@
...
+          file:
+            path: /etc/cron.daily/tmpreaper
+            state: absent
+          when: ansible_os_family == "Debian"
+
+        - name: Create tmp cleaning cronjob
+          cron:
+            name: "Clean /tmp"
+            special_time: daily
+            job: "{{ tmp_cleaner_name }} 30d /tmp/"
```
dmick, line 653: Does this mean run daily and remove anything that's 30 days old or older?  If so a comment wouldn't be out of place for obviousness without having to look up the utility

dmick (approved):

ansible/examples/slave.yml
djgalloway: This installs the package in case

1. We're not running the whole playbook
2. The package is not installed and needs to be

ansible/examples/slave.yml
djgalloway: Fair enough

ansible/examples/slave.yml
djgalloway: The package would get installed during `## PACKAGE INSTALLATION TASKS` but not if we're running `ansible-playbook slave.yml --tags tmp`

PR #1754, opened 2021-02-25
ceph-dev-new-trigger: build windows for pacific and master

dillaman:

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -118,7 +118,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8
+                    DISTROS=focal bionic centos8 windows
                 - project: 'ceph-dev-new'
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
@@ -143,7 +143,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8
+                    DISTROS=focal bionic centos8 windows
             - trigger-builds:
                 - project: 'ceph-dev-new'
                   predefined-parameters: |


dillaman:
Merge after the link error is resolved in the windows build [1]

[1] https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=windows,DIST=windows,MACHINE_SIZE=gigantic/52239/consoleFull

dillaman:
@petrutlucian94 FYI

dillaman:
It looks like `mingw-w64` project added the `GetOverlappedResultEx` function to `kernel32.lib` here [1], which landed in release v7. Unfortunately, our builders are based on Ubuntu Bionic which only includes `mingw-w64` v5.0.3 [2]. It is available in Ubuntu Focus since it includes `mingw-w64 v7.0.0` [3] so perhaps we can just update the build script to add the Focus repo to ensure the most up-to-date version is installed, or that recent change to start using `GetOverlappedResultEx` can be tweaked to remove it or just use `GetOverlappedResult` (if possible)

[1] https://github.com/mirror/mingw-w64/commit/24842d45e025db0d38fa2bbd932b95a83282efa2
[2] https://packages.ubuntu.com/bionic/mingw-w64-x86-64-dev
[3] https://packages.ubuntu.com/bionic/mingw-w64-x86-64-dev

petrutlucian94:
@dillaman thanks for looking into it. In this case, I think we can safely use ``GetOverlappedResult`` since we're getting the result of an async operation that's expected to be completed, so we don't really need the timeout parameter or alertable wait flag: https://github.com/ceph/ceph/blob/master/src/tools/rbd_wnbd/rbd_wnbd.cc#L402-L405

One issue with ``mingw-w64`` v5.0.3 is that the runtime is broken (fixed in v7.0.0). For example, the random number generator returns the same sequence, so we had separate processes using the same connection IDs :). That's ok as long as we only care about checking if it still compiles, but the resulting binaries are unusable. I think I'll have to include this in the Windows readme.

I'll also check if we can get the latest mingw version to work on Ubuntu Bionic.

petrutlucian94:
I've submitted a simple PR, switching to ``GetOverlappedResult``: https://github.com/ceph/ceph/pull/39707. I gave it a quick try as well, works as expected.

dillaman:
@petrutlucian94 The builders use `pbuilder` on bionic when building the different distros (xenial, bionic, focal, etc). If we cannot build on bionic, I think then the next available solution would be to tweak the builder script to execute inside a `pbuilder` environment.

dillaman:
The builder is passing now :partying_face: [1]. Once the builds are usable, though, we can merge this and then tie it into the master and pacific branch builders.

[1] https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=windows,DIST=windows,MACHINE_SIZE=gigantic/52253//consoleFull

dillaman:
... need to investigate why chacra isn't showing the `ceph.zip` artifact in the "repo", though [1]

https://2.chacra.ceph.com/r/ceph/wip-jd-testing/9e88f36e5c97454a70221e74f1c942dded67fa73/windows/1809/flavors/default/

dillaman:
> ... need to investigate why chacra isn't showing the `ceph.zip` artifact in the "repo", though [1]
> 
> https://2.chacra.ceph.com/r/ceph/wip-jd-testing/9e88f36e5c97454a70221e74f1c942dded67fa73/windows/1809/flavors/default/

Now tracked here [1]

[1] https://github.com/ceph/chacra/issues/284

petrutlucian94:
I've submitted a PR for using Ubuntu Focal through pbuilder: https://github.com/ceph/ceph-build/pull/1755

dillaman:
Builds appear to be successful under pbuilder [1]

[1] https://shaman.ceph.com/builds/ceph/wip-jd-testing/1f19e92c2c4adb936346f64684d80521e47a5ec6/default/258223/

PR #1755, opened 2021-03-02
ceph-dev-new-build: cross-compile Windows targets using pbuilder

petrutlucian94:
The Windows builds that are generated right now are unusable due to
known Mingw runtime bugs.

We need Ubuntu Focal or later, for which reason we're going to use
pbuilder.

### DIFF
--- a/ceph-dev-new-build/build/build_mingw
+++ b/ceph-dev-new-build/build/build_mingw
@@ -1,7 +1,26 @@
 #!/bin/bash
 set -ex
 
+# We need Ubuntu Focal to cross-compile Ceph for Windows.
+# Previous versions provide broken Mingw packages.
+# "DIST" will be set to "windows", so we're currently overriding it with
+# a hardcoded value.
+tmp_pbuild_script=$(mktemp /tmp/build_mingw_pbuild.XXXXXX)
+cat << EOF > $tmp_pbuild_script
+#!/bin/sh
+# Used by the build script
+apt-get install -y sudo git automake wget
+
+cd /mnt/ceph
 CMAKE_BUILD_TYPE=Release BUILD_ZIP=1 CLEAN_BUILD=1 timeout 3h ./win32_build.sh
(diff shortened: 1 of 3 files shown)


ceph-dev-new-build/build/setup_mingw
```diff
@@ -65,3 +65,11 @@ if [ "$THROWAWAY" = false ] ; then
     # TODO if this exits we need to post to shaman a success
     check_binary_existence $chacra_check_url
 fi
+
+# We need Ubuntu Focal to cross-compile Ceph for Windows.
+# Previous versions provide broken Mingw packages.
+# "DIST" will be set to "windows", so we're currently overriding it with
+# a hardcoded value.
+DIST="focal"
+setup_pbuilder
```
dillaman, line 74: I think you will need to enable networking which is already covered when you pass any value as a parameter to `setup_pbuilder`

```suggestion
setup_pbuilder use_gcc
```

ceph-dev-new-build/build/build_mingw
```diff
@@ -1,7 +1,25 @@
 #!/bin/bash
 set -ex
 
+# We need Ubuntu Focal to cross-compile Ceph for Windows.
+# Previous versions provide broken Mingw packages.
+# "DIST" will be set to "windows", so we're currently overriding it with
+# a hardcoded value.
+tmp_pbuild_script=$(mktemp /tmp/build_mingw_pbuild.XXXXXX)
+cat << EOF > $tmp_pbuild_script
```
dillaman, line 9: Nit: does the script need to be marked executable via `chmod +x` or does `pbuilder execute` not care?

petrutlucian94: I think pbuilder handles that by itself, it worked on my env. I guess we can set it explicitly though.

dillaman (approved):
lgtm

PR #1759, opened 2021-03-08
prune-quay.py: Provision for aarch64 containers, manifest lists, minor improvements

dmick:


djgalloway:
LGTM from a cursory view.  Did you try manually pushing and testing it?

dmick:
I did not, but I ran it (in dry-run mode) multiple times from local, and the results seemed correct

PR #1765, opened 2021-03-17
nfs-ganesha: do not build el7 rpm by default

guits:
building el7 nfs-ganesha rpm for the branch 'next' is not possible at
the moment because it tries to use ceph@master el7 builds.
This is not worth building this package at the moment so let's remove it
from default.

### DIFF
--- a/nfs-ganesha/config/definitions/nfs-ganesha.yml
+++ b/nfs-ganesha/config/definitions/nfs-ganesha.yml
@@ -57,8 +57,8 @@
 
       - string:
           name: DISTROS
-          description: "A list of distros to build for. Available options are: bionic, centos7, centos8"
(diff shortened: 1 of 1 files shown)


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -58,7 +58,7 @@
       - string:
           name: DISTROS
           description: "A list of distros to build for. Available options are: bionic, centos7, centos8"
```
dang, line 60: Description should be fixed too

guits: my original intent was more to not build it by default and leave it available if needed although I agree I'm not sure this is relevant

dang (approved):

Issue #1770, opened 2021-03-17
Signed-off-by failing when using python3


djgalloway:
Not sure what's going on here but hoping Kefu can take a look.

```
+ /tmp/venv.DVYjyKoxpV/bin/pip install -U pip
Traceback (most recent call last):
  File "/tmp/venv.DVYjyKoxpV/bin/pip", line 7, in <module>
    from pip._internal.cli.main import main
  File "/tmp/venv.DVYjyKoxpV/lib/python3.5/site-packages/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
                                   ^
SyntaxError: invalid syntax
```

This started after https://github.com/ceph/ceph-build/pull/1766 got merged

djgalloway:
Maybe we may just now be facing https://pip.pypa.io/en/stable/news/#id1 ?

djgalloway:
Confirmed reverting https://github.com/ceph/ceph-build/pull/1766 (manually) fixes it.  I tried setting the env to use `pip3` instead of `pip` if `pip3` was present but that didn't get us any further.

```
++ command -v /tmp/venv.jAyu84ZReu/bin/pip3
+ '[' /tmp/venv.jAyu84ZReu/bin/pip3 ']'
+ PIP=pip3
+ '[' '' == latest ']'
+ [[ -n '' ]]
+ echo 'Installing pip 10.0.0'
Installing pip 10.0.0
+ /tmp/venv.jAyu84ZReu/bin/pip3 install pip3==10.0.0
Traceback (most recent call last):
  File "/tmp/venv.jAyu84ZReu/bin/pip3", line 7, in <module>
    from pip._internal.cli.main import main
  File "/tmp/venv.jAyu84ZReu/lib/python3.5/site-packages/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
                                   ^
SyntaxError: invalid syntax
```

dmick:
Python 3.5.2 is current on Ubuntu 16.04.4 where the failing signed-off tests run, and it does not include support for f-strings.  pip3 apparently now requires at least 3.6.  This is the breaking commit:  https://github.com/pypa/pip/commit/cdcf74fb8eee47ac13f24dc4692c29bc144ea66c

tchaikov:
[...]

PR #1774, opened 2021-03-18
build_utils.sh: Prefer python3 instead of python2 for virtualenvs

djgalloway:
Let's try this again.  I updated the few Xenial builders we had to Bionic so we should get better results now.  See https://github.com/ceph/ceph-build/issues/1770#issuecomment-801559246

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -96,10 +96,12 @@ create_virtualenv () {
     if [ "$(ls -A $path)" ]; then
         echo "Will reuse existing virtual env: $path"
     else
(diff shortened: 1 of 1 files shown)


djgalloway:
jenkins retest this please

dmick:
Verified that install_python_packages is on the host, not inside the pbuilder env (where we may still run xenial for older releases) so this should be fine.

djgalloway:
@tchaikov just a FYI.  We put this back in place.  No more Xenial builders and also needed https://github.com/ceph/ceph-build/pull/1775

PR #1780, opened 2021-03-25
ceph-pull-requests: skip xunit publisher if no test files present

dmick:
otherwise make check, at least, fails

### DIFF
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -95,6 +95,7 @@
           types:
             - ctest:
                 pattern: "build/Testing/**/Test.xml"
+                skip-if-no-test-files: true
     wrappers:
       - ansicolor
       - credentials-binding:


dmick:
@tchaikov fyi

tchaikov:
@dmick thanks. sorry for introducing this issue. I created

- https://github.com/ceph/ceph/pull/40406
- https://github.com/ceph/ceph/pull/40407
- https://github.com/ceph/ceph/pull/40405

to address this. not sure how long it takes to get them merged. =(

jdurgin:
@tchaikov @epuertat can we avoid any jenkins changes over the next few days as we're trying to finalize pacific?

tchaikov:
@jdurgin will stop messing with jenkins before pacific is out. sorry again.

jdurgin:
thanks @tchaikov, no worries. maybe in the future we can find a way to use a test jenkins of some sort, so these sorts of things can be developed without impacting developer workflows

tchaikov:
@jdurgin the change was tested. and i created these PRs yesterday. but i failed to get them merged in time.

dmick:
Any reason not to merge this anyway?  Seems like a reasonable safety measure.

tchaikov:
@dmick if the test result cannot be found, I'd consider it a signal of test failure or something unexpected happened.

tchaikov:
see https://jenkins.ceph.com/job/ceph-pull-requests/72382/console for an example.

djgalloway:
> @dmick if the test result cannot be found, I'd consider it a signal of test failure or something unexpected happened.

The test result isn't being found because no tests get run.  It's okay that there is no test result.

djgalloway:
> The reason there was no test result was that the install-deps.sh was broken. I’d consider it a bug.

This particular job failed *before* the `No test report file(s) were found with the pattern 'build/Testing/**/Test.xml'` failure.  The job still would've failed even with this PR's `skip-if-no-test-files: true`.

We want `skip-if-no-test-files: true` in cases like https://jenkins.ceph.com/job/ceph-pull-requests/72461/console.  The job saw that only the `doc/` dir changed and `exit 0`ed.  The job passed and was happy.  But then xUnit didn't get its report (because there wasn't one) and made the job fail.  It shouldn't.

```
changed files:
doc/cephadm/rgw.rst
++ echo doc/cephadm/rgw.rst
++ grep -v '^doc/'
++ wc -l
+ '[' 0 -gt 0 ']'
+ DOCS_ONLY=true
+ popd
~/build/workspace/ceph-pull-requests
+ '[' true = true ']'
+ echo 'Only the doc/ dir changed.  No need to run make check.'
Only the doc/ dir changed.  No need to run make check.
+ exit 0
[Cobertura] Publishing Cobertura coverage report...

[Cobertura] No coverage results were found using the pattern 'src/pybind/mgr/dashboard/frontend/coverage/cobertura-coverage.xml' relative to '/home/jenkins-build/build/workspace/ceph-pull-requests'.  Did you enter a pattern relative to the correct directory?  Did you generate the XML report(s) for Cobertura?

[Cobertura] Skipped cobertura reports.

[PostBuildScript] - [INFO] Executing post build scripts.
[PostBuildScript] - [INFO] Build does not have any of the results [ABORTED]. Did not execute build step #0.
INFO: Starting to record.
INFO: Processing CTest-Version 3.x (default)
ERROR: Step ?Publish xUnit test result report? failed: [CTest-Version 3.x (default)] - No test report file(s) were found with the pattern 'build/Testing/**/Test.xml' relative to '/home/jenkins-build/build/workspace/ceph-pull-requests' for the testing framework 'CTest-Version 3.x (default)'.
Did you enter a pattern relative to (and within) the workspace directory?
Did you generate the result report(s) for 'CTest-Version 3.x (default)'?"
```

tchaikov:
@djgalloway yeah, I think I understand the root cause.  but could you please take a look at https://jenkins.ceph.com/job/ceph-pull-requests/72382/console? I also want to fail the test in that case. #1787 is created to address the failure of doc-only tests.

tchaikov (approved):

tchaikov:
sorry, thought that the install-deps.sh could fail without being noticed because of how it is called in `run-make-check.sh`. but I mistook it as `run()` in that very script. as `run()` always returns 0. but I was wrong. because `run-make-check.sh` 

- fails if `prepare()` fails. and `./install-deps.sh` is called by `prepare()`. with this change, publisher won't complain. but the test itself fails.
- succeeds even if `ctest` fails. but in that case, the xUnit XML file is generated. so the publisher will take care of this.

PR #1781, opened 2021-03-27
Fix up nfs-ganesha for arm64/centos8 builds

dmick:


dmick:
@liewegas @djgalloway

cbodley (approved):
looks fine to me if it works

dmick:
@kalebskeithley @mattbenjamin @cbodley can any of you speak to whether we should even be trying to build for centos7?  Ceph master is no longer built for centos7, and I think maybe this job is only really for master, and so it should probably stop trying and failing to do centos7 as well

dmick:
Also, @kalebskeithley @mattbenjamin, any comments, even noncommittal, about these changes, so I can move forward?

cbodley:
> @kalebskeithley @mattbenjamin @cbodley can any of you speak to whether we should even be trying to build for centos7? Ceph master is no longer built for centos7, and I think maybe this job is only really for master, and so it should probably stop trying and failing to do centos7 as well

afaik, rgw isn't using ganesha builds in teuthology at all. i'm guessing these are for cephfs? cc @batrick

kalebskeithley:
ceph-15 and later aren't being built for centos 7 (CentOS Storage SIG) at all, due to python3 issues.

I don't fully understand what's being asked here, but I think I recommend not doing anything with or for centos 7.

dmick:
@kalebskeithley there were two questions; 1) do these changes look okay, as @cbodley requested your review, and 2) should we just stop trying to build for centos7.  I think there's consensus on the latter question, which leaves the former.

Also, bionic builds are broken, with or without this, and I fully don't understand the branching scheme in nfs-ganesha-debian, so will probably need help with that.

batrick (approved):
LGTM

cc @jtlayton @varshar16

kalebskeithley:
@dmick I didn't review the changes as that stuff is meaningless to me. That said, they look fine. ;-)

WRT nfs-ganesha-debian branching scheme:

Ganesha has gone from ganesha-2.X.y to ganesha-3.x, having dropping the teeny version for periodic updates. Until recently the project was actively supporting nfs-ganesha-2.8.4; nfs-ganesha-3.4 is the current, actively maintained version.

The (lib)ntirpc versions that correspond to the respective ganesha versions are ntirpc-1.8.x and ntirpc-3.x. There is no ntirpc-2.x.

Thus, for Ubuntu, the branches that are currently active are {xenial,bionic,focal,groovy,hirsute}-libntirpc-3 and {xenial,bionic,focal,groovy,hirsute}-nfs-ganesha-3. (And you can see the packages that are built from them in Launchpad at [1].

Similarly for Debian they are {stretch,buster,bullseye}-libntirpc-3 and {stretch,buster,bullseye}-nfs-ganesha-3. Packages built from those are at [2].

I'm not a debian package expert — AFAIK, the debian-style package metadata doesn't have any sort of built-in support for conditionals like rpm does, hence the separate branches for each dist. This is a scheme that I inherited from a (presumed expert) debian package maintainer several years ago. If there's a better way though, I'd love to learn about it.

If you have any questions feel free to contact me on IRC or by email. Or here, although I may or may not see things here as quickly as IRC or email.


[1] https://launchpad.net/~nfs-ganesha
[2] https://download.nfs-ganesha.org/3/3.4/Debian/

ajarr:
> cc @ajarr does this affect Manila?

I don't think so

dmick:
Okay, the issues for bionic and centos7 are not resolved, but this should fix centos8.  More to follow.

PR #1786, opened 2021-03-30
ceph-pull-requests: Don't fail the job if there are no test reports

djgalloway:
There won't always be test report files like in the case of a doc-only change.  We `exit 0` those jobs.


ceph-pull-requests/config/definitions/ceph-pull-requests.yml
```diff
@@ -95,6 +95,7 @@
           types:
             - ctest:
                 pattern: "build/Testing/**/Test.xml"
+                skip-if-no-test-files: true
```
tchaikov, line 98: this duplicates #1780. and i think we need a more sophisticated solution.

PR #1792, opened 2021-04-02
scripts/build_utils.sh: install gnupg before ppa packages

tchaikov:
so gnupg can be installed for checking the ppa repo's signatures.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -461,6 +461,48 @@ get_distro_and_target() {
     esac
 }
 
+setup_updates_repo() {
+    local hookdir=$1
+
+    if [[ $DISTRO != ubuntu ]]; then
+        return
+    fi
+    if [[ "$ARCH" == "x86_64" ]]; then
+        cat > $hookdir/D04install-updates-repo <<EOF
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -655,6 +655,13 @@ setup_pbuilder_for_new_gcc() {
     local version=$1
     shift
 
+    cat >> $hookdir/D05install-gcc-7 <<EOF
+echo "deb [lang=none] http://ports.ubuntu.com/ubuntu-ports $DIST-updates main" >> \
```
dmick, line 659: 1) I think this should be http://archive.ubuntu.com/ubuntu/dists/$DIST-updates, no?
2) it ought to go on the end of /etc/apt/sources.list, as it's part of the ubuntu sources

dmick: (at least for x86_64; it's possible arm64 is different)

tchaikov: @dmick fixed and repushed.

tchaikov:
folks, i just deployed the change in this PR to address the FTBFS of nfs-ganesha.

jdurgin (approved):

PR #1795, opened 2021-04-07
scripts/build_utils.sh: set hookdir using pbuilderrc

tchaikov:


tchaikov:
being tested at https://shaman.ceph.com/builds/ceph/nautilus-test-3/0410e7c70d2da1bb52b1d26b6092f9eb52503679/

scripts/build_utils.sh
```diff
@@ -490,6 +490,7 @@ env DEBIAN_FRONTEND=noninteractive apt-get update -y -o Acquire::Languages=none
 env DEBIAN_FRONTEND=noninteractive apt-get install -y gnupg
 EOF
     fi
+    chmox +x $hookdir/D04install-updates-repo
```
djgalloway, line 493: chmox

scripts/build_utils.sh
```diff
@@ -616,7 +613,11 @@ setup_pbuilder() {
         # in newer versions. This ticket solves the specific issue in 8.1.1 (which vendors urllib3):
         # https://github.com/shazow/urllib3/issues/567
         echo "USENETWORK=yes" >> ~/.pbuilderrc
+        local hookdir
+        hookdir=$(recreate_hookdir)
```
djgalloway, line 617: should `$()` be double quoted?  genuine question

djgalloway (changes requested):

scripts/build_utils.sh
tchaikov: @djgalloway fixed and repushed.

scripts/build_utils.sh
tchaikov: no need. but it is advisable to be double quoted when the variable is *referenced* and it should be considered as a single string.

dmick:
1) --hookdir didn't *work*?  2) the reason for doing the updates was internal package inconsistencies, not strictly for ppa, but I'm now thinking that we should not be working from a previously-existing tgz at all, as the hooks are changing it from a "gold" image in ways that we can't control

PR #1798, opened 2021-04-08
nfs-ganesha*: backup golden pbuilder base.tgz before changing it

tchaikov:


badone (approved):
LGTM

alimaredia:
I think this one is up @tserlin 's ally

dmick:
I mean, I think this is a good idea, but I think this ought to be done generically for all the builds that use pbuilder; there may well be lurking inconsistencies and avoidable build failures happening in the various other projects as well (including ceph dev or stable releases).

tchaikov:
@dmick i'd like to help. but can we start with this `nfs-ganesha*`, and only backup/restore if necessary?

dmick:
It seems like doing this generically is even easier, and pretty easy to make safe: build a gold image, update it, and then don't touch it (maybe with periodic time-based updates) for the actual build, but rather always make a copy

tchaikov:
@dmick hi Dan,  is this change still not relevant? if not. i will close it.

dmick:
I cleared all the cached tarballs, and they immediately got polluted on virtually all the build machines.  I'm concerned for all the builds but haven't experienced any actual breakage recently.

tchaikov:
thank you Dan. glad to learn that the builds survived the polluted chroot env =)

PR #1799, opened 2021-04-09
nfs-ganesha: change patch to sed, drop some older distro/releases

dmick:
See commit comments.  I'm willing to be told I'm wrong about removing those distros.


dmick:
https://jenkins.ceph.com/view/dmick/job/nfs-ganesha/1986/console

tserlin (approved):
Thanks, Dan. Looks good to me.

I think it's fine to drop centos7/xenial.

PR #1802, opened 2021-04-15
ceph-pr-api: add WITH_ZBD=ON flag to cmake args

agayev:
Signed-off-by: Abutalib Aghayev <agayev@psu.edu>


tchaikov:
this is the wrong fix. there is not need to enable WITH_ZBD for the api test.

PR #1808, opened 2021-05-04
ceph-pr-docs: add annotation to build result

tchaikov:

### DIFF
--- a/ceph-pr-docs/build/build
+++ b/ceph-pr-docs/build/build
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ex
+set -xo pipefail
 
 # make sure any shaman list file is removed. At some point if all nodes
 # are clean this will not be needed.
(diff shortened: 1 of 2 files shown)


ceph-pr-docs/build/build
```diff
@@ -12,4 +12,34 @@ sudo rm -f /etc/apt/sources.list.d/ceph-boost*
 sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
 sudo apt-get install -y gcc python3-dev python3-pip python3-virtualenv libxml2-dev libxslt-dev doxygen graphviz ant ditaa cython3
 
-timeout 3600 ./admin/build-doc
+virtualenv -q --python python3 venv
+. venv/bin/activate
+pip install tox
+pip install git+https://github.com/ceph/githubcheck.git
+sha1=$(git rev-parse refs/remotes/origin/pr/${ghprbPullId}/head)
```
djgalloway, line 19: Since this is the second time I've seen this block, may I suggest it go in `build_utils.sh` instead?

tchaikov: @djgalloway can we do this in another PR? after a second thought, i think we need to fix https://github.com/ceph/ceph-build/blob/01968d02ab54b07eae57c357362d0b8a2e44e730/scripts/build_utils.sh#L5-L6 first. it *always* creates a temp directory every time we include this script. and it is never removed.

epuertat, line 19: [`GIT_COMMIT`](https://plugins.jenkins.io/git/#environment-variables)?

ceph-pr-docs/build/build
```diff
@@ -12,4 +12,34 @@ sudo rm -f /etc/apt/sources.list.d/ceph-boost*
...
+
+output=$(mktemp $PWD/build-doc-XXX.out)
+
+if timeout 3600 ./admin/build-doc 2>&1 | tee ${output}; then
```
epuertat, line 23: If build-doc fails there, wouldn't `pipefail` flag interrupt the script?

ceph-pr-docs/build/build
tchaikov: i think `GIT_COMMIT` is the *merge* commit, not the one of the sha1 of the change itself. and it somehow fails to serve us sometimes. see 74d8f325149969a7527355c9df2c40b62c354cf4 .

ceph-pr-docs/build/build
tchaikov: @epuertat i updated the commit message to explain the changes made in this commit. could you take another look?

ceph-pr-docs/build/build
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ex
+set -xo pipefail
```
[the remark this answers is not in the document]
tchaikov: i put more details in the commit message to explain why "set -e" should be dropped.

ceph-pr-docs/build/build
epuertat: Weird :s I just saw a few related bugs on Jenkins Jira...

ceph-pr-docs/build/build
epuertat: Sure.

ceph-pr-docs/build/build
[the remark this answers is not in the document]
epuertat: Ok, understood now. You could have also used `trap ... ERR` as an error handler?

epuertat (approved):
Looks good! Do we have any sample PR where this ran?

ceph-pr-docs/build/build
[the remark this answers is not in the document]
tchaikov: yes, but i think it's more straightforward and more explicit to use a variable than using a trap.

PR #1794, opened 2021-04-06
ceph-dev-new-trigger: do not build master on bionic anymore

tchaikov:
since we've replaced the bionic facet with the focal one in master,
there is no need to build bionic packages.

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #1811, opened 2021-05-14
ceph-dev-*-setup: use libc allocator when building quincy on bionic

tchaikov:
see also https://tracker.ceph.com/issues/39703, to silence the warning
std::allocator<void> is deprecated in C++17 and will be removed in
C++20, so stop using allocator<void> in quincy and up.

but the libtcmalloc shipped by bionic is still buggy (v2.5), so we should not
build qunincy and up with tcmalloc on this distro.

in this change, we only build with tcmalloc on non-bionic distro.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -821,15 +821,34 @@ gen_debian_version() {
 # Flavor Builds support
 # - CEPH_EXTRA_RPMBUILD_ARGS is consumed by build_rpms()
 # - CEPH_EXTRA_CMAKE_ARGS is consumed by debian/rules and ceph.spec directly
-ceph_build_args_from_flavor() {
+ceph_build_args_from_flavor_and_dist() {
     local flavor=$1
+    shift
+    local dist=$1
     shift
 
     # shellcheck disable=SC2034
     case "${flavor}" in
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+        # we have to build quincy and up on bionic, because some teuthology tests are
+        # still using ubuntu bionic:
+        # - the perftest depends on cosbench which is not compatible with ubuntu/focal, see
+        #   https://tracker.ceph.com/issues/49139
+        # - we need to test old clients which are built on bionic. for instance, we don't build
+        #   nautilus on focal yet.
+        # - some upgrade tests still include bionic facets.
+        if [ "$dist" == "bionic" ]; then
+            local ver
+            ver=$(git describe --abbrev=8 --match "v*" | sed s/^v// | cut -f1 -d'.')
+            # use libc allocator when building on quincy and up on bionic
+            if [ ${ver} -lt 17 ]; then
+                CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+            fi
+        else
+            # always use tcmalloc on non-bionic dists
+            CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+        fi
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
--- a/ceph-dev-new-setup/build/build
+++ b/ceph-dev-new-setup/build/build
@@ -64,7 +64,7 @@ else
     echo "forcing."
 fi
 
-ceph_build_args_from_flavor ${FLAVOR}
+ceph_build_args_from_flavor_and_dist ${FLAVOR} ${DIST}
 
 mkdir -p release
 
--- a/ceph-dev-setup/build/build
+++ b/ceph-dev-setup/build/build
@@ -36,7 +36,7 @@ rm -rf release
 echo "Running submodule update ..."
 git submodule update --init
 
-ceph_build_args_from_flavor ${FLAVOR}
+ceph_build_args_from_flavor_and_dist ${FLAVOR} ${DIST}
 
 # When using autotools/autoconf it is possible to see output from `git diff`
 # since some macros can be copied over to the ceph source, triggering this


scripts/build_utils.sh
```diff
@@ -821,15 +821,28 @@ gen_debian_version() {
...
     case "${flavor}" in
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+        # https://github.com/ceph/ceph/pull/41340,
+        if [ "$dist" == "bionic" ]; then
+            local ver
+            ver=$(git describe --abbrev=8 --match "v*" | sed s/^v// | cut -f1 -d'.')
+            # use libc allocator when building on quincy and up on bionic
+            if [ ${ver} -lt 17 ]; then
```
idryomov, line 839: Why are building quincy on bionic at all?  Do we expect to support quincy on bionic in 2022 (and beyond)?

tchaikov: @idryomov see #1794

idryomov: I see, thanks.  Might be nice to add the explanation that you closed #1794 with to the above "use libc allocator ..." comment.

idryomov: And drop the stray comma after the PR link.

tchaikov: i don't understand how closing #1794 is related with this very change. but i amend the comment with more background on why we need to build with ubuntu/bionic.

idryomov: Well, this change adds a special case for something that you yourself thought wasn't necessary until the need for continuing building quincy on bionic was cleared up in #1794.  No one else is going to have that context and if I hadn't asked #1794 wouldn't even be linked in this PR, so I think having an explanation is useful.

tchaikov: @idryomov could you help me understand what else explanation is necessary in addition to the comment? the special case is to use libc allocator on ubuntu when building quincy and up. how is it related to the fact of closing #1794 ?

scripts/build_utils.sh
```diff
@@ -821,15 +821,34 @@ gen_debian_version() {
...
+    local dist=$1
+    shift
 
     # shellcheck disable=SC2034
     case "${flavor}" in
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+        # we have to build quincy and up on bionic, because of some teuthology tests are
+        # still using ubuntu focal:
```
idryomov, line 835: "... because some teuthology tests are still using ubuntu bionic"

or

"... because of some teuthology tests that are still using ubuntu bionic"

scripts/build_utils.sh
```diff
@@ -821,15 +821,34 @@ gen_debian_version() {
...
+        # - the perftest depends on cosbench which is not compatible with ubuntu/focal, see
+        #   https://tracker.ceph.com/issues/49139
+        # - we need to test old clients which is built on bionic. for instance, we don't build
```
idryomov, line 838: which are built

scripts/build_utils.sh
```diff
@@ -821,15 +821,34 @@ gen_debian_version() {
...
+        #   nautilus on focal yet.
+        # - some upgrade tests are still include bionic facets
```
idryomov, line 840: tests still include

tchaikov:
@idryomov thank you for your review. fixed and repushed.

scripts/build_utils.sh
tchaikov: fixed.

scripts/build_utils.sh
tchaikov: fixed.

scripts/build_utils.sh
tchaikov: fixed.

scripts/build_utils.sh
idryomov: Yes, I assumed that we don't need to build quincy on bionic, as were you until you opened #1794.  I'm pretty sure that many others assume the same.  So for someone looking through this script and encountering a special case for quincy on bionic (and people usually don't look at these scripts unless there is a build or infrastructure issue which is often urgent), I thought it would be nice to explain *why* quincy is being built on bionic in the first place.

Does that answer your question?

scripts/build_utils.sh
idryomov: Not completely.  It still says ubuntu focal -- that should be ubuntu bionic.

scripts/build_utils.sh
tchaikov: sort of. i understand your point, but i don't agree with you. i think people would be more interested in

> why we have special case for not using tcmalloc on bionic

instead of 

> why are we stilling building on bionic

#1794 is related. but i don't think this change is responsible to explain to the audience all the related facts. what it should do, is to explain why we need this change. anyway, i believe that a change should be specific and should be focus on the **change** itself instead of being an encyclopedia on all related topics.

idryomov: True, but "why we have special case for not using tcmalloc on bionic" is explained in the commit message (you included a link to the tracker ticket, etc) whereas "why are we stilling building on bionic" wasn't explained anywhere.

scripts/build_utils.sh
tchaikov: i am sorry, should be fixed now.

scripts/build_utils.sh
tchaikov: > whereas "why are we stilling building on bionic" wasn't explained anywhere.

because it is not very related to the change, otherwise i'd add it in the first revision. but i added it anyway as it helps to provide more context, and so i can get my change merged. again, this change does not change the build matrix of distro x ceph-release, but it changes the matrix of distro x ceph-release x allocator, that's why i didn't want to include related information in the change.

idryomov:
Need to change the title of the commit and the PR to "ceph-dev-*-setup: use libc allocator when building quincy on bionic".  Right now it makes the impression that *all* bionic builds are switched to libc allocator.

tchaikov:
@idryomov updated title and repushed. could you kindly take another look?

tchaikov:
ahh, forgot to update the title of this PR. thanks for fixing it!

idryomov (commented):
Thanks for addressing my nits!  This looks good to me, but I don't fully understand why it is OK to call `ceph_build_args_from_flavor_and_dist` without `DIST` in ceph-dev-new-setup case, so leaving the approval to others.

tchaikov:
@idryomov d'oh! that was a bug. just fixed it.

idryomov (approved):

PR #1814, opened 2021-05-15
ceph-dev-*-setup: do not specify ALLOCATOR

tchaikov:
$DIST env variable is not set by ceph-dev-new-setup, as it prepares
the dist tarball for building on different distros. to build with
tcmalloc if gperftools 2.6.2 is available, we leave it to cmake to
decide the allocator to use. because, if ALLOCATOR is not specified,
cmake first tries to find gperftools, and falls back to JeMalloc, and
the libc allocator. after quincy, the fix for gperftools <2.6.2 is
dropped, and gperftools >= 2.6.2 is required, so it's safe. before
quincy, the fix gperftools <2.6.2 is still included, so it's also safe.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -821,10 +821,8 @@ gen_debian_version() {
 # Flavor Builds support
 # - CEPH_EXTRA_RPMBUILD_ARGS is consumed by build_rpms()
 # - CEPH_EXTRA_CMAKE_ARGS is consumed by debian/rules and ceph.spec directly
-ceph_build_args_from_flavor_and_dist() {
+ceph_build_args_from_flavor() {
     local flavor=$1
-    shift
-    local dist=$1
     shift
 
     # shellcheck disable=SC2034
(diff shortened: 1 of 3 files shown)


tchaikov:
quote from https://jenkins.ceph.com/job/ceph-dev-new-setup/55103//consoleFull

```
+ HOST=braggi12
+ echo 'Building on braggi12'
Building on braggi12
+ echo '  DIST='
  DIST=
...
+ ceph_build_args_from_flavor_and_dist crimson
+ local flavor=crimson
+ shift
+ local dist=
+ shift
```

scripts/build_utils.sh
```diff
@@ -838,17 +836,14 @@ ceph_build_args_from_flavor_and_dist() {
...
-            # use libc allocator when building on quincy and up on bionic
-            if [ ${ver} -lt 17 ]; then
-                CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
-            fi
-        else
-            # always use tcmalloc on non-bionic dists
-            CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
-        fi
+        #
+        # do not specify the allocator and let cmake figure it out:
```
idryomov, line 840:
```suggestion
        # do not specify -DALLOCATOR=tcmalloc in CEPH_EXTRA_CMAKE_ARGS
        # and let cmake figure it out:
```

scripts/build_utils.sh
```diff
@@ -838,17 +836,14 @@ ceph_build_args_from_flavor_and_dist() {
...
+        # - on quincy and up, gperftools 2.6.2 is required, so we are not impacted by
+        #   https://tracker.ceph.com/issues/49139. and cmake uses tcmalloc if gperftools 2.6.2 and
```
idryomov, line 842: I think you meant https://tracker.ceph.com/issues/39703 here.

scripts/build_utils.sh
tchaikov: fixed.

scripts/build_utils.sh
tchaikov: yes, sorry. was looking at the wrong browser tab.

tchaikov:
@idryomov fixed and repushed.

idryomov (approved):

idryomov:
For linkage: this PR fixes a build issue introduced in https://github.com/ceph/ceph-build/pull/1811.

PR #1823, opened 2021-05-25
ansible: Clean up and reconfigure ccache

djgalloway:

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -57,6 +57,7 @@
           - ditaa
           - ant
           - tmpreaper
+          - ccache
         tmp_cleaner_name: tmpreaper
       when: ansible_os_family == "Debian"
       tags: always
@@ -183,6 +184,7 @@
           - python-virtualenv
           - mock
           - docker
+          - ccache
         container_service_name: docker
         container_certs_path: "/etc/docker/certs.d/{{ container_mirror }}"
       when:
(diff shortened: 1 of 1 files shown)


ansible/examples/slave.yml
```diff
@@ -551,6 +554,26 @@
...
+      file:
+        path: "/home/{{ jenkins_user }}/.ccache"
+        state: directory
+      become_user: "{{ jenkins_user }}"
+
+    - name: Configure max ccache
+      lineinfile:
+        path: "/home/{{ jenkins_user }}/.ccache/ccache.conf"
+        regexp: '^max_size'
+        line: 'max_size = 20G'
```
dmick, line 573: I think it's really tough to come up with a useful limit for this; I don't have any idea how much is enough, if having too-small a cache actually makes things slower or just wastes space, if there's some sort of "cache prune" operation that might reduce the size in a way that makes use of knowledge of the caching algorithms, etc.  I understand there's a goal to prevent builders from dying, but it seems like a good opportunity to revisit this and figure out if what we're doing is right.  I also don't know who knows about this; IIRC it might have been @ldachary who did some studies to push us up past what we'd been using on gitbuilder (which ISTR was something like 20G :))

djgalloway: *sigh*, this PR wouldn't change anything anyway.

https://github.com/ceph/ceph/commit/5e9a1d95c9f735955b92204df1f711d5098e2b2d#diff-23f41255872328167c5e0748e25c21b4706c7915c41027fbc7470bcc4b6a475cR105

djgalloway:
Just got rid of the smaller builders.

PR #1826, opened 2021-05-26
Add ceph-grafana job

dmick:
This job builds a ceph-grafana container for a specific branch of Ceph and a specific architecture.  There are variables in the ceph.git Makefile that it invokes for versions of grafana and other included tools, but those are not currently exposed as parameters in this build job.

This job could be called by a parent job that builds both architectures on demand or via some trigger yet to be discussed.

This job's code depends on changes in https://github.com/ceph/ceph/pull/41559.

### DIFF
--- a/ceph-grafana/build/build
+++ b/ceph-grafana/build/build
@@ -1,0 +1,5 @@
+sudo yum install -y buildah
+cd monitoring/grafana/build
+make build
+make push
+make clean
--- a/ceph-grafana/config/definitions/ceph-grafana.yml
+++ b/ceph-grafana/config/definitions/ceph-grafana.yml
@@ -1,0 +1,54 @@
+- job:
+    name: ceph-grafana
+    description: 'Builds the ceph-grafana container.'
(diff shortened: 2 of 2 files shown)


ceph-grafana/build/build
```diff
@@ -0,0 +1,5 @@
+sudo yum install -y buildah
```
djgalloway, line 1: Is `buildah` just a wrapper like `podman-docker`?  Is it going to cause conflicts with other jobs that install podman/docker packages?

djgalloway:
Are the nfs-ganesha commits supposed to be in here?

ceph-grafana/build/build
dmick: No, just another tool that uses some of the same libraries as podman.  It builds container images.

dmick:
sigh, no, the nfs-ganesha commits should not be.  Fixing.

djgalloway (approved):

PR #1830, opened 2021-06-04
ceph-dev-*-trigger: stop building master on bionic

tchaikov:
in short, we don't need to test on bionic anymore for couple reasons:

- for accessing newer GCC
- for accessing more packages

see the discussion at
https://lists.ceph.io/hyperkitty/list/dev@ceph.io/thread/BW5NKQ4KTOPV7OYUBUHVDVCSH72LVIEO/

the only qa suite referencing ubuntu 18.02 is the qa/suites/orch/rook/smoke,
but it deploys ceph using Rook, and the container image of Ceph is
based on CentOS, so stop building on ubuntu/bionic does not
break it, despite the fact that 18.02 is included by
qa/suites/orch/rook/smoke/0-distro.

### DIFF
--- a/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
+++ b/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
@@ -101,7 +101,7 @@
                     DISTROS=centos8
                     FLAVOR=crimson
       # build master on:
-      # default: focal bionic centos8 leap15
+      # default: focal centos8 leap15
       # crimson: centos8
       - conditional-step:
           condition-kind: regex-match
@@ -118,7 +118,7 @@
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
                     FORCE=True
-                    DISTROS=focal bionic centos8
+                    DISTROS=focal centos8
                 - project: 'ceph-dev'
                   predefined-parameters: |
                     BRANCH=${GIT_BRANCH}
--- a/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
+++ b/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
@@ -101,7 +101,7 @@
                     DISTROS=centos8
                     FLAVOR=crimson
       # build master on:
-      # default: focal centos8 leap15
+      # default: focal centos8
       # crimson: centos8
       - conditional-step:
           condition-kind: regex-match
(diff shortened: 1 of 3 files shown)


idryomov (commented):
Probably should also drop
https://github.com/ceph/ceph-build/blob/f95a5c76557f87e47f717e0d2605ee8b1871ff5f/scripts/build_utils.sh#L847-L854

tchaikov:
@idryomov thanks, updated and repushed.

idryomov:
`s/18.02/18.04/` in the commit message?

idryomov (approved):

tchaikov:
> `s/18.02/18.04/` in the commit message?

fixed and repushed.

idryomov (commented):
One more nit, this comment appears to be outdated:
https://github.com/ceph/ceph-build/blob/a5cfb5c0cfe855670582a0b078f4bf837b69e593/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml#L104

tchaikov:
@idryomov thank you for your reviews. i fixed another place as well.

ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
```diff
@@ -101,7 +101,7 @@
                     DISTROS=centos8
                     FLAVOR=crimson
       # build master on:
-      # default: focal bionic centos8 leap15
+      # default: focal centos8 leap15
```
idryomov, line 104: I don't think master is built on `leap15` either.

tchaikov: thanks. i added a commit for it. as it's not quite related to the bionic change.

idryomov (approved):

djgalloway (approved):
Fine with me but can we wait till Monday to merge?

tchaikov:
> Fine with me but can we wait till Monday to merge?

@djgalloway on the contrary, i'd like use the week-end to check if it does not break anything. if it introduces regression, i will revert it.

jdurgin (approved):

djgalloway:
This is causing a few failures.

https://pulpito.ceph.com/pdonnell-2021-06-10_06:02:45-fs-wip-pdonnell-testing-20210610.025127-distro-basic-smithi/6164231/

tchaikov:
should be fixed by https://github.com/ceph/teuthology/pull/1654

PR #1831, opened 2021-06-04
scripts/build_utils: increase mem per job to 3000 MiB

tchaikov:
see also https://github.com/ceph/ceph/pull/41677


tchaikov:
debian packaging is not impacted yet, because braggi servers have smaller amount of cores. but to be future proof, let's update debian side as well.

PR #1838, opened 2021-06-15
ansible: It's called tmpwatch in OpenSUSE

djgalloway:

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -241,6 +241,7 @@
           # obs requirements
           - osc
           - build
+        tmp_cleaner_name: tmpwatch
       when: ansible_os_family == "Suse"
       tags: always
 


dmick (commented):
So, I'm confused how this works.  It appears that --runtime 0 is an argument to tmp_cleaner_name, but tmpwatch on CentOS8 at least doesn't document such an option.  tmpreaper on Ubuntu does.  It seems like the command also needs to be parameterized.

djgalloway:
You didn't need to go read the man page.  We're just setting a var :stuck_out_tongue_closed_eyes:.

But you are right.  And fair enough.  I will fix in a separate PR.

djgalloway:
Crap.  I accidentally pushed to this branch.  I can just change the PR title?

djgalloway:
Can merge https://github.com/ceph/ceph-build/pull/1849 after this one.

dmick (approved):

PR #1842, opened 2021-06-16
project config: dashboard-cephadm-e2e

alfonsomthd:
To be merged after https://github.com/ceph/ceph/pull/42263

Project: run Dashboard E2E tests with cephadm as orchestrator backend
  on dashboard/cephadm PRs.

Fixes: https://tracker.ceph.com/issues/51301

Signed-off-by: Alfonso Martínez <almartin@redhat.com>

### DIFF
--- a/ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
+++ b/ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
@@ -1,0 +1,83 @@
+- job:
+    name: ceph-dashboard-cephadm-e2e
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: huge && focal && x86_64
+    display-name: 'ceph: Dashboard + Cephadm E2E'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          num-to-keep: 300
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: https://github.com/ceph/ceph/
+      - rebuild:
+          auto-rebuild: true
+      - inject:
+          properties-content: |
+            TERM=xterm
+
+    parameters:
+      - string:
+          name: sha1
+          description: "commit id or a refname, like 'origin/pr/72/head'"
+
+    triggers:
+      - github-pull-request:
+          cancel-builds-on-update: true
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          white-list-labels:
+            - cephadm
+            - dashboard
+          black-list-target-branches:
+            - luminous
+            - mimic
+            - nautilus
+          trigger-phrase: 'jenkins test dashboard cephadm'
+          skip-build-phrase: '^jenkins do not test.*'
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "ceph dashboard cephadm e2e tests"
+          started-status: "running ceph dashboard cephadm e2e tests"
+          success-status: "ceph dashboard cephadm e2e tests succeeded"
+          failure-status: "ceph dashboard cephadm e2e tests failed"
+
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph.git
+          branches:
+            - origin/pr/${ghprbPullId}/merge
+          refspec: +refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*
+          browser: auto
+          timeout: 20
+          skip-tag: true
+          shallow-clone: true
+          wipe-workspace: true
+
+    builders:
+      - shell:
+          !include-raw:
+            - ../../../scripts/dashboard/install-cephadm-e2e-deps.sh
+      - shell: |
+          export CYPRESS_ARGS="--record --key $CYPRESS_RECORD_KEY --tag $ghprbTargetBranch" COMMIT_INFO_MESSAGE="$ghprbPullTitle"
+          timeout 7200 ./src/pybind/mgr/dashboard/ci/cephadm/run-cephadm-e2e-tests.sh
+
+    wrappers:
+      - inject-passwords:
+          global: true
+          mask-password-params: true
+      - credentials-binding:
+          - text:
+              credential-id: cd-cypress-record-key
+              variable: CYPRESS_RECORD_KEY
+      - ansicolor
--- a/scripts/dashboard/install-cephadm-e2e-deps.sh
+++ b/scripts/dashboard/install-cephadm-e2e-deps.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+set -ex
+
+on_error() {
+    if [ "$1" != "0" ]; then
+        printf "\n\nERROR $1 thrown on line $2\n\n"
+        printf "\n\nCollecting info...\n\n"
+        sudo journalctl --since "10 min ago" --no-tail --no-pager -x
+        printf "\n\nERROR: displaying containers' logs:\n\n"
+        docker ps -aq | xargs docker logs
+        printf "\n\nTEST FAILED.\n\n"
+    fi
+}
+
+trap 'on_error $? $LINENO' ERR
+
+sudo apt -y install libvirt-daemon-system libvirt-daemon-driver-qemu qemu-kvm libvirt-clients
+
+sudo usermod -aG libvirt $(id -un)
+newgrp libvirt  # Avoid having to log out and log in for group addition to take effect.
+sudo systemctl enable --now libvirtd
+
+if [[ $(command -v docker) == '' ]]; then
+    # Set up docker official repo and install docker.
+    sudo apt update -y
+    sudo apt install \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gnupg \
+    lsb-release
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+    echo \
+        "deb [arch=amd64 signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
+        $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+    sudo apt update -y
+    sudo apt install -y docker-ce docker-ce-cli containerd.io
+fi
+sudo groupadd docker || true
+sudo usermod -aG docker $(id -un)
+sudo systemctl start docker
+sudo chgrp "$(id -un)" /var/run/docker.sock
+
+docker info
+docker container prune -f
+
+KCLI_CONFIG_DIR="${HOME}/.kcli"
+mkdir -p ${KCLI_CONFIG_DIR}
+if [[ ! -f "${KCLI_CONFIG_DIR}/id_rsa" ]]; then
+    ssh-keygen -t rsa -q -f "${KCLI_CONFIG_DIR}/id_rsa" -N ""
+fi
+
+: ${KCLI_CONTAINER_IMAGE:='quay.io/karmab/kcli'}
+
+echo "#!/usr/bin/env bash
+
+docker run --net host --security-opt label=disable \
+    -v ${KCLI_CONFIG_DIR}:/root/.kcli \
+    -v ${PWD}:/workdir \
+    -v /var/lib/libvirt/images:/var/lib/libvirt/images \
+    -v /var/run/libvirt:/var/run/libvirt \
+    -v /var/tmp:/ignitiondir \
+    ${KCLI_CONTAINER_IMAGE} \""'${@}'"\"
+" | sudo tee /usr/local/bin/kcli
+sudo chmod +x /usr/local/bin/kcli
+
+# Install required deps.
+sudo apt update -y
+sudo apt install -y nodejs npm openssh-server
+
+# KCLI cleanup function can be found here: https://github.com/ceph/ceph/blob/master/src/pybind/mgr/dashboard/ci/cephadm/start-cluster.sh
+sudo mkdir -p /var/lib/libvirt/images/ceph-dashboard
+kcli create pool -p /var/lib/libvirt/images/ceph-dashboard ceph-dashboard
+kcli create network -c 192.168.100.0/24 ceph-dashboard


scripts/dashboard/install-kcli-deps.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo "
+alias kcli='podman run --rm --net host --security-opt label=disable \
+    -v ${HOME}/.ssh:/root/.ssh \
+    -v ${HOME}/.kcli:/root/.kcli \
+    -v ${PWD}/src/pybind/mgr/dashboard/ci/cephadm:/workdir \
+    -v /var/lib/libvirt/images:/var/lib/libvirt/images \
+    -v /var/run/libvirt:/var/run/libvirt \
+    -v /var/tmp:/ignitiondir \
+    docker.io/jolmomar/kcli'
```
jmolmo, line 23: I think it could be interestiong to use a constant to "easy" change the kcli image... one day we will use latest ... (i hope)

scripts/dashboard/install-kcli-deps.sh
```diff
@@ -0,0 +1,30 @@
...
+newgrp libvirt
+systemctl enable --now libvirtd
+
+sudo dnf install -y podman
+
+mkdir -p ~/.ssh
+cp -R src/pybind/mgr/dashboard/ci/cephadm/.kcli ~
+
+echo "
+alias kcli='podman run --rm --net host --security-opt label=disable \
```
jmolmo, line 16: what about the  "--it" parameter?

scripts/dashboard/install-kcli-deps.sh
```diff
@@ -0,0 +1,30 @@
...
+    -v ${HOME}/.ssh:/root/.ssh \
+    -v ${HOME}/.kcli:/root/.kcli \
+    -v ${PWD}/src/pybind/mgr/dashboard/ci/cephadm:/workdir \
+    -v /var/lib/libvirt/images:/var/lib/libvirt/images \
+    -v /var/run/libvirt:/var/run/libvirt \
+    -v /var/tmp:/ignitiondir \
```
jmolmo, line 22: I think you forget: 
`-v $PWD:/workdir `

jmolmo (approved):
Minor nits with the kcli command... apart of that ... fantastic job Alfonso!!

scripts/dashboard/install-kcli-deps.sh
alfonsomthd: With that I cannot capture the output that I need later on:
```
if [[ -z "${CYPRESS_BASE_URL}" ]]; then
    CYPRESS_BASE_URL="https://$(kcli info vm ceph-node-00 -f ip -v):8443"
fi
```

scripts/dashboard/install-kcli-deps.sh
alfonsomthd: In fact is set to:
-v ${PWD}/src/pybind/mgr/dashboard/ci/cephadm:/workdir

scripts/dashboard/install-kcli-deps.sh
alfonsomthd: Yes, for now I'm testing the job. I'll replace it with an env. var.

scripts/dashboard/install-kcli-deps.sh
alfonsomthd: Updated anyway.

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    triggers:
+      - github-pull-request:
+          cancel-builds-on-update: true
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          white-list-labels:
+#            - dashboard
+            - wip-alfonsomthd-testing
```
epuertat, line 39: Update?

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
```diff
@@ -0,0 +1,84 @@
...
+          black-list-target-branches:
+            - luminous
+            - mimic
+            - nautilus
+            - octopus
```
epuertat, line 44: Shouldn't we run this on octopus ones too?

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
```diff
@@ -0,0 +1,84 @@
...
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: huge && focal && x86_64
+    display-name: 'ceph: Dashboard + Cephadm E2E'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
```
epuertat, line 12: As discussed with you about avoiding concurrent jobs in the same node, we'd need the [Throttle plugin](https://plugins.jenkins.io/throttle-concurrents/) installed and [configured in JJB](https://docs.openstack.org/infra/jenkins-job-builder/properties.html?highlight=throttle#properties.throttle).

epuertat (approved):
Looks good! Thanks @alfonsomthd ! Just an observation about the throttling.

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
alfonsomthd: @djgalloway Can you help us here?

djgalloway: `concurrent` means "at the same time anywhere."  The number of workers/executors per Jenkins builder determines how many jobs can run on a node at a time.

tl;dr - you want `concurrent: true`

scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -0,0 +1,59 @@
...
+sudo systemctl enable --now libvirtd
+
+# Add podman repo for ubuntu focal.
+VERSION_ID='20.04'
+sudo sh -c "echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list"
+wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_${VERSION_ID}/Release.key -O- | sudo apt-key add -
+sudo apt update -y
+
+# Install required deps.
+sudo apt install -y nodejs npm openssh-server podman
```
djgalloway, line 31: I foresee all of this being very problematic.  To be honest, I'm a little disheartened to see it's already been run against some of the Jenkins builders as it will no doubt create headaches in the future.

These aren't ephemeral builders.  I do my best to make sure all the builders are running the same OS/kernel/packages/etc.  All of this prep is bound to eventually cause package conflicts and `apt` transaction errors when opensuse's servers are down, etc.

scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -0,0 +1,59 @@
...
+
+mkdir -p ~/.kcli ~/.ssh
+if [[ ! -f "$HOME/.ssh/id_rsa" ]]; then
+    ssh-keygen -t rsa -q -f "$HOME/.ssh/id_rsa" -N ""
```
djgalloway, line 35: Please name it something unique to this test

djgalloway: Maybe symlink to it inside the container temporarily or something.

scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -0,0 +1,59 @@
...
+    -v /var/run/libvirt:/var/run/libvirt \
+    -v /var/tmp:/ignitiondir \
+    ${KCLI_CONTAINER_IMAGE} \""'${@}'"\"
+" | sudo tee /usr/local/bin/kcli
+sudo chmod +x /usr/local/bin/kcli
+
+podman info
+sudo podman container prune -f
+
+kcli create pool -p /var/lib/libvirt/images default
```
djgalloway, line 57: Does this ever get cleaned up?

scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -0,0 +1,59 @@
...
+kcli download image fedora34 -u https://fedora.mirror.liteserver.nl/linux/releases/34/Cloud/x86_64/images/Fedora-Cloud-Base-34-1.2.x86_64.qcow2
+kcli create network -c 192.168.122.0/24 default
```
djgalloway, line 59: Same question here.

scripts/dashboard/install-cephadm-e2e-deps.sh
djgalloway: Or better yet, write the key to `$WORKSPACE/id_rsa` and mount that inside the container at `~/.ssh`

djgalloway:
This all looks great in theory but it modifies the Jenkins builders and relies on third-party infrastructure a bit more than I would like.

I would rather have a dedicated builder or few builders that have libvirt/podman stuff installed and give them a new Jenkins label (`node:` in the jjb yaml).  What do you think?

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
epuertat: @djgalloway just the opposite: as per @alfonsomthd words, more than a single `kcli` instance running might cause errors. That Throttling plugin allows to set the number of concurrent jobs per node.

scripts/dashboard/install-cephadm-e2e-deps.sh
epuertat: @alfonsomthd tried to avoid docker for the root/daemon security issue. I think it's probably more standard/straightforward to rely on root/daemon docker than to have to install podman from an non-standard source.

@djgalloway we can label these nodes as 'cephadm-kcli' (and use that label in this very job) so we have a precise control over which hosts this is run. And for avoiding installing packages during the testing process in the future, what about labeling a node or nodes as 'sandbox' so we have some 'damage control'.

alfonsomthd: > I foresee all of this being very problematic. To be honest, I'm a little disheartened to see it's already been run against some of the Jenkins builders as it will no doubt create headaches in the future.
> 

I ran it as a PoC, is it problematic testing the job to check if the idea is feasible?

> These aren't ephemeral builders. I do my best to make sure all the builders are running the same OS/kernel/packages/etc. All of this prep is bound to eventually cause package conflicts and `apt` transaction errors when opensuse's servers are down, etc.

I chose podman toavoid installing docker as a sustemd service (for simplicity).
I rely on that build because ubuntu focal does not have yet podman available as deb pkg.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: The keys are not only used by podman but by KCLI, the VM manager, so the VMs are using it so form the host you can SSH to any VM. I tried with a dedicated key and I was not able to SSH into any VM.
Nevertheless I'll investigate this a bit more.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: I do some cleanup in the associated ceph script but it does not include this.

Following your idea of having dedicated builders with libvirt+podman(or docker) and taking into account that the VM image is never modified, not deleting the image can improve the speed of the job.
Does it make sense to not do the cleanup in the dedicated nodes?

alfonsomthd:
> This all looks great in theory but it modifies the Jenkins builders and relies on third-party infrastructure a bit more than I would like.
> 
> I would rather have a dedicated builder or few builders that have libvirt/podman stuff installed and give them a new Jenkins label (`node:` in the jjb yaml). What do you think?

Sounds good to me!
So I assume that the job shouldn't install anymore these tools and only take care of setting the proper environment and launch the job script.

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
alfonsomthd: This will be the last step before merging XD

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
alfonsomthd: As per our previous conversation will run it in octopus but take into account that the octopus ceph backport will be needed to succeed.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: @djgalloway Now we store the key in KCLi config folder:
```
KCLI_CONFIG_DIR="${HOME}/.kcli"
mkdir -p ${KCLI_CONFIG_DIR}
if [[ ! -f "${KCLI_CONFIG_DIR}/id_rsa" ]]; then
    ssh-keygen -t rsa -q -f "${KCLI_CONFIG_DIR}/id_rsa" -N ""
fi
```
Let us know if this works for you.

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
alfonsomthd: @djgalloway As per our previous conversations we concluded that putting `concurrent: true` does not imply that more than 1 execution of the same job can run on the same machine, so it is safe to set it as true.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: @djgalloway As per our previous conversation we agreed on using docker instead of podman.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: @djgalloway As per our last conversation we agreed on using the `node: huge && focal && x86_64` machines and a dedicated libvirt pool and network (ceph-dashboard) to avoid conflicts in the future.

scripts/dashboard/install-cephadm-e2e-deps.sh
alfonsomthd: Answered above.

scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -0,0 +1,74 @@
...
+" | sudo tee /usr/local/bin/kcli
+sudo chmod +x /usr/local/bin/kcli
+
+# Install required deps.
+sudo apt update -y
+sudo apt install -y nodejs npm openssh-server
+
+sudo mkdir -p /var/lib/libvirt/images/ceph-dashboard
+kcli create pool -p /var/lib/libvirt/images/ceph-dashboard ceph-dashboard
+kcli create network -c 192.168.100.0/24 ceph-dashboard
```
djgalloway, line 74: Can you add a teardown script to clean this up after the job completes?

alfonsomthd: @djgalloway The cleanup is tirggered when the job finishes on success/error here (in the other PR):
https://github.com/ceph/ceph/pull/42263/files?file-filters%5B%5D=.sh#diff-ccaf2a6af4475351e29cd34ed246c9fbd3cbbc14c93d961a360c283eecf02e88R5

In our last conversation we agreed to not erase the libvirt folder/pool/network but to rename it (s/default/ceph-dashboard) in order to not constantly download the VM fedora image and to avoid name conflicts in the future.
Do you prefer to completely erase it?

djgalloway: Ah, right.  Fair enough.  Can you just put a comment in the file pointing to that cleanup function please?  The canonical link to the file in `master` is fine even though that PR isn't merged yet.

alfonsomthd: Sure.

alfonsomthd: Done.

djgalloway (approved):
Can we wait to merge until Monday please?  I'm on PTO tomorrow and won't be available to deal with the fallout if things go bad.

alfonsomthd:
> Can we wait to merge until Monday please? I'm on PTO tomorrow and won't be available to deal with the fallout if things go bad.

Sure. We'll merge https://github.com/ceph/ceph/pull/42263 first.

PR #1848, opened 2021-06-17
ceph-website: Use proper build env per branch

djgalloway:
Fixes: https://github.com/ceph/ceph.io/issues/161

### DIFF
--- a/ceph-website/build/build
+++ b/ceph-website/build/build
@@ -11,7 +11,15 @@ BRANCH=$(echo $GIT_BRANCH | sed 's:.*/::')
 
 npm install
 
-npm run build
+if [ "$BRANCH" == "master" ]; then
+  npm run build:production
+elif [ "$BRANCH" == "main" ]; then
+  npm run build:production
+elif [ "$BRANCH" == "develop" ]; then
+  npm run build:production
+else
+  npm run build:development
+fi
 
 if [ ! -d /opt/www/${BRANCH} ]; then
   mkdir -p /opt/www/${BRANCH}


ceph-website/build/build
```diff
@@ -11,7 +11,15 @@ BRANCH=$(echo $GIT_BRANCH | sed 's:.*/::')
 
 npm install
 
-npm run build
+if [ "$BRANCH" == "master" ]; then
+  npm run build:production
+elif [ "$BRANCH" == "main" ]; then
+  npm run build:production
+elif [ "$BRANCH" == "develop" ]; then
+  npm run build:production
```
adamduncan, line 19: @djgalloway I think we can omit the `develop` clause here and let it fall through to the `else` clause and build for `npm run build:development` 👍

djgalloway: Will be using the new `feature` branch -> `main` branch workflow by the time the site is live?  My thinking was we could merge this now and when the workflow is finalized and the site is live, then I could remove that stanza.

adamduncan: Ah gotcha. My understanding is that we will use the feature-branch approach as soon as we go live, yep.

Either way, with this change (https://github.com/ceph/ceph.io/commit/0ac9abaf980ace881b1ecbbb93f65a6adcccdd5b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14-R15), we have discrete `development` and `production` scripts already set up. Really the develop.ceph.io should've always been building in development mode.

Hope that helps? Happy to take your lead on this one

djgalloway: So when this jenkins job was running `npm run build`, it was already doing https://github.com/ceph/ceph.io/commit/0ac9abaf980ace881b1ecbbb93f65a6adcccdd5b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14-R15 ?

adamduncan: Since yesterday, yep. Thoughprior to that `build` was running for develop.ceph.io with the `NODE_ENV=production`, which was technically incorrect

djgalloway: So this PR is not required then, right?

djgalloway: Nevermind.  I'm pretty sure it is.

adamduncan: Correct, I believe it is. 

Given we've decided on a `main`/feature branch approach, it my point above re: `develop` isn't of great importance.

Thank you

PR #1862, opened 2021-07-02
build_utils: set extra args for jaeger based on branch name

ideepika:
1. if jaeger in branch name for ceph-ci: this shall only build focal and centos 8
   default packages with jaeger on, the FLAVOR=jaeger gets exported to ceph_extra_args_flavor 

Jaeger can be useful when someone wants to do analysis with tracing, most preferably we want to use it on centos 8 and focal, for component of interest, that can be a perf/latency analysis for crimson to debugging operations in client, please use preferred method to produce ceph builds with jaeger as adding jaeger suffix to branch name. 

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -131,7 +131,7 @@
       - conditional-step:
           condition-kind: shell
           condition-command: |
-            echo "${GIT_BRANCH}" | grep -v '\(luminous\|mimic\|nautilus\|octopus\|pacific\|crimson-only\)'
+            echo "${GIT_BRANCH}" | grep -v '\(luminous\|mimic\|nautilus\|octopus\|pacific\|crimson-only\|jaeger)'
           on-evaluation-failure: dont-run
           steps:
             - shell:
@@ -171,6 +171,27 @@
                     FORCE=True
                     DISTROS=centos8
                     FLAVOR=crimson
+                    ARCHS=x86_64
+      # Build jaegertracing branch on needed env,  don't waste resources on the default one.
+      # Useful for testing specific builds failure
+      # default: focal, centos8
+      - conditional-step:
+          condition-kind: regex-match
+          regex: .*jaeger.*
+          label: '${GIT_BRANCH}'
+          on-evaluation-failure: dont-run
+          steps:
+            - shell:
+                !include-raw:
+                - ../../../scripts/build_utils.sh
+                - ../../build/notify
+            - trigger-builds:
+                - project: 'ceph-dev-new'
+                  predefined-parameters: |
+                    BRANCH=${GIT_BRANCH}
+                    FORCE=True
+                    DISTROS=centos8 focal
+                    FLAVOR=jaeger
                     ARCHS=x86_64
 
     wrappers:
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -840,7 +840,7 @@ gen_debian_version() {
 # Flavor Builds support
 # - CEPH_EXTRA_RPMBUILD_ARGS is consumed by build_rpms()
 # - CEPH_EXTRA_CMAKE_ARGS is consumed by debian/rules and ceph.spec directly
-# - PROFILES is consumed by pbuilder
+# - PROFILES is consumed by build_debs()
 ceph_build_args_from_flavor() {
     local flavor=$1
     shift
@@ -914,7 +914,12 @@ build_debs() {
 
     CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS $(extra_cmake_args)"
     DEB_BUILD_OPTIONS="parallel=$(get_nr_build_jobs)"
-    PROFILES="nocheck,$PROFILES"
+
+    if [ -z "$PROFILES" ]; then
+      PROFILES="nocheck,$PROFILES"
+    else
+      PROFILES="nocheck"
+    fi
 
     # pass only those env vars specifically noted
     sudo \
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -840,6 +840,7 @@ gen_debian_version() {
 # Flavor Builds support
 # - CEPH_EXTRA_RPMBUILD_ARGS is consumed by build_rpms()
 # - CEPH_EXTRA_CMAKE_ARGS is consumed by debian/rules and ceph.spec directly
+# - PROFILES is consumed by pbuilder
 ceph_build_args_from_flavor() {
     local flavor=$1
     shift
@@ -854,6 +855,10 @@ ceph_build_args_from_flavor() {
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
         CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
         CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_SEASTAR=ON"
+        ;;
+    jaeger)
+        CEPH_EXTRA_RPMBUILD_ARGS="--with jaeger"
+        PROFILES="pkg.ceph.jaeger"
         ;;
     *)
         echo "unknown FLAVOR: ${FLAVOR}" >&2
@@ -909,6 +914,7 @@ build_debs() {
 
     CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS $(extra_cmake_args)"
     DEB_BUILD_OPTIONS="parallel=$(get_nr_build_jobs)"
+    PROFILES="nocheck,$PROFILES"
 
     # pass only those env vars specifically noted
     sudo \
@@ -918,7 +924,7 @@ build_debs() {
         --distribution $DIST \
         --basetgz $pbuilddir/$DIST.tgz \
         --buildresult $releasedir/$cephver \
-        --profiles nocheck \
+        --profiles $PROFILES \
         --use-network yes \
         $releasedir/$cephver/ceph_$bpvers.dsc
 
@@ -1392,10 +1398,15 @@ setup_rpm_build_deps() {
 
     sed -e 's/@//g' < ceph.spec.in > $DIR/ceph.spec
 
-    if [ "$FLAVOR" = "crimson" ]; then
-        # enable more build depends required by crimson
-        sed -i -e 's/%bcond_with seastar/%bcond_without seastar/g' $DIR/ceph.spec
-    fi
+    # enable more build depends required by build flavor(jaeger, crimson)
+    case "${FLAVOR}" in
+    crimson)
+      sed -i -e 's/%bcond_with seastar/%bcond_without seastar/g' $DIR/ceph.spec
+        ;;
+    jaeger)
+      sed -i -e 's/%bcond_with jaeger/%bcond_without jaeger/g' $DIR/ceph.spec
+        ;;
+    esac
 
     # Make sure we have all the rpm macros installed and at the latest version
     # before installing the dependencies, python3-devel requires the
--- a/ceph-dev-new/config/definitions/ceph-dev-new.yml
+++ b/ceph-dev-new/config/definitions/ceph-dev-new.yml
@@ -56,8 +56,9 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           choices:
             - default
             - crimson
+            - jaeger
           default: "default"
-          description: "Type of Ceph build, choices are: crimson, default. Defaults to: 'default'"
+          description: "Type of Ceph build, choices are: crimson, jaeger, default. Defaults to: 'default'"
 
       - string:
           name: CI_CONTAINER
--- a/ceph-dev-setup/build/build
+++ b/ceph-dev-setup/build/build
@@ -36,6 +36,7 @@ rm -rf release
 echo "Running submodule update ..."
 git submodule update --init
 
+# export args for building optional packages
 ceph_build_args_from_flavor ${FLAVOR}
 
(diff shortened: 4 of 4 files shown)


ideepika:
needed for: https://shaman.ceph.com/builds/ceph/wip-osd-tracing

ideepika:
@tchaikov is there a way to test this PR before merging?

ideepika:
@djgalloway @tchaikov I cannot test https://github.com/ceph/ceph/pull/38783 
without resolving build issues we cannot publish: https://wip-jaeger-blog.ceph.io/en/news/blog/2021/distributed-tracing-in-ceph-using-jaeger/ 
the context is quite disintegrated to follow, I will try to summarize if it helps with review:
1. ceph-build triggers Jenkins job: 
2. pull CMakeLists.txt file from ceph-ci for pushed branch, check if WITH_JAEGER=ON, if yes > export WITH_JAEGER=true
- Debian: add pkg.ceph.jaeger to pbuilder profiles
- rpm: change `s/with jaeger/without jaeger` in spec file, add --with jaeger to rpm extra args
- set WITH_JAEGER=ON as CMAKE_EXTRA_OPTS
HTH

scripts/build_utils.sh
```diff
@@ -837,6 +837,16 @@ gen_debian_version() {
     echo "${raw}${bptag}"
 }
 
+setup_cmakeoptions_from_ci() {
+  if [-z BRANCH ];then
```
djgalloway, line 841: $BRANCH will always be set.  Kind of unnecessary.  Even if it wasn't, returning 0 silently isn't ideal if the function is always called.  An `else echo "\$BRANCH is not defined"` would suffice.

scripts/build_utils.sh
```diff
@@ -837,6 +837,16 @@ gen_debian_version() {
...
+    curl "https://raw.githubusercontent.com/ceph/ceph-ci/$BRANCH/CMakeLists.txt" --output CMakeLists.txt.$BRANCH
```
djgalloway, line 842: A couple things here.

1. It's not necessary.  The Jenkins job has already cloned the ceph repo and checked out whatever branch is being built.  Just `grep -q "^WITH_JAEGER.*ON" CMakeLists.txt`.
2. You're running this function in `ceph-dev-setup` which builds the tarballs, etc. for branches in ceph.git.  But you're attempting to `curl` a file from ceph-ci.git.

ceph-ci.git:BRANCH != ceph.git:BRANCH.

Additionally, ceph-ci.git doesn't have pacific, octopus, etc. which is all the ceph-dev pipeline builds.

ceph-dev = ceph.git
ceph-dev-new = ceph-ci.git

tchaikov: i'd suggest just conditionalize on `BRANCH`, instead of checking a certain file in repo.

scripts/build_utils.sh
ideepika: if that is the case, i am okay removing it

scripts/build_utils.sh
ideepika: thank you for the suggestions! updated: https://github.com/ceph/ceph-build/pull/1862/files#diff-1326d7a2d74cace6f00e1cd5d3ad15ead3fc6359eed098d2133fd72c0ec6700bR840

djgalloway:
This looks better.  I'd like @tchaikov's review again before merging though.

scripts/build_utils.sh
```diff
@@ -859,6 +865,12 @@ ceph_build_args_from_flavor() {
         echo "unknown FLAVOR: ${FLAVOR}" >&2
         exit 1
     esac
+
+    if [ -z "$WITH_JAEGER" ]; then
+      CEPH_EXTRA_RPMBUILD_ARGS="--with jaeger"
+      CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_JAEGER=ON"
```
tchaikov, line 871: i think this line is not necessary anymore. as you've passed this message down to `debian/rules` by setting `profiles` already.

the crimson flavor is not updated yet, that's why we are still using `CEPH_EXTRA_CMAKE_ARGS` to build the crimson flavor.

scripts/build_utils.sh
```diff
@@ -837,6 +837,12 @@ gen_debian_version() {
     echo "${raw}${bptag}"
 }
 
+set_extra_buildargs() {
```
tchaikov, line 840:
```suggestion
maybe_set_jaeger() {
```

tchaikov: but a better solution is to follow the example of https://github.com/ceph/ceph-build/blob/c696e3df69544d0753f19db47b4d4cf0a611602f/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml#L157-L174

ideepika: ahan, thanks for pointing out, this is so useful!!

scripts/build_utils.sh
ideepika: makes sense, removed

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -172,6 +172,28 @@
                     DISTROS=centos8
                     FLAVOR=crimson
                     ARCHS=x86_64
+      # Build jaegertracing branch on needed env,  don't waste resources on the default one.
+      # Useful for testing specific builds failure
+      # default: focal
+      # crimson: centos8
```
tchaikov, line 179: this line does not apply.

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -172,6 +172,28 @@
...
+            - shell:
+                !include-raw:
+                - ../../../scripts/build_utils.sh
+                - ../../build/notify
+            - trigger-builds:
+                - project: 'ceph-dev-new'
+                  predefined-parameters: |
+                    BRANCH=${GIT_BRANCH}
+                    FORCE=True
+                    DISTROS=centos8 focal
```
tchaikov, line 195: what you need is to add a flavor of "jaeger" to `ceph-dev-new/config/definitions/ceph-dev-new.yml`, so it's another choice of `FLAVOR` parameter. and this parameter can be consumed by the function of `ceph_build_args_from_flavor()` in the shell script of `ceph-dev-new-setup/build/build`

tchaikov: or you can define another parameter instead of reusing to FLAVOR, but the idea is 

1. do not use `FLAVOR=crimson` for buliding jaeger, as they are not related.
2. use this machinery to pass a message to script used by the downstream project of `ceph-dev-new`  that we want build with jaeger.

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -172,6 +172,28 @@
...
+                    FLAVOR=crimson
```
tchaikov, line 196: this is wrong.

ideepika: sure, we can keep them exclusive, but if a user builds ceph with both jaeger and crimson, they might hit build failures found crimson as more exhaustive testing set. But it's still in development, we can defer that :)

tchaikov: > if branch will have `jaeger-only` in branch name, it shall only build
>  focal, crimson-centos8 x86_64

sorry, i assumed that you wanted to build jaeger only build using the same machinery i proposed at https://github.com/ceph/ceph-build/pull/1862#discussion_r665488883 . if you think otherwise. could you please elaborate in the commit message? as i don't understand how `jaeger-only` in the branch name  is connected to crimson flavor build.

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
ideepika: done, jaeger could be tested both by specifying in branch name, as well as passing FLAVOR=jaeger now

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
ideepika: sure, I will add

ideepika: https://github.com/ceph/ceph-build/pull/1862#issue-682340359

scripts/build_utils.sh
```diff
@@ -837,6 +837,12 @@ gen_debian_version() {
     echo "${raw}${bptag}"
 }
 
+maybe_set_jaeger() {
+    if (echo $BRANCH | grep -i "jaeger"); then
+      export WITH_JAEGER=true
+    fi
+}
```
tchaikov: i think there is no point to enable jaeger support in 3 ways. one way would suffice.

djgalloway: +1.  ceph-build.git is already a house of cards.  Let's not add more :smile:

ideepika: ack, got it, tried to keep it minimal this time

scripts/build_utils.sh
```diff
@@ -855,6 +856,10 @@ ceph_build_args_from_flavor() {
         CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
         CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_SEASTAR=ON"
         ;;
+    jaeger)
+	CEPH_EXTRA_RPMBUILD_ARGS="--with jaeger"
+	export PROFILES="pkg.ceph.jaeger"
```
tchaikov, line 861: could you please use space for indent? there is no need to use `export` here, i think.

scripts/build_utils.sh
```diff
@@ -1398,16 +1398,6 @@ setup_rpm_build_deps() {
...
     sed -e 's/@//g' < ceph.spec.in > $DIR/ceph.spec
 
-    # enable more build depends required by build flavor(jaeger, crimson)
-    case "${FLAVOR}" in
-    crimson)
-        sed -i -e 's/%bcond_with seastar/%bcond_without seastar/g' $DIR/ceph.spec
-        ;;
-    jaeger)
-	sed -i -e 's/%bcond_with jaeger/%bcond_without jaeger/g' $DIR/ceph.spec
-        ;;
```
tchaikov: this change should be squashed into the change of "build_utils: export jaeger build args if FLAVOR == jaeger".

tchaikov: and please note, `yum-builddep` does not consume `--with jaeger/crimson`. so i think we probably have to live with this.

scripts/build_utils.sh
ideepika: ack, done

scripts/build_utils.sh
ideepika: :( added to https://github.com/ceph/ceph-build/pull/1862/commits/13a4afe33e55147c08af86739e650e0e9266689e

scripts/build_utils.sh
```diff
@@ -1392,10 +1398,15 @@ setup_rpm_build_deps() {
...
-    if [ "$FLAVOR" = "crimson" ]; then
-        # enable more build depends required by crimson
+    # enable more build depends required by build flavor(jaeger, crimson)
+    case "${FLAVOR}" in
+    crimson)
         sed -i -e 's/%bcond_with seastar/%bcond_without seastar/g' $DIR/ceph.spec
-    fi
+        ;;
+    jaeger)
+	sed -i -e 's/%bcond_with jaeger/%bcond_without jaeger/g' $DIR/ceph.spec
```
tchaikov, line 1407: please use spaces instead of tabs.

ideepika: done

tchaikov (approved):

scripts/build_utils.sh
```diff
@@ -909,6 +914,7 @@ build_debs() {
 
     CEPH_EXTRA_CMAKE_ARGS="$CEPH_EXTRA_CMAKE_ARGS $(extra_cmake_args)"
     DEB_BUILD_OPTIONS="parallel=$(get_nr_build_jobs)"
+    PROFILES="nocheck,$PROFILES"
```
djgalloway, line 917: Is that comma there going to be a problem if `$PROFILES` is empty?  Won't that run pbuilder with `--profiles nocheck,` instead of `--profiles nocheck`?

djgalloway:
This is a lot cleaner.  Thanks for your flexibility, @ideepika.  Just that one [question](https://github.com/ceph/ceph-build/pull/1862#discussion_r667148319) and I think we're good.

scripts/build_utils.sh
ideepika: i did check for it in docs it mentions it's supported, but could not test it:
```
ore than one build profile can be activated at a time. Multiple profiles are specified by separating them with commas in commandline arguments and by separating them with spaces in the DEB_BUILD_PROFILES environment variable. Some common profile names are "nocheck", "nodoc", "nobiarch" and "cross".

```
https://wiki.debian.org/BuildProfileSpec

ideepika: @djgalloway to be safe I added a commit to address this

scripts/build_utils.sh
```diff
@@ -840,6 +840,7 @@ gen_debian_version() {
 # Flavor Builds support
 # - CEPH_EXTRA_RPMBUILD_ARGS is consumed by build_rpms()
 # - CEPH_EXTRA_CMAKE_ARGS is consumed by debian/rules and ceph.spec directly
+# - PROFILES is consumed by pbuilder
```
tchaikov, line 843:
```suggestion
# - PROFILES is consumed by build_debs()
```

ideepika: done

tchaikov (approved):

PR #1873, opened 2021-07-10
scripts/build_utils: do not include nocheck in $PROFILES

tchaikov:
we don't use nocheck for disabling the "make check" dependencies,
instead, <pkg.ceph.check> is used to enable "make check" dependencies.
so they are not enabled by default.


ideepika (approved):
Thanks @tchaikov

PR #1874, opened 2021-07-10
ceph-dev-new-trigger: add escape ) by adding \ before it

tchaikov:
this should address the regression introduced by
99ac2b60d3e9c236802daf752d9924076151bf82


ideepika (approved):
Thanks @tchaikov !

PR #1875, opened 2021-07-10
scripts/build_utils.sh: quote $PROFILES

tchaikov:
otherwise the following "--use-network" would be taken as a profile,
and "yes" is used as the path to package's dsc file.


ideepika (approved):
Pretty sentitive change! Thanks again @tchaikov!

PR #1876, opened 2021-07-12
build_utils: fix $PROFILES not storing requried profiles

ideepika:
theory: ceph_build_args_from_flavor runs and set's variables for subshell
running `ceph-dev-new-setup/build/build` while we consume $PROFILES from
subshell for `ceph-build/build/build_deb` which is why $PROFILES is
always --profiles '' irrespective of flavor used.

Falied example: 
https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=focal,DIST=focal,MACHINE_SIZE=gigantic/55927//consoleFull

flavor: jaeger, working fine since I can see, --with jaeger being set.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -847,17 +847,17 @@ ceph_build_args_from_flavor() {
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
         CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
-        PROFILES=""
+        export PROFILES=""
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -847,17 +847,17 @@ ceph_build_args_from_flavor() {
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
         CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
-        PROFILES=""
+        export PROFILES=""
```
tchaikov, line 850: i think #1877 is a more correct fix.

ideepika: yep, I like the idea! closing in favor

PR #1879, opened 2021-07-15
add ceph-grafana-trigger job

dmick:
Build both x86_64 and arm64 containers using ceph-grafana job,
and then pull, create a combined manifest, and push, currently to
both docker.io and quay.io


djgalloway (approved):
LGTM

PR #1883, opened 2021-07-21
ceph-container: publish ceph/ceph to quay.io

dsavineau:
We're switching from docker.io registry for the ceph/ceph container images
to quay.io registry.
This requires to have a new credential-id in jenkins and update the
environment variables used by the ceph-container build script.


guits (approved):

leseb:
@dsavineau what about `ceph/daemon-base` images?

dsavineau:
> @dsavineau what about `ceph/daemon-base` images?

That's the next step ;)

PR #1888, opened 2021-07-31
ansible/examples/slave.yml: add python3-venv to python3_debs of debian

tchaikov:
let's add python3-venv.

keep python3-virtualenv around for a while just in case.

### DIFF
--- a/ansible/examples/slave.yml
+++ b/ansible/examples/slave.yml
@@ -106,6 +106,7 @@
           - python3
           - python3-dev
           - python3-pip
+          - python3-venv
           - python3-virtualenv
       when:
         - ansible_os_family == "Debian"


djgalloway (approved):

ktdreyer:
To expand on "just in case" -> do you have any use-cases in mind?

Do we need to switch 100% of the Jenkins jobs over to `venv` in order to drop `virtualenv`?

tchaikov:
> To expand on "just in case" -> do you have any use-cases in mind?

after a second thought, i can think of two use cases:

1. we do use `virtualenv` otherwhere. see https://github.com/ceph/ceph-build/search?p=3&q=virtualenv.
2. we are still testing branches without https://github.com/ceph/ceph/pull/42575 on ubuntu focal. for instance, pacific does not include this change.

> 
> Do we need to switch 100% of the Jenkins jobs over to `venv` in order to drop `virtualenv`?

i doubt that. as i think the goal is to drop `virtualenv` for building ceph on distro without `python3-virtualenv`, not to drop `virtualenv`.  to minimize the divergency of our building / testing process on different distros, namely rhel8 (rhel9 in future) and ubuntu focal, all `virtualenv` were replaced with `python3 -m venv` in https://github.com/ceph/ceph/pull/42575.

to prepare the jenkins worker nodes for running, for instance, "make check" and "doc build" jobs with the master branch which contains the `python3 -m venv`, we have to install `python3-venv`, but what about the older branches, like pacific.

PR #1895, opened 2021-08-19
DNM: ceph-dev-*: Trigger separate job to inform shaman of failure

djgalloway:
This doesn't work.  I couldn't figure out a way to pass the `$NORMAL_DISTRO`, `$NORMAL_DISTRO_VERSION`, and `$NORMAL_ARCH` vars from the failed job to the new triggered job.

Using `current-parameters: true` almost worked but since we're doing multijob projects, those vars look like `DISTROS='centos8 focal'`.

I just didn't want to lose the work.

### DIFF
--- a/build-failure/build/failure
+++ b/build-failure/build/failure
@@ -1,0 +1,42 @@
+#!/bin/bash -ex
+
+# note: the failed_build_status call relies on normalized variable names that
+# are infered by the builds themselves. If the build fails before these are
+# set, they will be posted with empty values
+BRANCH=`branch_slash_filter $BRANCH`
+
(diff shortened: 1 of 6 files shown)


build-failure/build/failure
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash -ex
+
+# note: the failed_build_status call relies on normalized variable names that
+# are infered by the builds themselves. If the build fails before these are
+# set, they will be posted with empty values
+BRANCH=`branch_slash_filter $BRANCH`
+
+# This build-failure job will only have one word for the $DISTROS parameter
```
djgalloway, line 8: This is false

PR #1903, opened 2021-09-08
chacra-pull-requests: Define postgresql_version on >= xenial

djgalloway:

### DIFF
--- a/chacra-pull-requests/setup/playbooks/tasks/postgresql.yml
+++ b/chacra-pull-requests/setup/playbooks/tasks/postgresql.yml
@@ -14,6 +14,16 @@
       postgresql_version: "9.5"
   when: ansible_distribution_release == "xenial"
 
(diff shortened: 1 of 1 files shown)


dmick:
9.5 seems pretty old at this point; 14 is in beta.  I don't know anything about breaking changes, but it might be worth making sure we're not moving into "unsupported" releases

djgalloway:
> 9.5 seems pretty old at this point; 14 is in beta. I don't know anything about breaking changes, but it might be worth making sure we're not moving into "unsupported" releases

Better now?

dmick:
I guess.  I was thinking "install the latest", which might mean "install 'postgresql'", but I don't know 1) for sure, 2) how good our unit tests are or 3) how good their backward compatibility is.  This is probably middle ground

dmick (approved):

PR #1927, opened 2021-11-04
Add ceph-windows-pull-requests job & Update ceph-windows-test job

ionutbalutoiu:
* Update `ceph-windows-test` job
  * Create `scripts/ceph-windows` directory with the common scripts for the Ceph Windows testing.
  * Move `run_tests` script to `scripts/ceph-windows`.
  * Remove VM image caching from the `run_tests` script.
  * Add `cleanup_tests` script to `scripts/ceph-windows`, and call this as a post-build step when the job is aborted.
  * Add `get_chacra_build` build script, and refactor away this logic from the `run_tests` script.
  * Move cleanup Windows tests function to `scripts/build_utils.sh`.
* Add `ceph-windows-pull-requests` job
  * Used to run Windows tests against Ceph pull requests.

### DIFF
--- a/ceph-windows-pull-requests/config/definitions/ceph-windows-pull-requests.yml
+++ b/ceph-windows-pull-requests/config/definitions/ceph-windows-pull-requests.yml
@@ -1,0 +1,90 @@
+- job:
+    name: ceph-windows-pull-requests
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    node: amd64 && focal && libvirt
+    display-name: 'ceph-windows: Pull Requests'
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          num-to-keep: 300
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: https://github.com/ceph/ceph/
+      - rebuild:
+          auto-rebuild: true
+      - inject:
+          properties-content: |
+            TERM=xterm
+
+    parameters:
+      - string:
+          name: ghprbPullId
+          description: "The GitHub pull request id, like '72' in 'ceph/pull/72'"
+
+    triggers:
+      - github-pull-request:
+          cancel-builds-on-update: true
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          trigger-phrase: 'jenkins test windows'
+          skip-build-phrase: '^jenkins do not test.*'
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "ceph windows tests"
+          started-status: "running ceph windows tests"
+          success-status: "ceph windows tests succeeded"
+          failure-status: "ceph windows tests failed"
+
+    scm:
+      - git:
+          url: https://github.com/ceph/ceph.git
+          branches:
+            - origin/pr/${ghprbPullId}/merge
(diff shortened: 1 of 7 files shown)


ceph-windows-pull-requests/config/definitions/ceph-windows-pull-requests.yml
```diff
@@ -0,0 +1,89 @@
...
+          permit-all: true
+          auto-close-on-fail: false
+          status-context: "ceph windows tests"
+          started-status: "running ceph windows tests"
+          success-status: "ceph windows tests succeeded"
+          failure-status: "ceph windows tests failed"
+
+    scm:
+      - git:
+          url: https://github.com/ionutbalutoiu/ceph.git
```
djgalloway, line 50: Is this just for testing purposes?

ionutbalutoiu: Yes, correct. I fixed this in the last PR update.

ceph-windows-test/build/utils.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+
+CEPH_WINDOWS_ZIP=${CEPH_WINDOWS_ZIP:-"$WORKSPACE/ceph.zip"}
+
+VM_NAME="ceph-win-ltsc2019-${BUILD_ID}"
```
djgalloway, line 7: You may not want to put `${BUILD_ID}` at the end;  Or maybe add a character or something after it.  I've run into issues in the past where the Build ID was 1, for example, and then all builds starting with 1 were also affected.

ionutbalutoiu: I can change the name to `ceph-win-ltsc2019-${JOB_NAME}-${BUILD_ID}`.

Adding `${JOB_NAME}` as part of the name should assure that we get an unique name for the VM.

ionutbalutoiu:
@djgalloway @dmick 

I updated this PR according to the latest discussions.

A couple of key points about the changes here. There's common scripting between the new `ceph-windows-pull-requests` job, and the previous `ceph-windows-test` job.

Therefore:

1. I added a new directory (`./scripts/ceph-windows`) with the scripts that are called by both. For now, they only contain `run_tests`, and `cleanup_tests`.
1. The VM image caching logic, from the `run_tests` script, was removed.

    * The caching was done in the `$WORKSPACE` directory, and the `run_tests` is a common script now, potentially to be called into multiple places. That would mean wasted space on the disk. We could switch to a cache directory location that's outside `$WORKSPACE`, but I believe that we can get rid of this since the VM image is downloaded super fast into the Jenkins agents.

Let me know what do you think about the changes here.

ceph-windows-test/build/utils.sh
djgalloway: The `${BUILD_ID}` being at the end of the VM name is the problem.

`$ grep ceph-win-ltsc2019-1` will return results for:
```
ceph-win-ltsc2019-1
ceph-win-ltsc2019-12
ceph-win-ltsc2019-123
etc.
```

djgalloway: Nevermind.  You're not grepping anywhere.  Apologies.

ionutbalutoiu: > Nevermind. You're not grepping anywhere. Apologies.

I grep in the `cleanup_windows_tests_env` function to delete the virsh VM.

ionutbalutoiu:
@dmick @djgalloway 

Let me know if there's anything else to do to have this PR merged.

PR #1928, opened 2021-11-04
ceph-dev*: Start building Windows binaries for master

djgalloway:
Tested and working https://shaman.ceph.com/builds/ceph/master/884f89ea6f89f10f96a4dd2ad0922a823e26d96d/default/286496/


zmc (approved):
Looks good to me! Only change I would consider suggesting would be to be more specific about the windows version, but that's likely premature.

PR #1931, opened 2021-11-05
container-{flake8,lint}: always pull the image

dsavineau:
Otherwise we will still use an outdated container image.


dsavineau:
> Aren't we going to get rate limited when pulling that?

for those images I don't think so because we don't run those jobs very often

djgalloway (approved):

PR #1933, opened 2021-11-16
Update ceph-windows-pull-requests job

ionutbalutoiu:
Run the job only against PRs that target the `master` branch.


djgalloway (approved):

dmick (commented):
Minor issue, but, when titling things, try to include information.  All pull requests update code, so the title doesn't add anything; "ceph-windows-pull-request: build only for master" would be much more useful.

next time.

ionutbalutoiu:
> Minor issue, but, when titling things, try to include information. All pull requests update code, so the title doesn't add anything; "ceph-windows-pull-request: build only for master" would be much more useful.
> 
> next time.

Noted. Thank-you for pointing it out!

PR #1945, opened 2021-12-28
ceph-dashboard-cephadm-e2e-job: update workspace path 

nizamial09:
In the nightly runs we saw that the ceph_cluster.yml file which we used
to deploy kcli plans is not up-to-date with the master one. It cause
problems in the nightlies. We've narrowed down the problem is with the basedir
and and also with the path difference in the dependency script and the location of
ceph. So changing the workspace path to workspace/ceph.

Also jenkins adviced against using basedir in the jenkins config
Screenshot from 2022-01-04 20-00-04

### DIFF
--- a/ceph-dashboard-cephadm-e2e-nightly/config/definitions/ceph-dashboard-cephadm-e2e-nightly.yml
+++ b/ceph-dashboard-cephadm-e2e-nightly/config/definitions/ceph-dashboard-cephadm-e2e-nightly.yml
@@ -18,6 +18,7 @@
     block-downstream: false
     block-upstream: false
     retry-count: 3
+    workspace: /home/jenkins-build/build/workspace/{name}/ceph
     properties:
       - build-discarder:
           days-to-keep: 15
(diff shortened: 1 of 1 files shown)


scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -76,7 +76,7 @@ echo "#!/usr/bin/env bash
 
 docker run --net host --security-opt label=disable \
     -v ${KCLI_CONFIG_DIR}:/root/.kcli \
-    -v ${PWD}:/workdir \
+    -v ${PWD}/ceph:/workdir \
```
alfonsomthd, line 79: @nizamial09 This script should also run in a local environment, and AFAIK in a local environment this extra folder does not exist.

nizamial09: Updated, thanks.

alfonsomthd (approved):

djgalloway (approved):

PR #1947, opened 2022-01-07
ansible: Add Apache Arrow repo to Ubuntu buildhosts

djgalloway:
Couple notes:
  - I moved the GPG key tasks before package installation tasks because we need the GPG key in place before calling it from the repo file
  - The additional [arch=$arch] parameter is to avoid an apt warning about the repo not supporting i386


djgalloway:
This is getting taken care of in install-deps.sh instead.

PR #1948, opened 2022-01-07
ceph-perf-pull-requests: update libarchive

chrisphoffman:
Ensure version of libarchive is installed due to cmake dependency: https://access.redhat.com/solutions/6068431

Error message encountered:
cmake3: symbol lookup error: cmake3: undefined symbol: archive_write_add_filter_zstd

Related: https://github.com/ceph/ceph/pull/42294

Signed-off-by: Christopher Hoffman <choffman@redhat.com>


djgalloway:
Does the version *need* to be 3.3.3?  I try to avoid hardcoding versions if possible.

chrisphoffman:
> Does the version _need_ to be 3.3.3? I try to avoid hardcoding versions if possible.

Very good point. A yum update of the package proved to be sufficient to accomplish this.

djgalloway (approved):

PR #1953, opened 2022-01-21
ceph-{,dev,dev-new}-build:  fix "test for cephadm package" again

dmick:
Try #3 to get the "check for existence of a file" correct.
The construct [ -f <glob> ] will throw a syntax error if <glob>
expands to nothing.  Try for f in glob; do if [ -f "$f" ] instead.


dmick:
I mentioned my frustration with [[ to friends, and they pointed out that what I chose is also buggy (if the file doesn't exist it'll blow up).  This should be better.  Sorry for the churn.

mgfritch:
> I mentioned my frustration with [[ to friends, and they pointed out that what I chose is also buggy (if the file doesn't exist it'll blow up). This should be better. Sorry for the churn.

interesting, the prior form seems to work for me, in which way does this fail?

dmick:
on second thought I think this is probably not necessary assuming the shell doesn't have "nullglob" set, which is its default state.  I think I had set that and then a glob for a nonexistent file *does* cause a problem, because rather than not expanding and just being the literal glob pattern, it's elided, and then -f has no operand.  But that won't be the case without nullglob.

I'll withdraw this PR.

PR #1955, opened 2022-01-25
arrow: remove apache jfrog repositories

cbodley:
arrow is a runtime dependency, so we won't be relying on this apache repo

~~for ubuntu, we'll submodule the arrow library. ubuntu/debian have all of arrow's dependencies~~

~~for centos, we need this 'storage sig' repo for both the arrow library and its missing utf8proc dependency~~

edit: we'll build arrow from submodule until it's available from distros

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -480,9 +480,6 @@ echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu/ $DIST-backports main
 echo "deb [arch=amd64] http://security.ubuntu.com/ubuntu $DIST-security main restricted universe multiverse" >> /etc/apt/sources.list
 env DEBIAN_FRONTEND=noninteractive apt-get update -y -o Acquire::Languages=none -o Acquire::Translation=none || true
 env DEBIAN_FRONTEND=noninteractive apt-get install -y gnupg
-
-curl -L https://dist.apache.org/repos/dist/dev/arrow/KEYS | apt-key add -
-add-apt-repository "deb [arch=amd64] https://apache.jfrog.io/artifactory/arrow/ubuntu $DIST main"
 EOF
     elif [[ "$ARCH" == "arm64" ]]; then
         cat > $hookdir/D04install-updates-repo <<EOF
@@ -1353,8 +1350,6 @@ setup_rpm_build_deps() {
         $SUDO dnf config-manager --add-repo http://apt-mirror.front.sepia.ceph.com/lab-extras/8/
         $SUDO dnf config-manager --setopt=apt-mirror.front.sepia.ceph.com_lab-extras_8_.gpgcheck=0 --save
 
-        $SUDO rpm -import https://dist.apache.org/repos/dist/dev/arrow/KEYS
-        $SUDO dnf config-manager --add-repo="https://apache.jfrog.io/artifactory/arrow/centos/${RELEASE}/x86_64"
     fi
 
     DIR=/tmp/install-deps.$$


cbodley:
i don't suppose there's a way to test this before merging, to make sure this `centos-release-ceph-quincy` package is found?

galsalomon66:
what about this?
https://github.com/ceph/ceph-container/pull/1925/files

scripts/build_utils.sh
```diff
@@ -480,9 +480,6 @@ echo "deb [arch=amd64] http://us.archive.ubuntu.com/ubuntu/ $DIST-backports main
 echo "deb [arch=amd64] http://security.ubuntu.com/ubuntu $DIST-security main restricted universe multiverse" >> /etc/apt/sources.list
 env DEBIAN_FRONTEND=noninteractive apt-get update -y -o Acquire::Languages=none -o Acquire::Translation=none || true
 env DEBIAN_FRONTEND=noninteractive apt-get install -y gnupg
-
-curl -L https://dist.apache.org/repos/dist/dev/arrow/KEYS | apt-key add -
-add-apt-repository "deb [arch=amd64] https://apache.jfrog.io/artifactory/arrow/ubuntu $DIST main"
```
galsalomon66, line 485: :+1:

scripts/build_utils.sh
```diff
@@ -1353,10 +1350,12 @@ setup_rpm_build_deps() {
         $SUDO dnf config-manager --add-repo http://apt-mirror.front.sepia.ceph.com/lab-extras/8/
         $SUDO dnf config-manager --setopt=apt-mirror.front.sepia.ceph.com_lab-extras_8_.gpgcheck=0 --save
 
-        $SUDO rpm -import https://dist.apache.org/repos/dist/dev/arrow/KEYS
```
galsalomon66, line 1356: :+1:

cbodley:
> what about this? https://github.com/ceph/ceph-container/pull/1925/files

hmm thanks. i don't understand how ceph-container fits in here. are container builds really not covered by either of ceph's install-deps.sh or ceph-build's build_utils.sh?

galsalomon66:
> > what about this? https://github.com/ceph/ceph-container/pull/1925/files
> 
> hmm thanks. i don't understand how ceph-container fits in here. are container builds really not covered by either of ceph's install-deps.sh or ceph-build's build_utils.sh?

it seems not.
without this change, the container(centos) build had failed. 

reviewing the email-threads, 
I found out that the following change was also required 
https://github.com/ceph/ceph-cm-ansible/pull/654
for the test-nodes (teuthology)

djgalloway:
> are container builds really not covered by either of ceph's install-deps.sh or ceph-build's build_utils.sh?

They are not.  Separate set of scripts.  (I know...)

I think you want to install it here: https://github.com/ceph/ceph-container/blob/master/ceph-releases/ALL/centos/8/daemon/__DOCKERFILE_INSTALL__

djgalloway:
@cbodley `centos-release-ceph-pacific` exists but not `centos-release-ceph-quincy`

cbodley:
> `centos-release-ceph-pacific` exists but not `centos-release-ceph-quincy`

ok, thanks. @kalebskeithley do you know what the gap is here?

kalebskeithley:
> > `centos-release-ceph-pacific` exists but not `centos-release-ceph-quincy`
> 
> ok, thanks. @kalebskeithley do you know what the gap is here?

that I haven't asked the powers that be to build+release centos-release-ceph-quincy yet. :-)

But I thought you wanted a version agnostic solution because teuthology doesn't know what version it's building And I had suggested that you install centos-release-ceph instead, which will give you the latest available (which would be centos-release-ceph-quincy once it's available.)

I'll have the PTB build -quincy.

cbodley:
okay, using `centos-release-ceph` does sound ideal for all of ceph's build infrastructure. i guess my question is, can we actually ship something on centos 8 that relies on the stuff from -quincy? even with the submoduled arrow, we would still need its utf8proc dependency. lack of this utf8proc-devel package has prevented me from testing the submodule stuff against centos

cbodley:
@djgalloway okay, @kalebskeithley confirmed that the `centos-release-ceph-pacific` repo _does_ contain the arrow-related dependencies we need for centos 8. i updated this pr to use `centos-release-ceph`, which will pull that in for now, and eventually include `centos-release-ceph-quincy` when it's available

so i think this PR is ready for merge, what do you think?

djgalloway:
> so i think this PR is ready for merge, what do you think?

Sure.  I'll merge on Monday.

djgalloway (approved):

cbodley:
i followed up on the ceph-container and ceph-cm-ansible repos too:
https://github.com/ceph/ceph-container/pull/1990
https://github.com/ceph/ceph-cm-ansible/pull/667

PR #1960, opened 2022-02-01
build_utils.sh: CentOS 9 Stream support for get_rpm_dist

djgalloway:

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -257,9 +257,11 @@ get_rpm_dist() {
     # use in constructing chacra urls for rpm distros
 
     LSB_RELEASE=/usr/bin/lsb_release
-    [ ! -x $LSB_RELEASE ] && echo unknown && exit
-
-    ID=`$LSB_RELEASE --short --id`
+    if [ ! -x $LSB_RELEASE ]; then
+        source /etc/os-release
+    else
+        ID=`$LSB_RELEASE --short --id`
+    fi
 
     case $ID in
     RedHatEnterpriseServer)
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -257,18 +257,24 @@ get_rpm_dist() {
...
         DIST=rhel$RELEASE
         DISTRO=rhel
         ;;
-    CentOS|CentOSStream)
-        RELEASE=`$LSB_RELEASE --short --release | cut -d. -f1`
+    CentOS|CentOSStream|centos)
+        if [ "$VERSION_ID" == "9" ]; then
+            RELEASE="9"
+        else
+            RELEASE=`$LSB_RELEASE --short --release | cut -d. -f1`
```
guits, line 276: is the binary `lsb_release` installed by default on CentOS stream 8?
Just asking, because I had to install the package `redhat-lsb-core` in a CentOS stream 8 VM in order to get that binary.

I see the condition above (L260) checks whether it is present and it sources `/etc/os-release` if not.
If we are running a CentOS stream 8 OS without the binary `lsb_release`, this line will throw an error right?

in the case where we are sure the binary will be available, is `| cut -d. -f1` really needed? would `lsb_release --short --release` return something different between `CentOS` / `CentOSStream` / `centos` ?
```
# lsb_release -a
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOSStream
Description:    CentOS Stream release 8
Release:        8
Codename:       n/a
# lsb_release --short --release
8
#
```

djgalloway: > is the binary `lsb_release` installed by default on CentOS stream 8?

[Yes](https://github.com/ceph/ceph-build/blob/master/ansible/examples/slave.yml#L246)

>  If we are running a CentOS stream 8 OS without the binary lsb_release, this line will throw an error right?

Sure, but that was already the case.  There should be no scenario where we're running `scripts/build_utils.sh` against CentOS 8 Stream without `lsb_release`.

> in the case where we are sure the binary will be available, is | cut -d. -f1 really needed

Yes.  It returns the version release number.
```
[dgalloway@vossi03 ~]$ lsb_release  --short --release
8.3.2011
```

guits: > # lsb_release --short --release
> 8
> #

weird, I'm not seeing the same result:

```
# lsb_release --short --release
8
#
```

djgalloway: I don't know what the difference is but some machines spit out the full version string.  It doesn't hurt anything to have the `cut` either way.

guits: indeed

guits (approved):

djgalloway:
https://github.com/ceph/ceph-build/pull/2003/

PR #1965, opened 2022-02-04
dep: delete the network before creating it again

nizamial09:
Need to recreate the network to avoid issues like ssh ing into the
environment.

Last runs with this change in the configuration is a success.
Screenshot from 2022-02-04 14-33-14

Fixes: https://tracker.ceph.com/issues/54131
Signed-off-by: Nizamudeen A <nia@redhat.com

### DIFF
--- a/scripts/dashboard/install-cephadm-e2e-deps.sh
+++ b/scripts/dashboard/install-cephadm-e2e-deps.sh
@@ -86,5 +86,7 @@ sudo chmod +x /usr/local/bin/kcli
 
 # KCLI cleanup function can be found here: https://github.com/ceph/ceph/blob/master/src/pybind/mgr/dashboard/ci/cephadm/start-cluster.sh
 sudo mkdir -p /var/lib/libvirt/images/ceph-dashboard
+kcli delete plan ceph -y || true
+kcli delete network ceph-dashboard -y
 kcli create pool -p /var/lib/libvirt/images/ceph-dashboard ceph-dashboard
 kcli create network -c 192.168.100.0/24 ceph-dashboard


scripts/dashboard/install-cephadm-e2e-deps.sh
```diff
@@ -86,5 +86,7 @@ sudo chmod +x /usr/local/bin/kcli
 
 # KCLI cleanup function can be found here: https://github.com/ceph/ceph/blob/master/src/pybind/mgr/dashboard/ci/cephadm/start-cluster.sh
 sudo mkdir -p /var/lib/libvirt/images/ceph-dashboard
+kcli delete plan ceph -y || true
+kcli delete network ceph-dashboard -y
```
epuertat, line 90: Aren't we performing this clean-up after executing the job? It might not make sense in a local run, but in Jenkins we should leave the node as it was before starting the job (ok, perhaps we can leave the VM disk images stored in the cache to avoid downloading them on every run).

epuertat (approved):

scripts/dashboard/install-cephadm-e2e-deps.sh
nizamial09: We remove the plan after every job but we didn't removed the network in the [cleanup](https://github.com/ceph/ceph/blob/master/src/pybind/mgr/dashboard/ci/cephadm/start-cluster.sh#L9). I added this today in [this PR](https://github.com/ceph/ceph/pull/44825/files). But since most of the machine has the network already its best to recreate the network before starting the test for now.

epuertat: ok, make sense!

PR #1970, opened 2022-02-28
ceph-grafana-trigger: update grafana container image version to 8.2.6

aaSharma14:
update grafana container image version to 8.2.6

Signed-off-by: Aashish Sharma <aasharma@redhat.com>

### DIFF
--- a/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
+++ b/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
@@ -23,7 +23,7 @@
       - string:
           name: CONTAINER_VERSION
           description: "The version tag for the containers; will have -ARCH added to it.  By convention, the version of Grafana"
-          default: "6.7.4"
+          default: "8.2.6"
     axes:
         - axis:
             type: label-expression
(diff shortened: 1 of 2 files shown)


ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
```diff
@@ -23,7 +23,7 @@
       - string:
           name: CONTAINER_VERSION
           description: "The version tag for the containers; will have -ARCH added to it.  By convention, the version of Grafana"
-          default: "6.7.4"
```
epuertat, line 26: Just curious: any specific reason for 8.2.6 and not 8.4.2 or 8.1.2?

epuertat (approved):
LGTM as far as there were were no regressions testing with the existing Dashboards.

BTW, any chance to add a trigger so that this is automatically run:

image

ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
aaSharma14: I was checking this out - https://hub.docker.com/r/grafana/grafana and the latest stable version i found here is 8.2.6, so decided to go forward with this.

epuertat: Makes sense!

epuertat:
Build launched: https://jenkins.ceph.com/view/all/job/ceph-grafana-trigger/2/

PR #1971, opened 2022-02-28
reenable ubuntu's toolchain ppa for c++20 support

cbodley:
we've used this repo in the past to get access to newer compiler versions. reenable it for ubuntu focal for access to gcc 11

note that these changes don't preserve the original behavior against old distros like trusty/xenial. if there is still interest in building against those releases, i can work through these differences:

trusty: PPA 7 -> 11
xenial: PPA 8 -> 11
bionic: does not use PPA (unchanged)

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -632,25 +632,18 @@ setup_pbuilder() {
 }
 
 use_ppa() {
+    # only use ppa on focal for reef+
     case $vers in
-        10.*)
-            # jewel
+        15.*) # o
             use_ppa=false;;
-        11.*)
-            # kraken
+        16.*) # p
             use_ppa=false;;
-        12.*)
-            # luminous
+        17.2*) # q
             use_ppa=false;;
         *)
-            # mimic, nautilus, *
             case $DIST in
-                trusty)
+                focal)
                     use_ppa=true;;
-                xenial)
-                    use_ppa=true;;
-                bionic)
-                    use_ppa=false;;
                 *)
                     use_ppa=false;;
             esac
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -779,10 +779,7 @@ setup_pbuilder_for_old_gcc() {
 setup_pbuilder_for_ppa() {
     local hookdir=$1
     if use_ppa; then
-        local gcc_ver=7
-        if [ "$DIST" = "bionic" ]; then
-            gcc_ver=9
-        fi
+        local gcc_ver=11
         setup_pbuilder_for_new_gcc $hookdir $gcc_ver
     else
         setup_pbuilder_for_old_gcc $hookdir
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -703,21 +703,21 @@ setup_pbuilder_for_new_gcc() {
     local version=$1
     shift
 
-    # need to add the test repo and install gcc-7 after
+    # need to add the test repo and install new gcc after
     # `pbuilder create|update` finishes apt-get instead of using "extrapackages".
-    # otherwise installing gcc-7 will leave us a half-configured build-essential
-    # and gcc-7, and `pbuilder` command will fail. because the `build-essential`
+    # otherwise installing gcc will leave us a half-configured build-essential
+    # and gcc, and `pbuilder` command will fail. because the `build-essential`
     # depends on a certain version of gcc which is upgraded already by the one
     # in test repo.
     if [ "$ARCH" = "arm64" ]; then
-        cat > $hookdir/D05install-gcc-7 <<EOF
+        cat > $hookdir/D05install-new-gcc <<EOF
 echo "deb [lang=none] http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu $DIST main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
 echo "deb [lang=none] http://ports.ubuntu.com/ubuntu-ports $DIST-updates main" >> \
   /etc/apt/sources.list.d/ubuntu-toolchain-r.list
 EOF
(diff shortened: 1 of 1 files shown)


cbodley:
i also see some logic around the gcc toolset in https://github.com/ceph/ceph-build/blob/master/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml#L176-L182. shall i raise `gcc_toolset_ver` to 10 there?

scripts/build_utils.sh
```diff
@@ -651,6 +651,8 @@ use_ppa() {
                     use_ppa=true;;
                 bionic)
                     use_ppa=false;;
+                focal)
+                    use_ppa=true;;
```
tchaikov, line 655: just need to make sure that https://tracker.ceph.com/issues/50218 does not apply to focal.

scripts/build_utils.sh
```diff
@@ -1334,15 +1334,14 @@ get_nr_build_jobs() {
...
         if [ "$ARCH" = x86_64 ]; then
-            $SUDO yum install -y centos-release-scl
+            $SUDO dnf install -y centos-release-scl
         elif [ "$ARCH" = arm64 ]; then
-            $SUDO yum install -y centos-release-scl-rh
-            $SUDO yum-config-manager --disable centos-sclo-rh
-            $SUDO yum-config-manager --enable centos-sclo-rh-testing
+            $SUDO dnf install -y centos-release-scl-rh
+            $SUDO dnf config-manager --disable centos-sclo-rh
+            $SUDO dnf config-manager --enable centos-sclo-rh-testing
```
tchaikov, line 1342: probably we don't need SCL for accessing GTS-11. see https://centos.pkgs.org/8-stream/centos-appstream-x86_64/gcc-toolset-11-build-11.0-0.el8.x86_64.rpm.html

what we need is but to enable the "AppStream" repo. and
```sh
dnf install gcc-toolset-11-build
```

cbodley: ok, thanks. i was being conservative by requiring gcc 10.2, but i can update https://github.com/ceph/ceph/pull/45133 with `BuildRequires: gcc-toolset-11-gcc-c++`

does that mean ubuntu should use_ppa 11 to match?

i'm not familiar with AppStream, and grep didn't find anything under ceph or ceph-build; is that repo expected to be enabled by default, or do i need something extra here?

scripts/build_utils.sh
cbodley: good to know; if that shows up, it will be immediately obvious from teuthology testing

scripts/build_utils.sh
cbodley: ah, i see that ceph.spec.in has `. /opt/rh/gcc-toolset-9/enable` for this, i'll make that unconditional and raise it to 11

tchaikov (approved):
lgtm. need to keep an eye on the build status after merging this change though.

tchaikov:
> i also see some logic around the gcc toolset in https://github.com/ceph/ceph-build/blob/master/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml#L176-L182. shall i raise `gcc_toolset_ver` to 10 there?

would be great if you could. probably should point clang to gcc11 since we have gcc-11 here.

BTW, now that we have llvm-toolset-12 in the official repo, probably it's a good opportunity for us to ditch my personal copr repo and start using clang-12 from AppStream! but don't worry! as it's not in the scope of this change, i will take care of this. clang is being used by the ceph-pull-requests jenkins job for a better compile test for the tree.

i will implement this change this week end.

cbodley:
note that octopus doesn't build cleanly with gcc11, as discussed in backport pr https://github.com/ceph/ceph/pull/45206#issuecomment-1057094145. i'm not sure whether this affects pacific yet

we might need to hold off on this until we either resolve those build issues, or use an earlier version for octopus

cbodley:
> i'm not sure whether this affects pacific yet

yes, pacific PRs are also failing in spdk on nodes that enabled gcc11 - ex. https://github.com/ceph/ceph/pull/45245

cbodley:
> yes, pacific PRs are also failing in spdk on nodes that enabled gcc11 - ex. [ceph/ceph#45245](https://github.com/ceph/ceph/pull/45245)

i updated `use_ppa()` so that it hopefully only applies to 'reef' and later, so we shouldn't see these failures against octopus/pacific/quincy

tchaikov:
@cbodley can we drop the "do-not-merge" label? i think it's ready to go.

scripts/build_utils.sh
```diff
@@ -632,25 +632,18 @@ setup_pbuilder() {
...
-            # jewel
+        15.*) # o
             use_ppa=false;;
-        11.*)
-            # kraken
+        16.*) # p
             use_ppa=false;;
-        12.*)
-            # luminous
+        17.*) # q
```
tchaikov, line 641:
```suggestion
        17.2*) # q
```

as reef is not tagged yet. the commits in main are all named something like `v17.2.0-267-g8bae27e49f9`,

```console
$ git describe --tag
v17.2.0-267-g8bae27e49f9
```

and now that quincy is released, we are not likely to build for 17.1.x anymore. so `17.2*` is a safe choice.

tchaikov:
after applying the proposed change, i updated the job of `ceph-dev-new-build` . and the tree built fine: https://shaman.ceph.com/builds/ceph/wip-cxx20-fixes-kefu-1/377ef0a88af30c85a507c9d7fa4d62e5fc14b4ef/

cbodley:
addressed review comment and removed the dnm

cbodley:
the use of `WITH_STATIC_LIBSTDCXX` in ubuntu builds is causing radosgw to crash on startup: https://tracker.ceph.com/issues/57050

cbodley:
looks like we can resolve it with https://github.com/ceph/ceph/pull/47504

PR #1972, opened 2022-03-02
dashboard-e2e-job: get mgr logs as an artifact

nizamial09:
archive the logs as an artifact in both pr and nightly jobs of
dashboard frontend e2es.

Screenshot 2022-03-02 143840

### DIFF
--- a/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
+++ b/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
@@ -89,3 +89,8 @@
     publishers:
       - email:
           recipients: ceph-qa@ceph.io
+
+      - archive:
+          artifacts: 'build/out/mgr.*.log'
+          allow-empty: true
+          latest-only: false
--- a/ansible/examples/master.yml
+++ b/ansible/examples/master.yml
@@ -28,6 +28,7 @@
       - 'cobertura'
       - 'code-coverage-api'
       - 'command-launcher'
+      - 'compress-artifacts'
       - 'conditional-buildstep'
       - 'configuration-as-code'
       - 'copyartifact'
--- a/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
+++ b/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
@@ -94,3 +94,9 @@
               <applitoolsApiKey/>
             </com.applitools.jenkins.ApplitoolsBuildWrapper>
       - ansicolor
+
+    publishers:
+      - archive:
+          artifacts: 'build/out/mgr.*.log'
+          allow-empty: true
+          latest-only: false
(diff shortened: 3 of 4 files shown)


nizamial09:
sample run: https://jenkins.ceph.com/job/ceph-dashboard-pull-requests/10139/

ceph-api-nightly/config/definitions/ceph-api-nightly.yml
```diff
@@ -89,3 +89,8 @@
     publishers:
       - email:
           recipients: ceph-qa@ceph.io
+
+      - archive:
+          artifacts: 'build/out/mgr.x.log'
```
epuertat, line 94: What about this?
```suggestion
          artifacts: 'build/out/mgr.*.log'
```
And perhaps in the future we'd need to add every log... But let's wait till then.

epuertat (approved):
LGTM!

Shouldn't we add this to ceph-api ones too (the cephadm would be harder since the logs are inside VMs, so we would need to make those volumes accessible from the outside)?

BTW @djgalloway along with this PR it makes sense to install/enable the [artifact compression](https://plugins.jenkins.io/compress-artifacts/) (text logs are easily compressible).

nizamial09:
> we add this to ceph-api ones

will add it too. 

> make those volumes accessible from the outside

or at the end of the test we can right the mgr container log on a file and point that file on the artifact?

djgalloway:
> BTW @djgalloway along with this PR it makes sense to install/enable the [artifact compression](https://plugins.jenkins.io/compress-artifacts/) (text logs are easily compressible).

That should get added here: https://github.com/ceph/ceph-build/blob/master/ansible/examples/master.yml#L12.

Let's go ahead and add that then configure that plugin in the JJB config in this PR.

nizamial09:
> Let's go ahead and add that then configure that plugin in the JJB config in this PR.

@djgalloway I added it to the master.yml file but I am not sure how to configure the plugin in JJB. Could you help me with that? I couldn't find a documentation on how to configure the compress-artifacts.

epuertat:
> So I installed it but the plugin compresses ALL artifacts for ALL jobs. It can't be enabled per-job. Is it possible to compress the logs in the job before archiving them?

@djgalloway What's the downside of it? When you click on those links, does the browser uncompress those or you have to download+open+etc.? 😞

What file system are we using for these nodes? Can't we have transparent compression at file system level? Call me mad, but can't we use [Ceph/S3](https://plugins.jenkins.io/artifact-manager-s3/) for that (I've heard it supports compression :P)?

djgalloway:
> @djgalloway What's the downside of it? When you click on those links, does the browser uncompress those or you have to download+open+etc.? disappointed

I guess not.  It's been enabled for most of today and I can click filenames in the Web UI and download them no problem.

epuertat:
@djgalloway if we all are ok with compressing artifacts, could we get this merged plz? Thank you!

PR #1973, opened 2022-03-05
ceph-perf-pull-requests: use newer clang

tchaikov:


ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -166,12 +166,12 @@
           . /etc/os-release || ID=ubuntu
           case $ID in
           debian|ubuntu)
-              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable clang-10
+              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable clang-12
               ;;
           centos|rhel)
-              sudo dnf module disable -y llvm-toolset
-              sudo dnf copr enable -y tchaikov/llvm-toolset-10 centos-stream-8-x86_64
```
djgalloway, line 173: Do we maybe want to make sure these are both disabled?

tchaikov: good point. the repo is always removed. we can even drop that line after, probably 2 months, assuming it's long enough to get rid of the copr repo on every jenkins node.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -166,12 +166,13 @@
...
-              sudo yum install -y python3-pyyaml python3-lxml python3-prettytable clang
+              sudo dnf copr remove tchaikov/llvm-toolset-10 || true
+              sudo dnf module enable -y llvm-toolset
+              sudo dnf install llvm-toolset
```
djgalloway, line 174: Does this need `-y`?

tchaikov: @djgalloway sure, fixed and repushed.

djgalloway (approved):

PR #1980, opened 2022-03-24
scripts: Use bash instead of python to determine is_not_tar

djgalloway:
As we continue to slowly move on from python2, this is a relic that should get future-proofed.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -564,7 +564,7 @@ setup_pbuilder() {
     # ensure that the tgz is valid, otherwise remove it so that it can be recreated
     # again
     pbuild_tar="$basedir/$DIST.tgz"
-    is_not_tar=`python -c "exec 'try: import tarfile;print int(not int(tarfile.is_tarfile(\"$pbuild_tar\")))\nexcept IOError: print 1'"`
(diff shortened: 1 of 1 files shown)


badone (approved):
Thanks for this David. LGTM.

dmick (approved):

badone:
Turns out  this is causing even more failures as the string formatting is incompatible. Created https://github.com/ceph/ceph-build/pull/1981 to revert it if necessary.

scripts/build_utils.sh
```diff
@@ -564,7 +564,7 @@ setup_pbuilder() {
     # ensure that the tgz is valid, otherwise remove it so that it can be recreated
     # again
     pbuild_tar="$basedir/$DIST.tgz"
-    is_not_tar=`python -c "exec 'try: import tarfile;print int(not int(tarfile.is_tarfile(\"$pbuild_tar\")))\nexcept IOError: print 1'"`
+    is_not_tar=`python3 -c "exec 'try: import tarfile;print int(not int(tarfile.is_tarfile(\"$pbuild_tar\")))\nexcept IOError: print 1'"`
```
dmick, line 567: sigh,  print is a function; needs extra parens.

PR #1986, opened 2022-04-05
ceph-grafana: Update quay cred / erase dockerhub cred

djgalloway:

### DIFF
--- a/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
+++ b/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
@@ -51,7 +51,7 @@
           mask-password-params: true
       - credentials-binding:
           - username-password-separated:
-              credential-id: dmick-quay
+              credential-id: ceph-container-quay-io
               username: CONTAINER_REPO_USERNAME
               password: CONTAINER_REPO_PASSWORD
       - build-name:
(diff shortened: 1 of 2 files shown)


epuertat (approved):
LGTM, but this also needs to be updated in `ceph-grafana-trigger/config/definition/ceph-grafana-trigger.yml` too (I don't get why we have 2 jobs for this... I'll look at improving this after the Quincy release is unblocked :S)

epuertat (approved):
👍🏼

djgalloway:
> LGTM, but this also needs to be updated in `ceph-grafana-trigger/config/definition/ceph-grafana-trigger.yml` too (I don't get why we have 2 jobs for this... I'll look at improving this after the Quincy release is unblocked :S)

Fixed.

Because you need both x86 and arm64 containers.  AFAIK, you need a master/trigger job to trigger the separate arch jobs.

epuertat:
> > LGTM, but this also needs to be updated in `ceph-grafana-trigger/config/definition/ceph-grafana-trigger.yml` too (I don't get why we have 2 jobs for this... I'll look at improving this after the Quincy release is unblocked :S)
> 
> Fixed.
> 
> Because you need both x86 and arm64 containers. AFAIK, you need a master/trigger job to trigger the separate arch jobs.

Mmm, I think that the api/e2e nightly jobs are just a single job spec, but with templates + matrix it's able to do multiple builds...

djgalloway:
Yeah but this is for containers.  The trigger job is what makes and pushes the manifest.

PR #2003, opened 2022-05-18
Add Centos Stream 9

bigjust:


bigjust:
The biggest deal here is that we were forced to use /etc/os-release instead of lsb_release. With a couple Ceph PRs merged it was possible to build Ceph on a builder running Centos Stream 9 using these changes.

scripts/build_utils.sh
```diff
@@ -256,40 +256,26 @@ get_rpm_dist() {
...
-    ID=`$LSB_RELEASE --short --id`
+    RELEASE=$VERSION_ID
+    DISTRO=$ID
 
     case $ID in
-    RedHatEnterpriseServer)
-        RELEASE=`$LSB_RELEASE --short --release | cut -d. -f1`
+    rhel)
         DIST=rhel$RELEASE
-        DISTRO=rhel
```
djgalloway, line 268: We still need this function to return `DISTRO`

scripts/build_utils.sh
```diff
@@ -1351,6 +1337,16 @@ setup_rpm_build_deps() {
...
         $SUDO dnf config-manager --setopt=apt-mirror.front.sepia.ceph.com_lab-extras_8_.gpgcheck=0 --save
 
+    elif [ "$RELEASE" = 9 ]; then
+        $SUDO dnf -y copr enable ceph/el9
+        $SUDO dnf config-manager --setopt=copr\:copr.fedorainfracloud.org\:ceph\:el9.gpgcheck=0 --save
+
+        $SUDO dnf -y install epel-next-release
+
+        $SUDO dnf config-manager --add-repo http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/
+        $SUDO dnf config-manager --setopt=mirror.stream.centos.org_9-stream_CRB_x86_64_os_.gpgcheck=0 --save
```
ktdreyer, line 1347: Why do we need to add these repos? I'm also wondering why we disable GPG checking.

scripts/build_utils.sh
bigjust: I set it at line 262 directly after sourcing os-release

djgalloway: Did you test on CentOS, RHEL, and Fedora and make sure it outputs the same string?

You'll need to modify it still for OpenSUSE

```
dgalloway@braggi18:~> grep ID /etc/os-release
ID="opensuse-leap"
```

bigjust: yeah, the others (rhel, centos, fedora) returns the correct lowercase form

ack on changing this, working on it now

scripts/build_utils.sh
bigjust: We need CRB for librdkafka-devel and the el9 copr repo for the thrift-devel 

There was a time when enabling gpgcheck for these two repos caused package installation failures, but that doesn't appear to be the case any more

ktdreyer: Oh, I see now. In that case, you can enable the CRB definition that is already in `/etc/yum.repos.d`. This command works in the `centos:stream9` container:
```
dnf config-manager --set-enabled crb
```

scripts/build_utils.sh
ktdreyer: Seems like this is resolved now?

djgalloway: > Seems like this is resolved now?

Wouldn't https://github.com/ceph/ceph-build/pull/2003/files#diff-1326d7a2d74cace6f00e1cd5d3ad15ead3fc6359eed098d2133fd72c0ec6700bR277 say `opensuse-leap` then?

scripts/build_utils.sh
```diff
@@ -256,40 +256,27 @@ get_rpm_dist() {
...
-            ;;
-        *Enterprise*)
-                DIST=sles$RELEASE
-                DISTRO=sles
-                ;;
-            esac
+    sles)
+        DIST=sles$RELEASE
+        ;;
+    opensuse)
```
ktdreyer, line 277: I pulled `registry.opensuse.org/opensuse/leap:latest`, and here's `/etc/os-release`:

```
cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.3"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.3"
PRETTY_NAME="openSUSE Leap 15.3"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.3"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```

```suggestion
    opensuse-leap)
```

PR #2011, opened 2022-05-23
ceph-pull-requests/build/build: enable WITH_PMEM

CongMinYin:
WITH_PMEM is added in
https://github.com/ceph/ceph/commit/17d2bc3707bb0078e2fa1b4eef31b39804e45135.

Signed-off-by: Yin Congmin <congmin.yin@intel.com>

### DIFF
--- a/ceph-pull-requests/build/build
+++ b/ceph-pull-requests/build/build
@@ -9,6 +9,7 @@ fi
 export NPROC=$(nproc)
 export WITH_SEASTAR=true
 export WITH_RBD_RWL=true
+export WITH_PMEM=true
 if which apt-get > /dev/null ; then
     export WITH_ZBD=true
 fi


CongMinYin:
@tchaikov

CongMinYin:
If this `WITH_PMEM` is not enabled, pmem related libraries will not be installed in test environment.  Since the previous `cmake` did not find pmem related libraries, this environment variable was ignored.  This [PR46260 ](https://github.com/ceph/ceph/pull/46260) needs to find the necessary libraries including `ndctl` and `daxctl`, and the lack of `WITH_PMEM=true` will lead to the failure to install the necessary pmem related libraries, and then `do_cmake` failed.
more detail see https://github.com/ceph/ceph/pull/46260#issuecomment-1147510653

tchaikov (approved):

djgalloway:
This broke our Jenkins builders because we don't build pmem yet `install_libpmem_on_ubuntu` calls `install_pkg_on_ubuntu` which is expecting a repo to be available here: https://shaman.ceph.com/api/repos/pmem/.

Are `libpmem-dev` and `libpmemobj-dev` readily available in Ubuntu's repos?  If so, they should just get `apt install`ed here: https://github.com/ceph/ceph/blob/main/install-deps.sh#L330 intead of using the `install_pkg_on_ubuntu` function.

https://tracker.ceph.com/issues/56025

PR #2015, opened 2022-05-25
Use Jenkins' 'built-in' magic phrase

djgalloway:
https://www.jenkins.io/changelog-stable/#v2.319.1

I don't know if it's a bug or what but giving the controller a "controller" label and instructing the jobs to run on that label only, they would still get stuck on `(pending—Waiting for next available executor on ‘builtin’)`.


zmc (approved):
`built-in` is such a weird term

PR #2020, opened 2022-06-02
kernel: enable CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT

jtlayton:
The kernel builds don't currently include debug symbols. Enable them.


lxbsz (approved):
LGTM.

kernel/build/kernel-config-deb.sh
```diff
@@ -5939,7 +5939,7 @@ CONFIG_DEBUG_BUGVERBOSE=y
 CONFIG_DEBUG_INFO=y
```
idryomov, line 5939: How about this instead to avoid pinning the DWARF version?
```suggestion
CONFIG_DEBUG_INFO=y
CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y
```

idryomov (changes requested):

kernel/build/kernel-config-deb.sh
idryomov: (This is just the addition of `CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y`, `CONFIG_DEBUG_INFO=y` isn't touched -- it shows up this way due to how Github renders "Suggested changes".)

jtlayton: Sure, sounds good.

idryomov (approved):

PR #2030, opened 2022-06-30
dashboard-e2e: change artifact dir to mstart ones

nizamial09:
after #46848, the log dir will be changed because the test is going to
use the mstart instead of vstart

Screenshot from 2022-07-26 23-01-20

### DIFF
--- a/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
+++ b/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
@@ -99,6 +99,6 @@
           recipients: ceph-qa@ceph.io
 
       - archive:
-          artifacts: 'build/out/mgr.*.log'
+          artifacts: 'build/out/*.log, build/run/1/out/*.log, build/run/2/out/*.log'
           allow-empty: true
           latest-only: false
--- a/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
+++ b/ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
@@ -98,7 +98,7 @@
 
     publishers:
       - archive:
-          artifacts: 'build/out/mgr.*.log'
+          artifacts: 'build/out/*.log, build/run/1/out/*.log, build/run/2/out/*.log'
           allow-empty: true
           latest-only: false
 
--- a/ceph-pr-api/config/definitions/ceph-pr-api.yml
+++ b/ceph-pr-api/config/definitions/ceph-pr-api.yml
@@ -85,7 +85,7 @@
                 - shell: "sudo dpkg --configure -a"
 
       - archive:
-          artifacts: 'build/out/mgr.*.log'
+          artifacts: 'build/out/*.log'
           allow-empty: true
           latest-only: false
 


ceph-api-nightly/config/definitions/ceph-api-nightly.yml
```diff
@@ -99,6 +99,6 @@
           recipients: ceph-qa@ceph.io
 
       - archive:
-          artifacts: 'build/out/mgr.*.log'
+          artifacts: 'build/out/mgr.*.log, build/run/1/out/mgr.*.log, build/run/2/out/mgr.*.log'
```
epuertat, line 102: Recently there was an issue in an API Jenkins job where the RGW daemon crashed, so I thought it would be useful to store the RGW logs too. So what about:
```suggestion
          artifacts: 'build/out/mgr.*.log, build/run/1/out/*.log, build/run/2/out/*.log'
```

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
```diff
@@ -98,6 +98,6 @@
 
     publishers:
       - archive:
-          artifacts: 'build/out/mgr.*.log'
+          artifacts: 'build/out/mgr.*.log, build/run/1/out/mgr.*.log, build/run/2/out/mgr.*.log'
```
epuertat, line 101: Same here.

And let's use this PR to add all logs to the API job too

epuertat (approved):

ceph-api-nightly/config/definitions/ceph-api-nightly.yml
nizamial09: done!

ceph-dashboard-pull-requests/config/definitions/ceph-dashboard-pull-requests.yml
nizamial09: done!

nizamial09:
@epuertat 
it'll look like it.
Screenshot from 2022-08-01 15-48-24

but once you go in the path you'll get all the logs.
Screenshot from 2022-08-01 15-49-29

epuertat:
@nizamial09 That definitely looks ugly 🙈 We could move those logs to an inner directory (vstart.sh line 770, `[client]` section):
```diff
-log file = $CEPH_OUT_DIR/\$name.\$pid.log
+log file = $CEPH_OUT_DIR/client/\$name.\$pid.log
```

nizamial09:
> ```diff
> $CEPH_OUT_DIR/client/\$name.\$pid.log
> ```

Changes are being done here: https://github.com/ceph/ceph/pull/47371

nizamial09:
@djgalloway I think this can be merged now.

nizamial09:
> Can you implement something like this please: https://github.com/ceph/ceph-build/pull/2055/files

@djgalloway That's for the api tests right. And I think the current one will be enough. Because as per your implementation `build/out/mgr.foo.log` will be crested and as per the matching specified in the `ceph-pr-api`, it'll be matches against `build/out/*.log`

djgalloway:
> > Can you implement something like this please: https://github.com/ceph/ceph-build/pull/2055/files
> 
> @djgalloway That's for the api tests right. And I think the current one will be enough. Because as per your implementation `build/out/mgr.foo.log` will be crested and as per the matching specified in the `ceph-pr-api`, it'll be matches against `build/out/*.log`

Ah, right, fair enough.  I was looking at the other two jobs.

PR #2032, opened 2022-07-09
build_utils: do not pass -DWITH_SEASTAR=ON for building crimson flavor

tchaikov:

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -834,7 +834,6 @@ ceph_build_args_from_flavor() {
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
         CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_SEASTAR=ON"
         DEB_BUILD_PROFILES="pkg.ceph.crimson"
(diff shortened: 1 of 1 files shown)


scripts/build_utils.sh
```diff
@@ -834,7 +834,6 @@ ceph_build_args_from_flavor() {
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
         CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_SEASTAR=ON"
```
rzarzynski, line 837: In https://github.com/ceph/ceph/blob/main/ceph.spec.in I can't find any conditional append of `-DWITH_SEASTAR=ON` like:

```
%if 0%{with seastar}
    -DWITH_SEASTAR:BOOL=ON \
%endif
```

Nevertheless, dropping `CEPH_EXTRA_CMAKE_ARGS` is a good thing and I offer to make a follow-up for the spec file if we really miss setting the cmake's option.

tchaikov: @rzarzynski see https://github.com/ceph/ceph/pull/47036 . i meant to disable `WITH_JAEGER` also if crimson is to be built.

rzarzynski (approved):

PR #2043, opened 2022-08-02
add WITH_BOOST_VALGRIND to default build flavor

cbodley:
build boost with valgrind annotations enabled so they run cleanly under valgrind for https://tracker.ceph.com/issues/56500

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -827,6 +827,8 @@ ceph_build_args_from_flavor() {
     default)
         CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
         CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+        # build boost with valgrind=on for https://tracker.ceph.com/issues/56500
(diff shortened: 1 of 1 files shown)


cbodley:
cc @yuvalif @tchaikov @djgalloway

i'm working with @mkogan1 to test whether we can observe any overhead from these builds. DNM pending those results

tchaikov (approved):

idryomov:
```
in https://tracker.ceph.com/issues/56500, Mark Kogan showed that
WITH_BOOST_VALGRIND has little to no effect on rgw performance. Mark
Nelson did similar testing with rbd trying to rule out a performance
hit there as well - the results had quite a bit of variation, so it
was hard to rule out completely

to narrow the scope of this investigation, i did a grep of the entire
boost project and confirmed that boost::context and boost::coroutine
are the only two libraries that mention this BOOST_USE_VALGRIND flag.
so i think it's safe to assume that these valgrind builds will have no
effect outside of rgw
```
https://lists.ceph.io/hyperkitty/list/dev@ceph.io/message/ZGP75L3PJYETWKKW62FFPEFD2FY7MXXU/

idryomov (approved):

neha-ojha (approved):

cbodley:
yay, thanks everybody!

PR #2044, opened 2022-08-03
ceph-docs-rtd: Create job to keep releases docs up-to-date

djgalloway:
See https://github.com/ceph/ceph/pull/46917

This Jenkins job will receive webhooks from the github plugin.  It will then check if the main branch is what was updated and if the `doc/releases` dir was updated.

If true, it will trigger a build for each active Ceph release's docs.

https://github.com/ceph/ceph/pull/47442
https://github.com/ceph/ceph/pull/47443
https://github.com/ceph/ceph/pull/47444


ceph-docs-rtd/config/definitions/ceph-docs-rtd.yml
```diff
@@ -0,0 +1,47 @@
+- job:
+    name: ceph-docs-rtd
```
idryomov, line 2: Nit: something like `ceph-doc-releases-rtd` might be more descriptive.

idryomov (approved):

PR #2063, opened 2022-08-15
Stop using lsb for rpm-based distros

djgalloway:
lsb_release was removed in EL9 and we already have functions to read /etc/os-release and return the values we want.


zmc (approved):
👍

PR #2064, opened 2022-08-20
ceph-windows: remove installed packages

tchaikov:
`ceph-windows/run_tests` installs client side dependencies using
`cephadm install ceph-common`, which leaves us with a bunch ceph
packages after the test finishes. and they inteferes with succeeding
tests which, for instance, try to import python modules from test's
building directory.
but they end up with importing the python modules installed in system.

in this change, the installed package and its ceph dependencies are
removed in the cleanup script. the list of packages is collected by
inspecting `/var/log/dpkg.log` on the system where the test failures
were spotted and `ceph/debian/control` 's "ceph-common" section.

### DIFF
--- a/scripts/ceph-windows/cleanup
+++ b/scripts/ceph-windows/cleanup
@@ -18,6 +18,20 @@ if [[ -x $WORKSPACE/cephadm ]]; then
     fi
 fi
 
+# Uninstall packages installed by "cephadm install ceph-common"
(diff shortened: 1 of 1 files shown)


ionutbalutoiu (approved):
LGTM

epuertat:
Thanks a lot, @tchaikov, for taking care of this!

PR #2070, opened 2022-09-14
nfs-ganesha/rpm: Fix tcmalloc allocator setup

dsavineau:
The previous commit installed the gperftools-devel package and set the cmake ALLOCATION variable to tcmalloc but this isn't used at all during the RPM package build via mock.
We also need those changes in the RPM spec file.
This is a temporary solution until the upstream nfs-ganesha RPM spec file will support such change.


guits:
thanks

PR #2077, opened 2022-09-20
dashboard-deps: remove cortx from the apt-sources

nizamial09:
dashboard e2e is constantly failing with this error for many days. So I went for the simple solution here which is to remove the cortx from the apt-sources. What I am not sure is the impact of this change. Will it affect any other tests?

I also did some cleanups around cephadm e2e. Ideally this should be done in the on_cleanup() function in our start-cluster.sh script in the ceph repo (which I'll be doing), but for extra safety I am putting it here too.

### DIFF
--- a/scripts/dashboard/install-e2e-test-deps.sh
+++ b/scripts/dashboard/install-e2e-test-deps.sh
@@ -6,6 +6,8 @@ if [[ ! $(arch) =~ (i386|x86_64|amd64) ]]; then
     exit
 fi
 
+sudo rm -f /etc/apt/sources.list.d/cortx-motr.list
+
 if grep -q  debian /etc/*-release; then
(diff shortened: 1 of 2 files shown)


scripts/dashboard/install-e2e-test-deps.sh
```diff
@@ -6,6 +6,8 @@ if [[ ! $(arch) =~ (i386|x86_64|amd64) ]]; then
     exit
 fi
 
+sudo rm -f /etc/apt/sources.list.d/cortx-motr.list
```
epuertat, line 9: Are we adding this? If not, we shouldn't remove it. And that action should only be performed on a Debian/Ubuntu environment?

epuertat (changes requested):

PR #2079, opened 2022-09-22
ceph-windows: Update `ceph-windows-test` Jenkins job

ionutbalutoiu:
Make the Jenkins `ceph-windows-test` similar to `ceph-windows-pull-requests` job.

However, `ceph-windows-test` is meant to be triggered on-demand, and it exposes more parameters for `win32_build.sh` script, and artifacts collected at the end of the job run.

The new `ceph-windows-test` Jenkins job will have the following new parameters:

* `CEPH_GIT_REPO` - The Ceph git repo used to clone.
* `CEPH_GIT_BRANCH` - The Ceph git branch name used to clone.
* `CEPH_WIN32_BUILD_FLAGS` - Build flags passed to `win32_build.sh` besides the defaults.
* `INCLUDE_USERSPACE_CRASH_DUMPS` - Collect Windows userspace crash dumps with the job artifacts.
* `INCLUDE_CEPH_ZIP` - Collect the `ceph.zip` Windows build with the job artifacts. This is useful for debugging failures locally.

Also, create `scripts/ceph-windows/win32_build` common script. This is used in both Jenkins jobs to cross-build Ceph Windows.

Besides this, we collect artifacts with `trap EXIT` instance. Previously, if the tests failed, the artifacts wouldn't be collected.

Due to the fact that artifacts collected might occupy more space, we limit `artifact-num-to-keep` to `15`.

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>


petrutlucian94 (approved):
LGTM

To summarize, in some cases we need crash dumps and debug symbols in order to investigate test crashes that may not reproduce locally. We definitely don't want to do this for all the PR jobs since that would consume too much disk space.

Instead, we'll use Jenkins parameters to collect those debug artifacts on demand.

PR #2081, opened 2022-09-26
ceph-windows: Use vstart cluster for Ceph Windows tests

ionutbalutoiu:
Previously, we spawned a minimal cluster with `cephadm` just to have
a Ceph API available (without any OSDs configured).

New Ceph Windows tests will be available that require actual Ceph
store, so a fully functional Ceph cluster is needed.

The entire ceph-windows testing is done with libvirt VMs now.
This way, we avoid accidentally tainting the CI machines after
each job run.

The VMs configuration will be:
* Ubuntu VM, used only for `ceph-windows-pull-requests` and `ceph-windows-test` jobs:
  * vCPUs: `$(nproc)` (as many as host allows in order to build Ceph vstart faster)
  * RAM: `32 GB`
* Windows VM, used to run Windows Ceph tests:
  * vCPUs: `8`
  * RAM: `8 GB`

Building and running Ceph vstart doesn't drastically impact the job
running times, since the CI machines have a lot of CPU cores available.

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>


petrutlucian94 (approved):

petrutlucian94 (approved):
lgtm, thanks a lot for taking care of this! not only we get a proper test cluster, but it's also isolated in a libvirt vm, which significantly reduces the chance of leaking artifacts and impacting other jobs.

ionutbalutoiu:
I ran `ceph-windows-test` job against the changes from here, and everything is good.

PR #2090, opened 2022-11-22
ceph-ansible-nightly: Don't search tags endlessly

zmc:
If the reponse is in an unexpected format (e.g. it's an error response), we were looping infinitely; one job got to over 60k pages!

### DIFF
--- a/ceph-ansible-nightly/build/build
+++ b/ceph-ansible-nightly/build/build
@@ -25,6 +25,11 @@ function find_latest_tag {
   page=1
   while response=$(curl -s "https://quay.ceph.io/api/v1/repository/ceph-ci/daemon/tag/?page=${page}&limit=${page_size}"); do
     tag=$(echo "${response}" | jq -r ".tags[] | select((.name | contains(\"stable\")) and (.name | contains(\"${release}-centos-${el}-x86_64\"))) | .name")
+    local code=$?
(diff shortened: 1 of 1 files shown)


zmc:
I didn't bother trying to echo the actual error here, because it'll be visible in the console log immediately above the new message.

ljflores (approved):

ceph-ansible-nightly/build/build
```diff
@@ -25,6 +25,11 @@ function find_latest_tag {
   page=1
   while response=$(curl -s "https://quay.ceph.io/api/v1/repository/ceph-ci/daemon/tag/?page=${page}&limit=${page_size}"); do
     tag=$(echo "${response}" | jq -r ".tags[] | select((.name | contains(\"stable\")) and (.name | contains(\"${release}-centos-${el}-x86_64\"))) | .name")
+    let code=$?
```
dmick, line 28: little concerned about mixing 'let' (explicitly for arithmetic statements) and '[ !=' (which I'm not sure is); reading bash docs

zmc: I think I _meant_ `local=`, but wrote `let=` instead. According to the docs it's fine, but I'm going to change it anyway.

dmick (approved):

PR #2101, opened 2023-01-25
ceph-dev-*: bump pbuilder dist when targeting Windows

petrutlucian94:
Ceph now uses c++ 20, thus requiring gcc 10 or later.

We'll need to update Windows jobs that leverage pbuilder to from Ubuntu Focal (20.04) to Ubuntu Jammy (22.04) to get mingw 8.0 and gcc 10.3.0.

Fixes: https://tracker.ceph.com/issues/58562

### DIFF
--- a/ceph-dev-build/build/build_mingw
+++ b/ceph-dev-build/build/build_mingw
@@ -1,8 +1,7 @@
 #!/bin/bash
 set -ex
 
-# We need Ubuntu Focal to cross-compile Ceph for Windows.
-# Previous versions provide broken Mingw packages.
+# We need Ubuntu Jammy to cross-compile Ceph for Windows.
 # "DIST" will be set to "windows", so we're currently overriding it with
 # a hardcoded value.
 tmp_pbuild_script=$(mktemp /tmp/build_mingw_pbuild.XXXXXX)
@@ -17,8 +16,8 @@ EOF
 chmod a+x $tmp_pbuild_script
 sudo pbuilder execute \
     --bindmounts "$(pwd):/mnt/ceph" \
-    --distribution "focal" \
-    --basetgz $basedir/focal.tgz \
+    --distribution "jammy" \
+    --basetgz $basedir/jammy.tgz \
     -- $tmp_pbuild_script
 rm $tmp_pbuild_script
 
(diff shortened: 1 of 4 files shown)


idryomov:
> instead of Focal, which provides mingw 8.0 and gcc 10.3.0.

Did you mean to say "gcc 9.30" in the commit message?

ceph-dev-build/build/build_mingw
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 set -ex
 
-# We need Ubuntu Focal to cross-compile Ceph for Windows.
-# Previous versions provide broken Mingw packages.
+# We need Ubuntu jammy to cross-compile Ceph for Windows.
```
idryomov, line 4: Nit: I think Jammy is to be capitalized just like Focal used to be ;)

petrutlucian94:
> > instead of Focal, which provides mingw 8.0 and gcc 10.3.0.
> 
> Did you mean to say "gcc 9.3.0" here in the commit message?

nope, mingw 8.0 includes mingw-gcc 10.3.0:

```
$ dpkg -l | grep mingw

ii  gcc-mingw-w64                                               10.3.0-14ubuntu1+24.3                   all          
...
ii  mingw-w64                                                   8.0.0-1                                 all          Development environment targeting 32- and 64-bit Windows

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
```

ceph-dev-build/build/build_mingw
petrutlucian94: We'll eventually have to bump it again, using the same casing makes it easier to find/replace. If you prefer, I can update it.

idryomov:
This is on Jammy but that part of the commit message talks about Focal.

idryomov:
Perhaps reword this to say something like:

We'll need to update Windows jobs that leverage pbuilder to from Ubuntu Focal (20.04) to Ubuntu Jammy (22.04) to get mingw 8.0 and gcc 10.3.0.

petrutlucian94:
Right, it was a bit confusing, thanks for the suggestion.

ceph-dev-build/build/build_mingw
idryomov: It looks like you updated both `setup_mingw` files but neither of `build_mingw` files.  This is really a nit and not blocking by any means but I think it's better to be consistent.  `git grep -i` works pretty well ;)

petrutlucian94: Oh right, my bad. I'll send a new patch.

idryomov (approved):

PR #2111, opened 2023-02-14
ceph-windows: Fix event log artifacts

stefan-chivu:
Previously, the windows event log artifacts were being dumped in evtx format and then converted to txt by individually querying them using wevtutil.  After doing this, the resulting txt logs contained xml entries that omitted the actual event message.

Now, they will be directly dumped as txt using Get-WinEvent and the evtx logs can be obtained using the IncludeEvtxFiles flag.

Signed-off-by: Stefan Chivu <schivu@cloudbasesolutions.com>

### DIFF
--- a/scripts/ceph-windows/collect-event-logs.ps1
+++ b/scripts/ceph-windows/collect-event-logs.ps1
@@ -20,16 +20,13 @@ function DumpEventLogEvtx($path){
     }
 }
 
-function ConvertEvtxDumpToTxt($path){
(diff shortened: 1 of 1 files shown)


petrutlucian94 (approved):
Looks good. Right now, we're getting some XMLs that don't include the actual event message.

The ``get-winevent`` output is much easier to read. Just one downside: it's slower so we'll have to keep an eye on that.

PR #2113, opened 2023-02-15
windows: fix log collection

petrutlucian94:
The "Get-WinEvent" command used to retrieve Windows event log messages can fail if the specified log has no entries.

We're using the "SilentlyContinue" action to avoid erroring out in such cases.

However, the script still terminates abruptly while collecting logs. For this reason, we'll use the "Ignore" error action instead. We'd rather not have test failures just because we failed to retrieve some Windows event log entries.

Unlike "SilentlyContinue", "Ignore" doesn't populate the global $Error variable, which may be used when running the script remotely.

While at it, we're adding some log messages at the end of the "run_tests" and "collect-event-logs.ps1" scripts.

### DIFF
--- a/scripts/ceph-windows/collect-event-logs.ps1
+++ b/scripts/ceph-windows/collect-event-logs.ps1
@@ -5,7 +5,7 @@ param (
     [switch]$CleanupEventLog = $false
 )
 
-$ErrorActionPreference = "Stop"
+$ErrorActionPreference = "Ignore"
 
 function DumpEventLogEvtx($path){
     foreach ($i in (Get-WinEvent -ListLog * |  ? {$_.RecordCount -gt 0 })) {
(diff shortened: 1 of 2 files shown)


scripts/ceph-windows/collect-event-logs.ps1
```diff
@@ -5,7 +5,7 @@ param (
     [switch]$CleanupEventLog = $false
 )
 
-$ErrorActionPreference = "Stop"
+$ErrorActionPreference = "Ignore"
 
 function DumpEventLogEvtx($path){
     foreach ($i in (Get-WinEvent -ListLog * |  ? {$_.RecordCount -gt 0 })) {
```
ionutbalutoiu, line 11: Shouldn't the condition `$_.RecordCount -gt 0` from here make sure that this doesn't happen:
```
The "Get-WinEvent" command used to retrieve Windows event log messages can fail if the specified log has no entries.
```

ionutbalutoiu (approved):
Only a small question, the rest LGTM.

scripts/ceph-windows/collect-event-logs.ps1
petrutlucian94: There's an additional filter by timestamp:

```
Get-WinEvent `
            -ErrorAction "Ignore" `
            -FilterHashtable @{
                LogName=$i.LogName;
                StartTime=$(Get-Date).AddHours(-6)
```

So it could be that there's a record that occurred before the specified timeframe, in which case it errors out.

ionutbalutoiu: Fair enough!

PR #2115, opened 2023-02-16
windows: install dokany

petrutlucian94:
Dokany is required in order to map cephfs on Windows through ceph-dokan.

For this reason, we're updating the Windows image build script to include Dokany.


ionutbalutoiu (approved):
LGTM

PR #2119, opened 2023-03-02
Update releases - add reef, remove octopus and older

jdurgin:


ceph-volume-test/config/definitions/ceph-volume-test.yml
```diff
@@ -33,13 +33,13 @@
           default: "https://github.com/ceph/ceph.git"
 
     builders:
-      # centos8 based jobs for octopus & pacific & quincy & main
+      # centos8 based jobs for pacific & quincy & main
```
neha-ojha, line 36: nit: add reef

neha-ojha (approved):

PR #2120, opened 2023-03-04
windows: install python3-bcrypt dependency

petrutlucian94:
The Windows job is failing because of a missing mgr dependency.

We'll update the setup script, ensuring that python3-bcrypt is installed.


ionutbalutoiu (approved):
LGTM

PR #2122, opened 2023-03-22
ceph-release-rpm: add CentOS 9 build

epuertat:
Builds CentOS 9 build for ceph-release rpm.

### DIFF
--- a/ceph-release-rpm/config/definitions/ceph-release-rpm.yml
+++ b/ceph-release-rpm/config/definitions/ceph-release-rpm.yml
@@ -38,6 +38,7 @@ If this is checked, then the binaries will be built and pushed to chacra even if
           values:
             - centos7
             - centos8
+            - centos9
 
     builders:
       - shell:


dmick:
I think we'll want to do this if/when we release an official build for centos9.  I'm not sure what the plan is; maybe reef?

epuertat:
> I think we'll want to do this if/when we release an official build for centos9. I'm not sure what the plan is; maybe reef?

I thought we already were, or at least I see a [Quincy CentOS 9 repo available at Ceph Downloads](https://download.ceph.com/rpm-17.2.5/el9/noarch/)...

epuertat:
Thanks, @ktdreyer. I'll leave then up to @dmick whether to merge this or not/later.

ktdreyer:
Justin added most of the `centos9` configuration upstream a while ago in https://github.com/ceph/ceph-build/pull/2003 . I guess that we just missed adding this one `centos9` line here.

dmick:
OK.  I don't think our centos9 plan is what you'd call fleshed out, but maybe this unblocks stuff

epuertat:
Thank you, guys! I guess that for 17.2.6 we'll have this RPM first released for EL9.

PR #2123, opened 2023-04-03
ceph-grafana-trigger: update the grafana version

nizamial09:
after https://github.com/ceph/ceph/pull/50845/

### DIFF
--- a/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
+++ b/ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
@@ -23,7 +23,7 @@
       - string:
           name: CONTAINER_VERSION
           description: "The version tag for the containers; will have -ARCH added to it.  By convention, the version of Grafana"
-          default: "8.3.5"
+          default: "9.4.7"
     axes:
         - axis:
             type: label-expression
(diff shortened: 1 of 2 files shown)


ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
```diff
@@ -23,7 +23,7 @@
       - string:
           name: CONTAINER_VERSION
           description: "The version tag for the containers; will have -ARCH added to it.  By convention, the version of Grafana"
-          default: "8.3.5"
+          default: "9.4.7"
```
epuertat, line 26: That's quite a leap! Have we checked that the Dashboards look ok with this version? I checked the [9.0 release notes](https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v9-0/) and don't see any breaking changes, but it won't do any harm to check out.

epuertat (approved):
LGTM

ceph-grafana-trigger/config/definitions/ceph-grafana-trigger.yml
nizamial09: yup, i locally built the image and tested it with ceph-dev.

nizamial09: here is the [image I built](https://quay.io/repository/nizamial09/ceph-grafana?tab=tags).

PR #2125, opened 2023-04-07
Use createrepo_c, and change order of signing operations to avoid invalid repomd.xml sigs

dmick:
I couldn't validate repomd.xml.asc against repomd.xml without swapping these stanzas.

### DIFF
--- a/scripts/sign-rpms
+++ b/scripts/sign-rpms
@@ -70,15 +70,7 @@ for release in "${releases[@]}"; do
             fi
           done
 
-          # now sign the repomd.xml files
-          if [[ $update_repo -eq 1 ]]; then
-            for repomd in `find -name repomd.xml`; do
-              echo "signing repomd: $repomd"
-              gpg --batch --yes --passphrase "$GPG_PASSPHRASE" --detach-sign --armor -u $keyid $repomd
-            done
-          fi
-
-          # finally, update the repo metadata
+          # now, update the repo metadata
           if [[ $update_repo -eq 1 ]]; then
             for directory in $(ls $path/$distro/$distro_version); do
               cd $directory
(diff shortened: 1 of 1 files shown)


ktdreyer:
This looks good to me.

ktdreyer (approved):

ktdreyer:
I'm wondering how this ever worked. Does the script on the old signer VM match this copy in Git?

ktdreyer:
Oh, I guess we always had bad signatures on `repomd.xml`.

For v17.2.5 today, the quincy el9 repository does not have a valid signature. Eg.

```
wget https://download.ceph.com/rpm-quincy/el9/x86_64/repodata/repomd.xml
wget https://download.ceph.com/rpm-quincy/el9/x86_64/repodata/repomd.xml.asc
wget https://download.ceph.com/keys/release.asc
gpg --import release.asc
gpg --verify repomd.xml.asc repomd.xml

gpg: Signature made Mon 17 Oct 2022 07:12:56 PM EDT
gpg:                using RSA key 08B73419AC32B4E966C1A330E84AC2C0460F3994
gpg: BAD signature from "Ceph.com (release key) <security@ceph.com>" [unknown]
```

ktdreyer:
DNF does not validate these signatures by default. Users must set with `repo_gpgcheck=1` in the ceph .repo file in `/etc/yum.repos.d/`. By default the ceph-release RPM only sets `gpgcheck` (for the RPM files themselves), not `repo_gpgcheck`, so I guess it's not something any user would have found easily.

ktdreyer:
I also had to look up the `--compatibility` flag in the createrepo_c source code. This activates two behavior changes:
- Keep all RPM `%changelog` entries in the metadata instead of the final 10. (Not relevant to Ceph upstream because we don't use RPM `%changelog` entries).
- Change the behavior of `--retain-old-md` to keep the last `NUM` of metadata instead of simply keeping all old entries. (Not relevant to Ceph because we don't use `--update`. We publish entirely fresh metadata every time for simplicity.)

dmick:
re --compatibility: yeah, I just used it because it seemed safer to avoid any bad surprises

PR #2126, opened 2023-04-07
sign-rpms: remove GPG_PASSPHRASE

ktdreyer:
We don't need this argument, and I'm not sure it ever worked.

I entered the passphrase (Nitrokey PIN) into `gpg-agent` prior to running this script, and that seems sufficient.

### DIFF
--- a/scripts/sign-rpms
+++ b/scripts/sign-rpms
@@ -34,11 +34,6 @@ distros=( centos rhel )
 # are the same values used by the build system.
 distro_versions=( 7 8 9 )
 
-# To unlock the gpg keys for the current run, it is requested over STDIN as
-# a password and later passed into GPG directly as a variable.
-read -s -p "Key Passphrase: " GPG_PASSPHRASE
-echo
-
 for release in "${releases[@]}"; do
   for distro in  "${distros[@]}"; do
     for distro_version in  "${distro_versions[@]}"; do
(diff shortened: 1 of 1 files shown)


ktdreyer:
Well, maybe this needs more experimentation. I _thought_ the workflow was:

1. David boots the VM
2. David runs a command to unlock _gpg-agent_ once (`echo hi | gpg --clearsign -u security@ceph.com` would bring up the pinentry prompt)
3. Someone does a release, using the already-unlocked `gpg-agent`

I've never tested `--passphrase` with a hardware signing device before.

dmick:
@ktdreyer what do you think is the right path forward with this?

ktdreyer:
Recording the notes from our discussion today here, the path forward would be:

1. _Document_ the correct / expected way to activate `gpg-agent` (someone has to enter the PIN, minimally after every boot. We should not share that PIN widely.)
2. Instead of this giant `echo yes | ...` command, we should experiment with a more modern way to sign RPMs with this hardware signer. I think we can simply run `rpmsign --define "_gpg_name security@example.com" --addsign *.rpm` as [documented in Koji's docs](https://docs.pagure.org/koji/signing/)
3. Merge https://github.com/ceph/merfi/pull/65, so we have the implementation in unit-testable Python, and we don't have to maintain this script any more

PR #2127, opened 2023-04-19
ceph-dev-setup: save only 25 builds (rather than 2 weeks)

dmick:
ceph-dev-setup is using 67G out of the 200G on /var/lib/jenkins, far larger than any other job.  There's no reason to save all these jobs.


dmick:
ping

zmc (approved):

PR #2129, opened 2023-04-27
cephadm-ansible: add el9 testing

guits:
el9 RPMs weren't available so `el9` distribution was left commented out so far. Given that we now have el9 packages, let's add el9 testing.


asm0deuz (approved):
LGTM

PR #2132, opened 2023-05-05
Fix broken jenkins-job-builder job

zmc:
I'm unsure if this is exactly the right way to solve the issues I saw, but here's my rationale:

When the jjb job executes, sometimes it picks adami02, whose system python is 3.6; newer jjb versions appear to require `dataclasses`, which was new in 3.7 - a version that itself is EOL.

Other times, braggi02 gets selected. That host has 3.10, but was failing because of a lack of `wheel`.

### DIFF
--- a/jenkins-job-builder/build/build
+++ b/jenkins-job-builder/build/build
@@ -9,7 +9,7 @@
 set -euxo pipefail
 
 # the following two methods exist in scripts/build_utils.sh
-pkgs=( "jenkins-job-builder>=3.5.0" )
+pkgs=( "dataclasses" "jenkins-job-builder>=3.5.0" )
 TEMPVENV=$(create_venv_dir)
 VENV=${TEMPVENV}/bin
 install_python_packages $TEMPVENV "pkgs[@]" latest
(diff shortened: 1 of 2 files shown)


ljflores (approved):

dmick:
:+1:

guits:
@zmc @dmick I wonder if that changes got well applied in jenkins after it got merged.

I'm still seeing the `dataclasses` error https://2.jenkins.ceph.com/job/jenkins-job-builder/818/console

akraitman:
Hey @zmc I added "dataclasses" to the pkgs varible in the jenkins-job-builder config since it didn't pickup the changs while the job is failling

dmick:
@akraitman I don't understand; one commit in Zack's change was to add dataclasses.  What do you mean you added dataclasses?

akraitman:
@dmick What I meant is that since doing changes to "jenkins-job-builder/build/build" will only take affect when the jenkins-job-builder job is completed sucessfuly and in our case since the job is failing the change Zack did(adding dataclasses) won't be implmented in the job config, so what I did is I added "dataclasses" manully to the job config so from now on it will read the change Zack did in jenkins-job-builder/build/build

dmick:
Thanks, I get it now.  Did you do that for both jenkins hosts?

akraitman:
On the first instance (jenkins.ceph.com) the job didn't fail, it picked up adami01 which has a newer python version

PR #2133, opened 2023-05-20
ceph-dev-build: stop building on centos9/arm64 temporarily

dmick:
2/3 builders are down.  Revert this when they're back


zmc:
We probably don't want this now right?

dmick:
welll....sulcata01 and 02 are still down.  I dunno.  I guess we're not backing up at the moment, but that may be because the centos9 builds are disabled by a manual jenkins-jobs run

PR #2136, opened 2023-05-25
windows: use bluestore instead of memstore by default

petrutlucian94:
We've been using memstore in order to put as much load as possible on the Windows clients in a relatively short timeframe.

However, we've been hitting memstore segfaults:
https://tracker.ceph.com/issues/61416

For this reason, we'll make the OSD backend configurable, defaulting to bluestore for now.

### DIFF
--- a/scripts/ceph-windows/setup_ceph_vstart
+++ b/scripts/ceph-windows/setup_ceph_vstart
@@ -6,7 +6,10 @@ if [[ -z $UBUNTU_SSH_USER ]]; then echo "ERROR: The UBUNTU_SSH_USER env variable
 if [[ -z $UBUNTU_VM_IP ]]; then echo "ERROR: The UBUNTU_VM_IP env variable is not set"; exit 1; fi
 
 export VSTART_DIR="$WORKSPACE/ceph_vstart"
-export VSTART_MEMSTORE_BYTES="5368709120"  # 5GB
+
+USE_MEMSTORE="${USE_MEMSTORE:-no}"
+VSTART_MEMSTORE_BYTES="5368709120"  # 5GB
+VSTART_BLUESTORE_BYTES="5368709120"  # 5GB
 
 export SSH_USER=$UBUNTU_SSH_USER
 export SSH_ADDRESS=$UBUNTU_VM_IP
@@ -43,6 +46,12 @@ time rsync_cmd $WORKSPACE/ceph ${VSTART_DIR}/build-ceph-vstart.sh ${UBUNTU_SSH_U
 time SSH_TIMEOUT=1h ssh_exec ./build-ceph-vstart.sh
 ssh_exec sudo apt-get install -y python3-prettytable
 
+if [[ "$USE_MEMSTORE" == "yes" ]]; then
+    OBJECTSTORE_ARGS="--memstore -o memstore_device_bytes=$VSTART_MEMSTORE_BYTES"
+else
+    OBJECTSTORE_ARGS="--bluestore -o bluestore_block_size=$VSTART_BLUESTORE_BYTES"
+fi
+
 #
 # Run Ceph vstart
 #
@@ -51,7 +60,7 @@ mkdir -p \$HOME/ceph-vstart/out
 
 cd ~/ceph/build
 VSTART_DEST=\$HOME/ceph-vstart ../src/vstart.sh \
-    -n --memstore -o "memstore_device_bytes=$VSTART_MEMSTORE_BYTES" \
+    -n $OBJECTSTORE_ARGS \
     --without-dashboard -i "$UBUNTU_VM_IP" \
     2>&1 | tee \$HOME/ceph-vstart/vstart.log
 


scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -43,6 +45,14 @@ time rsync_cmd $WORKSPACE/ceph ${VSTART_DIR}/build-ceph-vstart.sh ${UBUNTU_SSH_U
 time SSH_TIMEOUT=1h ssh_exec ./build-ceph-vstart.sh
 ssh_exec sudo apt-get install -y python3-prettytable
 
+memstore_args=""
+bluestore_args=""
+if [ $USE_MEMSTORE == "yes" ]; then
+    memstore_args="--memstore -o memstore_device_bytes=$VSTART_MEMSTORE_BYTES"
+else
+    bluestore_args="-o bluestore_block_size=$VSTART_BLUESTORE_BYTES"
```
idryomov, line 53: I'd make this a single `objectstore_args` variable and pass `--bluestore` here for consistency.

petrutlucian94: Updated, thanks for the suggestion.

scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -6,7 +6,9 @@ if [[ -z $UBUNTU_SSH_USER ]]; then echo "ERROR: The UBUNTU_SSH_USER env variable
 if [[ -z $UBUNTU_VM_IP ]]; then echo "ERROR: The UBUNTU_VM_IP env variable is not set"; exit 1; fi
 
 export VSTART_DIR="$WORKSPACE/ceph_vstart"
+export USE_MEMSTORE=${USE_MEMSTORE:-no}
 export VSTART_MEMSTORE_BYTES="5368709120"  # 5GB
+export VSTART_BLUESTORE_BYTES="5368709120"  # 5GB
```
idryomov, line 11: Do any of these variables need to be exported?

scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -43,6 +45,13 @@ time rsync_cmd $WORKSPACE/ceph ${VSTART_DIR}/build-ceph-vstart.sh ${UBUNTU_SSH_U
 time SSH_TIMEOUT=1h ssh_exec ./build-ceph-vstart.sh
 ssh_exec sudo apt-get install -y python3-prettytable
 
+objectstore_args=""
```
idryomov, line 48: Is this line needed?  Regardless, I'd use the upper case for `objectstore_args` since all other variables are upper case here.

scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -43,6 +45,13 @@ time rsync_cmd $WORKSPACE/ceph ${VSTART_DIR}/build-ceph-vstart.sh ${UBUNTU_SSH_U
...
+if [ $USE_MEMSTORE == "yes" ]; then
```
idryomov, line 49: Since `USE_MEMSTORE` is user input, it's best to quote (and use `[[` instead of `[`).

scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -51,7 +60,7 @@ mkdir -p \$HOME/ceph-vstart/out
 
 cd ~/ceph/build
 VSTART_DEST=\$HOME/ceph-vstart ../src/vstart.sh \
-    -n --memstore -o "memstore_device_bytes=$VSTART_MEMSTORE_BYTES" \
+    -n $objectstore_args \
```
idryomov, line 63: Quoting here wouldn't hurt too.

scripts/ceph-windows/setup_ceph_vstart
petrutlucian94: not really, I'll drop the export keyword.

scripts/ceph-windows/setup_ceph_vstart
petrutlucian94: We can't quote it here since we're passing multiple arguments, e.g. ``--bluestore -o bluestore_block_size=$VSTART_BLUESTORE_BYTES``

idryomov: Argh, yeah -- it could be done with an array but it's not worth it here.

scripts/ceph-windows/setup_ceph_vstart
```diff
@@ -6,7 +6,10 @@ if [[ -z $UBUNTU_SSH_USER ]]; then echo "ERROR: The UBUNTU_SSH_USER env variable
 if [[ -z $UBUNTU_VM_IP ]]; then echo "ERROR: The UBUNTU_VM_IP env variable is not set"; exit 1; fi
 
 export VSTART_DIR="$WORKSPACE/ceph_vstart"
-export VSTART_MEMSTORE_BYTES="5368709120"  # 5GB
+export USE_MEMSTORE=${USE_MEMSTORE:-no}
```
idryomov, line 9: Does `USE_MEMSTORE` need to be exported?

petrutlucian94: Whops, thanks.

idryomov (approved):

PR #2138, opened 2023-06-07
Add ramen-dr job

rakeshgm:
This is an initial attempt to run ramendr/ramen e2e tests which should trigger for every PR. 
This PR will only run unit tests for now. 
Ramen Repo: https://github.com/RamenDR/ramen.git

### DIFF
--- a/ramen-pull-requests/config/definations/ramen-pull-requests.yaml
+++ b/ramen-pull-requests/config/definations/ramen-pull-requests.yaml
@@ -1,0 +1,52 @@
+- job:
+    name: ramen-pull-requests
+    node: huge
+    project-type: freestyle
+    defaults: global
+    concurrent: true
+    display-name: "ramen: Pull Requests"
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          num-to-keep: 30
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: ${ramen-repo}
+
+    parameters:
+      - string:
+          name: branch
+          default: origin/ramen-pr-ci
+
+      - string:
+          name: repo
+          default: https://github.com/rakeshgm/ramen/
+
+    triggers:
+      - github-pull-request:
+          trigger-phrase: "jenkins test.*|jenkins retest.*"
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: false
+          auto-close-on-fail: false
+
+    scm:
+      - git:
+          url: ${repo}
+          branches:
+            - ${branch}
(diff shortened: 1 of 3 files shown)


ramen-pull-requests/config/definations/ramen-pull-requests.yaml
```diff
@@ -0,0 +1,47 @@
...
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          num-to-keep: 30
+          artifact-days-to-keep: -1
+          artifact-num-to-keep: -1
+      - github:
+          url: https://github.com/rakeshgm/ramen/
```
rakeshgm, line 19: I have added my fork for testing purposes

ramen-pull-requests/config/definations/ramen-pull-requests.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - github-pull-request:
+          trigger-phrase: "jenkins test.*|jenkins retest.*"
+          only-trigger-phrase: false
+          github-hooks: true
+          permit-all: false
+          auto-close-on-fail: false
+
+    scm:
+      - git:
+          url: https://github.com/rakeshgm/ramen.git
```
rakeshgm, line 36: added my fork for testing purposes

dmick:
Good start.  You'll need to clean up anything you install, too.  A few notes:

* is installing go from a tarball really the right way to install go?  Are there not distro packages?

* if you make the git repo a variable, you can give a default of your workspace for now, and then just change it when you're ready with the job

* have you tested build/build on a clean installed image (container or vm, maybe)?

rakeshgm:
so installation from Tarball is what the official document says. [doc](https://go.dev/doc/installl). 
I will make the remaining changes

dmick (commented):
What triggers this build?

rakeshgm:
> What triggers this build?

mostly pull-requests to the repo.

dmick (changes requested):
I pulled the branch locally and tested pushing it to the server.  Noticed a few minor things:

1) name the file .yml like all the others (rather than .yaml)
2) Title the pull request "Add ramen-pull-requests" (which is the name of the actual job)
3) change the commit message to describe what the job does and for what purpose; the fact that it's the initial change is obvious from the git history, so note what ramen is and what role it serves in ceph

dmick:
I pushed the job to the jenkins server by hand.  It failed because it ran on a non-dnf system; you'll need to add labels to select the right sort of build host.  I suspect you want something like either 'centos8' or 'centos9'.  Also, I doubt it needs to be "huge".  Does/will this need testing on both x86_64 and arm64?

Here's the build log: https://jenkins.ceph.com/job/ramen-pull-requests/1/consoleFull

PR #2142, opened 2023-06-26
ceph-dev-trigger/config/definitions: don't build crimson for pacific

ljflores:
Crimson changes will not be backported to pacific, so it doesn't make sense to keep building it.


zmc (approved):
Excellent, this will save a lot of resources!

PR #2147, opened 2023-07-15
ansible: enhance library/jenkins_node to allow node update

dmick:
As part of enumerating jenkins node labels in ansible/inventory, the custom ansible module 'jenkins' works better if it can update existing nodes (for example, to change labels).  Add the functionality to update nodes.  (This was a little sticky because the Jenkins API for nodes only uses XML descriptions, so there was data conversion to do, and I decided it would be best to use the xmltodict module, not a part of python but present as an OS package in all the distributions I looked at.)

Another minor enhancement: make failures in the jenkins_node module a little more clear/complete.

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -119,6 +119,7 @@
           - python3-pip
           - python3-venv
           - python3-virtualenv
+          - python3-xmltodict
       when:
         - ansible_os_family == "Debian"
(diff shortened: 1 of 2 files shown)


dmick:
@zmc can you have a look please.

zmc (approved):
Apologies, I had gotten most of the way through this and... I don't know what happened. It looks good to me!

PR #2156, opened 2023-08-07
ceph-container-flake8: fix test

guits:
Looks like this job has never worked.
The file should be bind-mounted in /code, not in / There's no need to pass arguments, the entrypoint automatically scan /code and run flake8 against any present files.


asm0deuz (approved):
LGTM

PR #2159, opened 2023-08-16
scripts: lower concurrency for many-core machines

ljflores:
adami nodes have 96 processing units (nproc). Based on the old script logic, we were compiling 90 build jobs with 96 processing units. This combination (96/90) ends up causing many instances of memory overconsumption on adami nodes.

With this new logic, we take into account that jobs take up more memory than we expect. This will make it so adami nodes will compile 67 build jobs on 96 processing units, which will hopefully avoid so many instances of memory overconsumption. (Total memory on adami nodes is generally ~270036 MiB, so 270036 / 4000 = 67)

braggi nodes do not have this problem; they have 48 processing units. It has been working so far to compile 48 build jobs with 48 processing units on the braggi nodes, and this new logic will not change that.

Fixes: https://tracker.ceph.com/issues/57296

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -1409,11 +1409,19 @@ teardown_vagrant_tests() {
 }
 
 get_nr_build_jobs() {
-    # assume each compiling job takes 2200 MiB memory on average
+    # assume each compiling job takes 3000 MiB memory on average when nproc <= 50
+    # otherwise, assume 4000 MiB when nproc > 50
+    # See https://tracker.ceph.com/issues/57296
     local nproc=$(nproc)
-    local max_build_jobs=$(vmstat --stats --unit m | \
+    if [[ $nproc -gt 50 ]]; then
+        local max_build_jobs=$(vmstat --stats --unit m | \
+                               grep 'total memory' | \
+                               awk '{print int($1/4000)}')
+    else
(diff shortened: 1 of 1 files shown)


idryomov (approved):
LGTM as a (probably not so temporary :smile:) workaround.

For an actual fix, e.g. to be able to take advantage of additional CPUs on adami nodes, we need to partially serialize the build so that steps/targets that are known to be OOM killed most often are run in relative isolation.  Linking would probably be at the top of the list there.

Breaking up huge and/or template-heavy translation units into smaller pieces would help too.

ljflores:
> LGTM as a (probably not so temporary 😄) workaround.
> 
> For an actual fix, e.g. to be able to take advantage of additional CPUs on adami nodes, we need to partially serialize the build so that steps/targets that are known to be OOM killed most often are run in relative isolation. Linking would probably be at the top of the list there.
> 
> Breaking up huge and/or template-heavy translation units into smaller pieces would help too.

Agreed. This will require some analysis into which steps are most problematic. For now, we can use this workaround since at least one build (many times more) per day are affected by this. I will bring this to the Infrastructure meeting or the #sepia chat to discuss further on identifying problematic steps.

dmick (approved):

ljflores:
Okay we'll try it!

PR #2161, opened 2023-08-28
nfs-ganesha: pin the daily build to V5.5

tserlin:
Instead of just building the tip of whatever is in the "next" branch, let's pin the daily nfs-ganesha build to the latest stable tag: V5.5.

### DIFF
--- a/nfs-ganesha/config/definitions/nfs-ganesha.yml
+++ b/nfs-ganesha/config/definitions/nfs-ganesha.yml
@@ -38,7 +38,7 @@
       - string:
           name: NFS_GANESHA_BRANCH
           description: "The git branch (or tag) to build"
-          default: "next"
+          default: "V5.5"
 
       - string:
           name: NFS_GANESHA_DEBIAN_BRANCH


tserlin:
Laura Flores mentioned to me that the infrastructure team wanted use a stable nfs-ganesha tag for the Shaman builds.

ljflores (approved):

ljflores:
Thanks @tserlin!

ljflores:
Think we should change https://github.com/ceph/ceph-container/blob/main/ceph-releases/main/centos/__GANESHA_REPO__ in the same PR?

ljflores:
> Think we should change [https://github.com/ceph/ceph-container/blob/main/ceph-releases/main/centos/**GANESHA_REPO**](https://github.com/ceph/ceph-container/blob/main/ceph-releases/main/centos/__GANESHA_REPO__) in the same PR?

Disregard this, it's in a separate repo so combining in one PR wouldn't make sense.

ljflores:
AFAICT, the step remaining is:

1. Updating https://github.com/ceph/ceph-container/blob/main/ceph-releases/main/centos/__GANESHA_REPO__ so it points to v5.5 instead of next

ljflores:
Looks like we're good to go with step 1:
https://shaman.ceph.com/builds/nfs-ganesha/V5.5/

tserlin:
Thanks, Laura. I'm going to merge.

PR #2166, opened 2023-09-29
project config: adding rook orchestrator e2e testing job

rkachach:
Adding rook orchestrator e2e testing job

Fixes: https://tracker.ceph.com/issues/63019

Testing:

I've been testing this job using my local fork and it seems to be stable so far:

https://jenkins.ceph.com/job/ceph-orchestrator-rook-e2e/

### DIFF
--- a/ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
+++ b/ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
@@ -40,6 +40,9 @@
             - luminous
             - mimic
             - nautilus
+            - pacific
+            - quincy
+            - octopus
           trigger-phrase: 'jenkins test rook e2e'
           skip-build-phrase: '^jenkins do not test.*'
           only-trigger-phrase: false
--- a/ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
+++ b/ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
@@ -40,7 +40,7 @@
             - luminous
             - mimic
             - nautilus
-          trigger-phrase: 'jenkins test rook orchestrator'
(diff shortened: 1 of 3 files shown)


nizamial09 (approved):

ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
```diff
@@ -0,0 +1,96 @@
...
+    triggers:
+      - github-pull-request:
+          cancel-builds-on-update: true
+          allow-whitelist-orgs-as-admins: true
+          org-list:
+            - ceph
+          white-list-labels:
+            - orchestrator
+            - rook
+          black-list-target-branches:
```
nizamial09, line 39: for which releases (or branches) are we going to backport the ceph PR: https://github.com/ceph/ceph/pull/54056

Because, as it is, the script is not existing in any other branches apart from main. We can backport it for reef as well. But if this won't get backported to other release like quincy, pacific, octopus.. then we should blacklist those branches here so that this won't run on the PRs raised against those branches

rkachach: I think we should back-port to reef at least. This way we can run e2e testing when backproting other PRs, WDYT?

nizamial09: yup, sounds good to me!

ceph-rook-e2e/config/definitions/ceph-orchestrator-rook-e2e.yml
```diff
@@ -0,0 +1,98 @@
...
+            - luminous
+            - mimic
+            - nautilus
+            - pacific
+            - quincy
```
nizamial09, line 44:
```suggestion
            - quincy
            - octopus
```

PR #2167, opened 2023-10-03
ceph-dev-new-setup: enable debug options for dev builds

batrick:
Note: it has been this way since at least 35e1a715. It's difficult to tell when
or even if ceph was ever properly built with debugging configurations for QA as
there are corresponding changes in ceph with the switch to cmake which makes
this challenging to evaluate.
    
It's likely that it was wrongly assumed that cmake would set the build type to
Debug because the ".git" directory would be present. This is not the case
because the "make-dist" script (executed below) creates a git tarball that is
used for the actual untar/build. See also:
    
https://github.com/ceph/ceph/pull/53800

### DIFF
--- a/ceph-dev-new-setup/build/build
+++ b/ceph-dev-new-setup/build/build
@@ -64,6 +64,28 @@ else
     echo "forcing."
 fi
 
+# This is a dev release, enable some debug cmake configs. Note: it has been
+# this way since at least 35e1a715. It's difficult to tell when or even if ceph
+# was ever properly built with debugging configurations for QA as there are
+# corresponding changes in ceph with the switch to cmake which makes this
+# challenging to evaluate.
+#
+# It's likely that it was wrongly assumed that cmake would set the build type
+# to Debug because the ".git" directory would be present. This is not the case
+# because the "make-dist" script (executed below) creates a git tarball that is
+# used for the actual untar/build. See also:
+#
+# https://github.com/ceph/ceph/pull/53800
+#
+# Addendum and possibly temporary restriction: only enable these for branches
+# ending in "-debug".
+if [[ "$BRANCH" == *-debug ]]; then
+  CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug -DWITH_CEPH_DEBUG_MUTEX=ON"
+  printf 'Added debug cmake configs to branch %s. CEPH_EXTRA_CMAKE_ARGS: %s\n' "$BRANCH" "$CEPH_EXTRA_CMAKE_ARGS"
+else
+  printf 'No cmake debug options added to branch %s.\n' "$BRANCH"
+fi
+
 ceph_build_args_from_flavor ${FLAVOR}
 
 mkdir -p release


ljflores (approved):

dmick:
Could you elaborate on the comment, please?  1) the .git dir is "inaccessible" how?  2) how did this stop setting an env var?

batrick:
@dmick PTAL

batrick:
cc @idryomov

batrick:
> Could you elaborate on the comment, please? 1) the .git dir is "inaccessible" how? 2) how did this stop setting an env var?
> 
> Edit: Oh, I see what you mean; actual dev builds (on dev boxes) use the presence of .git to set it, but that of course doesn't apply to "proper builds". It does seem like a mistake, but, given that's where were are, I'm concerned that this will end up being a big change to Jenkins CI dev builds if they're not currently enabling debug:
> 
>     1. much more time to build each build
> 
>     2. much more space used to hold debug packages, on jenkins build spaces and on chacra
> 
>     3. potentially-new failures introduced by assertions etc. at runtime
> 
>     4. potential slowdown of runtime for testing due to included debug code (and probably performance testing would need to know about it and maybe not tolerate it?..)
> 
> 
> I'm not saying any of those are dealbreakers, just that this might not be free and may cause followon churn for a while. I wonder if it might be wise to do something like enabling it with a build variable until we get a handle on just how much of an issue the four above are

Yes, there could be unexpected fallout. I think as a reasonable compromise, let's restrict this to branches ending in "debug". I don't know of anyone currently doing tests ending with that name so this should be opt-in. I can do initial testing to see what may happen.

ceph-dev-new-setup/build/build
```diff
@@ -64,6 +64,28 @@ else
...
+#
+# It's likely that it was wrongly assumed that cmake would set the build type
+# to Debug because the ".git" directory would be present. This is not the case
+# because the "make-dist" script (executed below) creates a git tarball that is
+# used for the actual untar/build. See also:
+#
+# https://github.com/ceph/ceph/pull/53800
+#
+# Addendum and possibly temporary restriction: only enable these for branches
+# ending in "debug". The
```
idryomov, line 81: stray ` The`?

ceph-dev-new-setup/build/build
```diff
@@ -64,6 +64,28 @@ else
...
+if [[ "$BRANCH" == *debug ]]; then
```
idryomov, line 82: I would suggest `-debug` suffix -- a bit more restrictive but easier to spot.

ceph-dev-new-setup/build/build
```diff
@@ -64,6 +64,28 @@ else
...
+  CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug -DWITH_CEPH_DEBUG_MUTEX=ON"
```
idryomov, line 83: Just noting that `WITH_CEPH_DEBUG_MUTEX=ON` should be implied in `CMAKE_BUILD_TYPE=Debug`.  We rely on that in a typical vstart and other cases:

https://github.com/ceph/ceph/blob/0deb52e4242c7c2c0ba2181f5759930d66637618/CMakeLists.txt#L361-L362

ceph-dev-new-setup/build/build
batrick: Thanks, fixed!

ceph-dev-new-setup/build/build
batrick: Thanks, fixed!

ceph-dev-new-setup/build/build
batrick: That's right. I think it's good to set it here too as that implied behavior may change in the future accidentally or otherwise.

ceph-dev-new-setup/build/build
```diff
@@ -64,6 +64,28 @@ else
...
+#
+# It's likely that it was wrongly assumed that cmake would set the build type
+# to Debug because the ".git" directory would be present. This is not the case
+# because the "make-dist" script (executed below) creates a git tarball that is
+# used for the actual untar/build. See also:
+#
+# https://github.com/ceph/ceph/pull/53800
+#
+# Addendum and possibly temporary restriction: only enable these for branches
+# ending in "debug".
```
idryomov, line 81:
```suggestion
# ending in "-debug".
```

idryomov (commented):
Otherwise LGTM

ceph-dev-new-setup/build/build
batrick: fixed

idryomov (approved):

dmick (approved):
OK by me.  I'm not sure how to raise awareness about this but it'd be nice to collect success stories to inform the eventual decision whether to make this global.

batrick:
So here's a debug build:

https://shaman.ceph.com/builds/ceph/wip-batrick-testing-20231006.014828-debug/cfbdc475a5ca4098c0330e42cd978c9fd647e012/default/360987/

It took about 45 minutes. Here's a recent main build:

https://shaman.ceph.com/builds/ceph/main/6c4ed58ab15dbfced664970582384bd0a2fa9734/default/360933/

That took about 55 minutes. The difference is probably explained by compiler optimization work. So this actually speeds up builds.

Artifact size did not seem to change much. It's a mix of larger or smaller:

https://3.chacra.ceph.com/r/ceph/main/6c4ed58ab15dbfced664970582384bd0a2fa9734/centos/9/flavors/default/x86_64/

vs.

https://4.chacra.ceph.com/r/ceph/wip-batrick-testing-20231006.014828-debug/cfbdc475a5ca4098c0330e42cd978c9fd647e012/centos/9/flavors/default/x86_64/

dmick:
chacra3:/opt/binaries/ceph/wip-batrick-testing-20231006.014828-debug/cfbdc475a5ca4098c0330e42cd978c9fd647e012$ du -sh *
2.9G    centos
4.1G    ubuntu

there were differences (centos 9 vs 8, maybe crimson vs not) but the 25% inflation on debug info is about what I'd expect.  Don't know what exactly explained the centos drop, but it's not apples/apples so.  

Anyway, the biggest difference is of course in the debuginfo packages.
chacra4:/opt/binaries/ceph/main/6c4ed58ab15dbfced664970582384bd0a2fa9734# du -sh *
3.9G    centos
3.3G    ubuntu

idryomov:
> okay, so the `-debug` suffix is only available for `main` branch and not release based branches
> 
> i.e. `-pacific-debug` or `-debug-pacific` is not possible, right ?

No, it's not just for `main`.  It should work with for e.g. pacific with `-pacific-debug` syntax.

`-debug-pacific` won't work because `-debug` is matched as a suffix, not as a substring.

PR #2172, opened 2023-10-24
sync-push et al: add -ex to shebang, add staging area to sync-push

dmick:


dmick:
check failure should be fixed by https://github.com/ceph/ceph-build/pull/2173

yuriw (approved):

ljflores (approved):

PR #2177, opened 2023-11-13
Remove teuthology-pull-requests

zmc:
This job represents a small subset of what we test with GitHub Actions. It's also unreliable, for reasons including - but not limited to - the fact that it's not possible to request an agent with a particular OS version.


dmick (commented):
Does the job still have a use if it doesn't trigger on PR?  Should we just remove the job?

zmc:
> Does the job still have a use if it doesn't trigger on PR? Should we just remove the job?

I was writing out a reply to the effect of "maybe we should keep it in case we want it in the future" but then I realized if we ever revived it, it should be a pipeline - so no, I think we can just drop the whole job.

dmick:
retest this please

dmick:
retest this please

dmick:
sigh.  retest this *again*, please

dmick:
retest

dmick (approved):

PR #2183, opened 2024-01-04
Add verbosity to yum-builddep to try to get more info about failure

dmick:
No idea why sometimes the builddep phase fails, and I can't reproduce it on demand; perhaps -v will give a little more of a clue

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -1490,7 +1490,7 @@ setup_rpm_build_deps() {
     # python-rpm-macro we use for identifying the python related dependencies
     $SUDO yum install -y python3-devel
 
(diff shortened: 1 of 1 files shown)


ljflores (approved):

dmick:
manually updated the ceph-dev-new-build job, because apparently just updating build_utils.sh doesn't trigger the update

ljflores:
@dmick perhaps you have to run Jenkins Job Builder https://jenkins.ceph.com/view/all/job/jenkins-job-builder/
Just a guess

dmick:
Aware of that job; that's what runs when ceph-build is changed, generally.  But apparently not for updates to scripts.  The manual process I did is what the jjb job does.

PR #2187, opened 2024-01-17
ceph-pull-requests-arm64: only run make check (arm64) on focal and jammy

rosinL:
As the pr https://github.com/ceph/ceph/pull/40952 use build-profile to manage build deps. The feature `--build-profiles` is supported on focal and above, so let's only run make check (arm64) on focal and jammy.

Fixes: https://tracker.ceph.com/issues/64031

### DIFF
--- a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
+++ b/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
@@ -9,7 +9,7 @@
     concurrent: true
     disabled: false
     name: ceph-pull-requests-arm64
-    node: 'arm64 && !centos7 && !centos8 && !centos9'
+    node: 'arm64 && (jammy || focal) && !centos7 && !centos8 && !centos9'
     parameters:
     - string:
         name: ghprbPullId


ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml
```diff
@@ -9,7 +9,7 @@
     concurrent: true
     disabled: false
     name: ceph-pull-requests-arm64
-    node: 'arm64 && !centos7 && !centos8 && !centos9'
+    node: 'arm64 && (jammy || focal) && !centos7 && !centos8 && !centos9'
```
tchaikov, line 12: presumably, we have some nodes stuck with ubuntu bionic, i think the right fix is to upgrade these ancient installations to the latest LTS, i.e., jammy.

rosinL: yes, I agree, better to do that.

tchaikov:
@dmick hi Dan, what do you think?

dmick:
I'd rather upgrade the hosts, which I'm working on today

tchaikov:
@dmick much appreciated!

ljflores:
@dmick let's close this PR and/or update the tracker once you have the hosts upgraded.

dmick:
Hosts are updated.

PR #2191, opened 2024-02-07
centos8: add yum Configs for  ceph/grpc copr https://copr.fedorainfracloud.org/coprs/ceph/grpc/

baum:
centos8: add yum Configs for CentOS Cloud SIG in order to satisfy new build dependency `grpc-devel`

example of the [failing build](https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=centos8,DIST=centos8,MACHINE_SIZE=gigantic/76580//consoleFull).

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -1453,6 +1453,10 @@ setup_rpm_build_deps() {
             $SUDO yum-config-manager --enable centos-sclo-rh-testing
         fi
     elif [ "$RELEASE" = 8 ]; then
+        # for grpc-devel
+        # See https://copr.fedorainfracloud.org/coprs/ceph/grpc/
+        $SUDO yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+        $SUDO dnf copr enable -y ceph/grpc
         # centos 8.3 changes to lowercase repo names
         $SUDO dnf config-manager --set-enabled PowerTools || \
             $SUDO dnf config-manager --set-enabled powertools


ktdreyer:
Incidentally I've been building el9's grpc version for el8 here: https://copr.fedorainfracloud.org/coprs/ktdreyer/ceph-el8/

We should probably move this to the official Ceph copr https://copr.fedorainfracloud.org/coprs/ceph

ktdreyer:
More to the point - we need to drop el8 support instead of adding all this work

baum:
> More to the point - we need to drop el8 support instead of adding all this work

@ktdreyer proposing this PR, since centos8 is used as a base for creating https://quay.ceph.io/repository/ceph-ci/ceph container images.  It is a bit strange though the code here duplicates ceph's [install-deps.sh](https://github.com/ceph/ceph/blob/main/install-deps.sh)

dmick:
Yes, I'm confused about this too.  It sure seems like these belong in install-deps.sh; how else do, say, manual builds work?

baum:
> Yes, I'm confused about this too. It sure seems like these belong in install-deps.sh; how else do, say, manual builds work? Edit: in fact some of the preexisting repo enables _are_ in install-deps.sh as expected. When would this code in setup_rpm_build_deps be necessary? Aren't all builds done after install-deps.sh?
> 
> (and regardless, it definitely needs to be in install-deps.sh for things like make check and manual builds, so this would be incomplete on its face AFAICT)

@dmick thank you for looking into this. For RPM distros the build deps are declared in the RPM spec, install-deps.sh then sets up yum repos according to specific distribution version and applies the RPM spec. It might make sense that code here use install-deps.sh, allowing adding build dependency in one repo, i.e. in ceph. It does appear that currently for centos8 build deps should be added both in in ceph and ceph-build repos.

meanwhile, there is an issue of the container images creation in our dev branch using the new build dependency. One way would be to apply this patch, other better IMHO in the long run, would be to invoke in ceph-build glue scripts ceph's install-deps.sh, removing code duplication. WDYT?

baum:
The latest dev branch centos8 build is [green](https://shaman.ceph.com/builds/ceph/ceph-nvmeof-mon/e7a25f291c1057fd9e71a9f4a9c21cf635c02490/default/380612/). 
Not sure what is the source of inconsistency 🤷‍♂️.

ktdreyer:
OpenStack Wallaby could bring in other packages that override the OS, and that OpenStack release is [close to EOL](https://www.rdoproject.org/rdo/release-cadence/).

I won't block this PR, but it's another example of why we've got to get go from el8 to el9 in Ceph's upstream container.

dmick:
Does Centos 9 Stream contain grpc-devel in its base repos?

dmick:
Can someone bring me up to speed on this...I can't find 14dafef590928bb89728de91784b87b8f9bbf275 in ceph-ci.git, and I'm wondering what the history is here; is nvmeof adding a new dependency on grpc?  Where all has it been added so far (there's more than just enabling the dnf repo)?

baum:
> Can someone bring me up to speed on this...I can't find 14dafef590928bb89728de91784b87b8f9bbf275 in ceph-ci.git, and I'm wondering what the history is here; is nvmeof adding a new dependency on grpc? Where all has it been added so far (there's more than just enabling the dnf repo)?

To which repo does the 14dafef590928bb89728de91784b87b8f9bbf275 SHA refer to, is it ceph-build?

dmick:
I don't know anymore.  I was trying to find out what had been the motivation to add grpc-devel, and what had already been done to add it, because this last step 1) shouldn't be necessary, and 2) even if it is, is not sufficient

baum:
It seems that centos8 build for the dev branch are more stable, last one https://shaman.ceph.com/builds/ceph/ceph-nvmeof-mon/59fc57b90110b66ce213590dd07d16c6b58ec274/default/381309/ is green.

I do not mind closing this PR. Will revisit if the issue occurs again. Thank you 🙏 everyone.

ktdreyer:
@baum and @caroav discussed this on Slack with me. I discovered two things:

1. The nvmeof builds still require `grpc-devel`
2. The reason they were stable is because some Jenkins builders have my old [ktdreyer/ceph-el8 copr](https://copr.fedorainfracloud.org/coprs/ktdreyer/ceph-el8/) enabled, and I was testing building grpc there for the dashboard team.

Instead of enabling `centos-release-openstack-wallaby` and bringing in many more packages, we should enable the [ceph/grpc copr](https://copr.fedorainfracloud.org/coprs/ceph/grpc/) I set up today.

Long-term, we must stop building for el8 and move development to el9.

scripts/build_utils.sh
```diff
@@ -1453,6 +1453,10 @@ setup_rpm_build_deps() {
             $SUDO yum-config-manager --enable centos-sclo-rh-testing
         fi
     elif [ "$RELEASE" = 8 ]; then
+        # for grpc-devel
+        # See https://copr.fedorainfracloud.org/coprs/ceph/grpc/
+        $SUDO yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
ktdreyer, line 1458: @dmick is EPEL 8 already enabled on the builders by default?

baum: @ktdreyer , if already installed, this command would exit with code 0
```shell
[root@1937af58a4d5 /]# yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
Failed to set locale, defaulting to C.UTF-8
Extra Packages for Enterprise Linux 8 - x86_64                                                                  3.4 MB/s |  16 MB     00:04
Last metadata expiration check: 0:00:05 ago on Fri Feb 23 16:30:36 2024.
epel-release-latest-8.noarch.rpm                                                                                4.4 kB/s |  25 kB     00:05
Package epel-release-8-19.el8.noarch is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[root@1937af58a4d5 /]# echo $?
0
```

I'd aim to minimize assumptions about runners.
wdyt?

ktdreyer: I think it will be simpler, faster and easier to maintain if you drop this addition.

dmick:
OK, I get why it needs to be in build_utils now.  "real" builds use pbuilder/rpmbuild, and those have to work from the source tarballs.  Somehow Debian distros never seem to need all this extra repo dancing, except for compilers; the parallel is setup_updates_repo, which installs hooks for pbuilder to do the repo manipulations.

Do either of these supply things that need to be installed at runtime, and if so, do we also need to fix up ceph-container?

dmick:
I guess, regardless, this build change will enable CI building, and can go first.  Someone from the project needs to follow up on packaging dependency declaration and container dependency installation though.

scripts/build_utils.sh
dmick: I don't see any harm in making sure epel8 is available at the place where it must be.  If it's preinstalled on the builders, it's a long way away from here, and it's good to keep like operations together just for documentation reasons if nothing else.

dmick (approved):
a long set of side trips for this one, but now I think I understand.  We'll need the followups.

scripts/build_utils.sh
ktdreyer: Ok, thank you.

PR #2192, opened 2024-02-08
Update builder to work better on centos7: local xml parser, python3 on centos7

dmick:


dmick:
the check failure doesn't look like me, but I'll root-cause it tomorrow

dmick:
retest

zmc (approved):
Approving after a fairly casual review; I don't see any obvious problems, but of course if we run into bugs we may want to add some tests.

dmick:
retest

dmick:
retest

dmick:
retest

dmick:
retest

dmick:
The retest nonsense was about ceph-build-pull-requests being broken because of the breaking change worked around in https://github.com/ceph/ceph-build/pull/2193; with that pin back, this job (and hopefully other ceph-build changes, whose failures have just been ignored) will continue to work until we resolve 6.0.0's compatibility issues or we move away from jjb

PR #2194, opened 2024-02-13
windows: install libvirt

petrutlucian94:
The Windows jobs use Libvirt VMs to build Ceph and run the tests.

The issue is that we're assuming Libvirt to be preinstalled, which isn't necessarily the case.

This change will explicitly install Libvirt.

### DIFF
--- a/scripts/ceph-windows/setup_libvirt
+++ b/scripts/ceph-windows/setup_libvirt
@@ -74,14 +74,9 @@ function get_libvirt_vm_ssh_address() {
 #
 # Setup requirements (if needed)
 #
-if ! which virt-install >/dev/null; then
-    sudo apt-get update
-    sudo apt-get install -y virtinst
-fi
-if ! which cloud-localds >/dev/null; then
-    sudo apt-get update
-    sudo apt-get install -y cloud-image-utils
-fi
+sudo apt-get update
+sudo apt-get install -y libvirt-daemon-system virtinst cloud-image-utils
+
 if ! sudo virsh net-info default &>/dev/null; then
     cat << EOF > $LIBVIRT_DIR/default-net.xml
 <network>


scripts/ceph-windows/setup_libvirt_ubuntu_vm
```diff
@@ -32,6 +32,7 @@ ssh_authorized_keys:
 
 packages_update: true
 packages:
+  - libvirt-daemon-system
```
idryomov, line 35: > The issue is that we're assuming Libvirt to be preinstalled, which isn't necessarily the case.

We are always downloading the same image, from https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img, right?  Did it get updated recently?

petrutlucian94: My bad, this shouldn't go there. It's actually supposed to be installed on the host, updating right now.

petrutlucian94: I think it's just that some Jenkins slaves happened to have it preinstalled, so the scripts assumed libvirt to be already available.

scripts/ceph-windows/setup_libvirt
```diff
@@ -74,12 +74,13 @@ function get_libvirt_vm_ssh_address() {
 #
 # Setup requirements (if needed)
 #
+sudo apt-get update
+sudo apt-get install -y libvirt-daemon-system
```
idryomov, line 78: Is there any downside to also specifying `virtinst` and `cloud-image-utils` here and getting rid of the if statements below entirely?

petrutlucian94: Good point, right away.

idryomov (approved):

petrutlucian94:
Unfortunately we're still hitting this:

```
error: failed to connect to the hypervisor
error: Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory
```

Could be that the libvirtd service isn't running.

PR #2199, opened 2024-02-20
RFC: scripts/build_utils.sh: use clang for crimson builds

athanatos:
Due to https://tracker.ceph.com/issues/64375 and a bug @xxhdx1985126 ran into with std::counting_semaphore, we either need to start building on gcc13 or clang 17.  Due to better coroutine support, my vote is for clang.

@dmick @tchaikov Is this actually the right way to swap the default compiler for crimson shaman builds?

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -870,7 +870,8 @@ ceph_build_args_from_flavor() {
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
+	# use clang for crimson builds due to coroutine support, see https://tracker.ceph.com/issues/64375
+        CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
         DEB_BUILD_PROFILES="pkg.ceph.crimson"
         ;;
     jaeger)


athanatos:
@dmick Is there a way to test this prior to merge, some way to force the jenkins job to use a different branch?

scripts/build_utils.sh
```diff
@@ -870,7 +870,8 @@ ceph_build_args_from_flavor() {
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
-        CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
+	# use clang for crimson builds due to coroutine support, see https://tracker.ceph.com/issues/64375
+        CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++"
```
tchaikov, line 874: this change should suffice if we only want to use clang when building crimson in our CI. if we want to enable user to build crimson using clang, we would have to update `debian/rules` and `ceph.spec.in`. 

another concern is `clang` as a build dependency, shall we add it as a `BuildRequires` in `ceph.spec.in` and `Build-Depends` in `debian/control` ? and ideally with its version specfied, something like

```
BuildRequires:    clang >= 17
```
CentOS AppStream provides clang-17. so it should be fine.

```
Build-Depends:    clang-17 | clang-18
```
please note, we might need to add llvm's third-party repo for accessing newer clang versions, see https://apt.llvm.org

athanatos: Something like https://github.com/ceph/ceph/pull/55825?  Bookworm seems to have clang-16 available.  Not so much with jammy, but we aren't building crimson packages at the moment for ubuntu anyway, so I figure we can punt on that for a few months until 24.04 comes out.

athanatos:
This doesn't actually seem to be necessary -- the centos9 builder uses the pkg.ceph.crimson profile so https://github.com/ceph/ceph/pull/55825 is a better solution.

PR #2203, opened 2024-02-22
ansible: replace include: by include_tasks: and undo pin of ansible version

dmick:


dmick:
I created a modified ceph and ceph-setup, using my branch and running without actual tag operations, to verify that those roles worked.  It's a fairly simple transliteration, so I think it's probably fine for the untested roles (and has been done in the past to lots of our ansible code; this was just a backwater)

zmc (approved):

dmick:
retest

PR #2204, opened 2024-02-22
kernel: enable CONFIG_BPF_SYSCALL for rpm

idryomov:
It's already enabled for deb and is required by podman.


lxbsz (approved):
LGTM.

idryomov:
> with another change that Ilya mentioned over slack

This was https://github.com/ceph/ceph-build/pull/2207.

PR #2206, opened 2024-02-23
ceph-dev-new-trigger: start building quincy packages on jammy

cbodley:
as described in https://lists.ceph.io/hyperkitty/list/dev@ceph.io/thread/ONAWOAE7MPMT7CP6KH7Y4NGWIP5SZ7XR/, lack of jammy packages for quincy is an obstacle to dropping ubuntu focal support on squid

### DIFF
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -118,7 +118,7 @@
                     FORCE=True
                     DISTROS=focal bionic centos8 windows
       # build quincy on:
-      # default: focal centos8 centos9 leap15
+      # default: focal jammy centos8 centos9 leap15
       # crimson: centos8
       - conditional-step:
           condition-kind: regex-match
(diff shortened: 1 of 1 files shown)


dmick (commented):
I think this is likely sufficient

dmick:
I mean it'll make the overall build fail because one distroversion failed, so we should test after merge as much as we can

cbodley:
> I mean it'll make the overall build fail because one distroversion failed, so we should test after merge as much as we can

i just meant that one failure for jammy wouldn't prevent us from scheduling teuthology jobs against the existing supported distros

would you prefer to schedule a test build first? i will follow up on build failures either way

this just needs to be addressed before the next quincy point release and the squid release after. once we switch quincy->squid upgrade jobs to jammy, i'll raise another pr to remove focal builds on main/squid

cbodley:
a friendly reminder that we'll need to release jammy packages for some quincy release before we can release squid

dmick (approved):

cbodley:
i pushed a quincy branch and the jammy build was successful: https://shaman.ceph.com/builds/ceph/wip-quincy-jammy/

PR #2207, opened 2024-02-26
kernel: enable CONFIG_CGROUP_BPF

idryomov:
It's required by podman.


vshankar (approved):
LGTM.

Verified this by running it through fs suite using the testing kernel.

PR #2209, opened 2024-02-29
sync-images: fix the inclusion section

guits:
according to the documentation [1], my understanding is that `!include-raw` should be used instead of `!include-raw-escape`.

[1] https://jenkins-job-builder.readthedocs.io/en/latest/definition.html#inclusion-tags


asm0deuz (approved):
LGTM

PR #2212, opened 2024-03-04
build_utils: use custom playbook for collect_ceph_logs

clwluvw:
If provided, use the custom playbook to collect logs.


clwluvw:
jenkins test this please

PR #2215, opened 2024-03-14
ceph-pull-requests-arm64: add publishers to fix ci result inaccuracy

Svelar:
Fixes: https://tracker.ceph.com/issues/64117


Svelar:
@ljflores @akraitman @dmick Plz have a look.

dmick:
Does this just make arm64 do the same thing as x86_64?

Svelar:
> Does this just make arm64 do the same thing as x86_64?

Correct. Except for the different architectures of running machines, everything else should be the same.

dmick:
This adds the publishers section to analyze build results the same as ceph-pull-requests already does

dmick (approved):

PR #2216, opened 2024-03-16
Add squid to ceph-dev-cron and other jobs (except for ceph-volume)

idryomov:
Based on @jdurgin's commit from last year for reef.

> and other jobs (except for ceph-volume)

I'm clueless about `CEPH_ANSIBLE_BRANCH` and other version bits there, so I'm leaving ceph-volume jobs to @guits.

### DIFF
--- a/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
+++ b/ceph-api-nightly/config/definitions/ceph-api-nightly.yml
@@ -5,7 +5,6 @@
       - squid
       - reef
       - quincy
-      - pacific
     test_suite:
       - backend:
           test_suite_script: run-backend-api-tests.sh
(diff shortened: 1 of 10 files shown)


dmick (approved):
Looks good.  I'll note that ceph-dev-trigger is no longer used (it's marked "disabled"), and ceph-dev-cron does the periodic builds for branches from ceph-ci...but that's OK

dmick:
I've left it to you to merge so that you can do it during your daytime in case something goes wrong

idryomov:
> I've left it to you to merge so that you can do it during your daytime in case something goes wrong

I won't be able to do much and @akraitman is off today, so we will need you anyway ;)

PR #2217, opened 2024-03-18
ceph-pull-requests: remove WITH_ZBD

Svelar:
Fixes: https://tracker.ceph.com/issues/64962

Because of: https://github.com/ceph/ceph/pull/55054

### DIFF
--- a/ceph-pull-requests/build/build
+++ b/ceph-pull-requests/build/build
@@ -9,9 +9,6 @@ fi
 export NPROC=$(nproc)
 export WITH_SEASTAR=true
 export WITH_RBD_RWL=true
-if which apt-get > /dev/null ; then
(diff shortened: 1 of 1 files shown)


dmick:
Looks like it still exists in ceph's src/script/run-make.sh as well, and should probably be removed from there too.  @pereman2 do you agree about both deletions?

pereman2:
@dmick agree!

Svelar:
> Looks like it still exists in ceph's src/script/run-make.sh as well, and should probably be removed from there too. @pereman2 do you agree about both deletions?

Related: https://github.com/ceph/ceph/pull/56283

dmick:
@Svelar please file a followup for run-make.sh

PR #2218, opened 2024-03-18
ceph-pull-requests-arm64: enable RWL mode on arm64

Svelar:
WITH_RBD_RWL variable is added in https://github.com/ceph/ceph/pull/45872.


dmick (approved):
Looks straightforward to me

PR #2220, opened 2024-03-29
ceph-dev-new-trigger: stop building focal packages on main

cbodley:
depends on ~~https://github.com/ceph/ceph/pull/55976~~ (edit: merged) to remove the last dependencies on focal packages for quincy upgrade suites


cbodley:
followup for squid is staged in https://github.com/ceph/ceph-build/pull/2221

dmick (approved):

cbodley:
> Maybe reviewing this will help. I haven't dug into the current configuration yet: http://wiki.front.sepia.ceph.com/doku.php?id=production:jenkins.ceph.com#how_jobs_are_run

i don't think i have an account to access that wiki, but i gather from other prs that ceph-dev-cron is the culprit here and opened https://github.com/ceph/ceph-build/pull/2252 to address

PR #2227, opened 2024-04-21
windows: reduce Linux VM memory

petrutlucian94:
There have been a lot of Windows job failures lately due to qemu errors caused by insufficient Jenkins slave memory:

```
  ERROR    internal error: process exited while connecting to
  monitor: 2024-04-21T11:37:28.922183Z qemu-system-x86_64: cannot
  set up guest memory 'pc.ram': Cannot allocate memory
```

Each Windows job spins up two virtual machines: a Linux one with 64GB of ram and a Windows VM with 8GB of ram. Since the Windows job no longer uses memstore, we can probably reduce the Linux vm memory to 16GB.

### DIFF
--- a/scripts/ceph-windows/setup_libvirt_ubuntu_vm
+++ b/scripts/ceph-windows/setup_libvirt_ubuntu_vm
@@ -7,7 +7,7 @@ if [[ -z $LIBVIRT_DIR ]]; then echo "ERROR: The env variable LIBVIRT_DIR is not 
 export UBUNTU_VM_IMAGE_URL=${UBUNTU_VM_IMAGE_URL:-"https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img"}
 export UBUNTU_VM_NAME=${UBUNTU_VM_NAME:-"ceph-ubuntu-vstart-${JOB_NAME}-${BUILD_ID}"}
 export UBUNTU_VM_VCPUS="$(nproc)"  # use all the CPU cores avaiable on the host
-export UBUNTU_VM_MEMORY="65536"    # 64 GB
+export UBUNTU_VM_MEMORY=${UBUNTU_VM_MEMORY:-"16384"}  # 16 GB
 export UBUNTU_SSH_USER="ubuntu"
 
 #


petrutlucian94:
According to Dan Mick, there are 9 Jenkins slaves with about 16GB of ram, we'll still have to filter these out.

leonid-s-usov (commented):
Sorry, I'm not an expert here

scripts/ceph-windows/setup_libvirt_ubuntu_vm
```diff
@@ -7,7 +7,7 @@ if [[ -z $LIBVIRT_DIR ]]; then echo "ERROR: The env variable LIBVIRT_DIR is not
 export UBUNTU_VM_IMAGE_URL=${UBUNTU_VM_IMAGE_URL:-"https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img"}
 export UBUNTU_VM_NAME=${UBUNTU_VM_NAME:-"ceph-ubuntu-vstart-${JOB_NAME}-${BUILD_ID}"}
 export UBUNTU_VM_VCPUS="$(nproc)"  # use all the CPU cores avaiable on the host
-export UBUNTU_VM_MEMORY="65536"    # 64 GB
+export UBUNTU_VM_MEMORY=${UBUNTU_VM_MEMORY:-"16384"}  # 16 GB
```
idryomov, line 10: > Since the Windows job no longer uses memstore, we can probably reduce the Linux vm memory to 16GB.

It still builds Ceph in that VM before spawning a vstart cluster, right?  For a regular vstart cluster with 3 OSDs, 16G is probably on the high side, but the build is an entirely different story.

We should really figure out a way to build Ceph just once and share the artifacts for jobs that are a must for every PR: `make check`, `api`, `windows`, etc.

petrutlucian94: > It still builds Ceph in that VM before spawning a vstart cluster, right? For a regular vstart cluster with 3 OSDs, 16G is probably on the high side, but the build is an entirely different story.

Exactly, here's the full sequence: https://github.com/ceph/ceph-build/blob/d67fe99f91eea5e50a00844e3b54880619de7c20/ceph-windows-pull-requests/config/definitions/ceph-windows-pull-requests.yml#L70-L79.

> Makes sense, although that would require the artifacts to be stored on a shared environment and I have very limited access and knowledge of the Ceph CI infra. For now, all I can do is adjust the VM memory and/or filter nodes by labels.

We should really figure out a way to build Ceph just once and share the artifacts forbetween jobs that are a must for every PR: make check, api, windows, etc.

Sorry for editing your comment btw, I mean to leave a reply :).

idryomov: > Makes sense, although that would require the artifacts to be stored on a shared environment and I have very limited access and knowledge of the Ceph CI infra. For now, all I can do is adjust the VM memory and/or filter nodes by labels.

Well, I don't see anything wrong with trying to use less memory as is.  Let's see how it goes.

idryomov (approved):

idryomov:
> For a regular vstart cluster with 3 OSDs, 16G is probably on the high side, but the build is an entirely different story.

@petrutlucian94 As suspected, this is very likely an OOM kill:

```
[34/324] Building CXX object CMakeFiles/rocksdb.dir/db/experimental.cc.o
FAILED: CMakeFiles/rocksdb.dir/db/experimental.cc.o 
/usr/bin/g++-11 -DHAVE_PCLMUL -DHAVE_SSE42 -DLZ4 -DOS_LINUX -DROCKSDB_FALLOCATE_PRESENT -DROCKSDB_IOURING_PRESENT -DROCKSDB_LIB_IO_POSIX -DROCKSDB_MALLOC_USABLE_SIZE -DROCKSDB_NO_DYNAMIC_EXTENSION -DROCKSDB_PLATFORM_POSIX -DROCKSDB_PORTABLE -DROCKSDB_PTHREAD_ADAPTIVE_MUTEX -DROCKSDB_RANGESYNC_PRESENT -DROCKSDB_SCHED_GETCPU_PRESENT -DSNAPPY -DZLIB -I/home/ubuntu/ceph/src/rocksdb -I/home/ubuntu/ceph/src/rocksdb/include -isystem /home/ubuntu/ceph/src/rocksdb/third-party/gtest-1.8.1/fused-src -isystem /home/ubuntu/ceph/build/src/liburing/src/include -Wno-deprecated-copy -Wno-pessimizing-move -W -Wextra -Wall -pthread -Wsign-compare -Wshadow -Wno-unused-parameter -Wno-unused-variable -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers -Wno-strict-aliasing -Wno-invalid-offsetof -fno-omit-frame-pointer -momit-leaf-frame-pointer -fno-builtin-memcmp -O3 -DNDEBUG -DROCKSDB_USE_RTTI -fPIC -MD -MT CMakeFiles/rocksdb.dir/db/experimental.cc.o -MF CMakeFiles/rocksdb.dir/db/experimental.cc.o.d -o CMakeFiles/rocksdb.dir/db/experimental.cc.o -c /home/ubuntu/ceph/src/rocksdb/db/experimental.cc
g++-11: fatal error: Killed signal terminated program cc1plus
compilation terminated.
```

Seen here https://jenkins.ceph.com/job/ceph-windows-pull-requests/39006.

I don't think "use all the CPU cores avaiable on the host" and the default ninja concurrency setting is going to be compatible with just 16G of memory allocted to the VM.

idryomov:
Looking at https://jenkins.ceph.com/job/ceph-windows-pull-requests/39004, the Windows job passed on a different Jenkins worker with 

> --vcpus 48 --memory 16384

petrutlucian94:
Indeed, I may have been a bit overoptimistic :) We could go with something like this:

```proc=$(nproc); ninja -j $((proc > ^C ? 16 : proc))``.

Although, I'm a bit concerned that it may speed down the build too much. The alternative would be to increase the vm memory to 32GB.

idryomov:
> We could go with something like this:
> 
> `proc=$(nproc); ninja -j $((proc > 16 ? 16 : proc))`

I think this should be placed in the script that prepares the VM.  There is no point in allocating more CPU to the VM than it would use for the most CPU-intensive task running inside of it.

> 
> Although, I'm a bit concerned that it may slow down the build too much. The alternative would be to increase the vm memory to 32GB.

FWIW we are far more conservative in other jobs:

https://github.com/ceph/ceph-build/blob/ec557a54cdddea9927845bc5d61dd410d6c4db76/scripts/build_utils.sh#L1425-L1449

petrutlucian94:
Thanks for the ``build_utils.sh`` link. We'd get 5 parallel jobs with 16GB of memory based on that, which would probably be very slow.

Since the job completed with 48 vcpus, I hope we'd be safe with 16 vcpus. I sent another patch: https://github.com/ceph/ceph-build/pull/2228

PR #2228, opened 2024-04-22
windows: limit number of Linux VM vcpus

petrutlucian94:
We reduced the amount of memory allocated to Linux VMs used by the Windows PR jobs to 16GB but we're now hitting OOM errors while building Ceph.

To mitigate this, we'll limit the number of parallel jobs and allocated vcpus.


idryomov (approved):

petrutlucian94:
@idryomov almost there, I'll need to increase some command timeouts:

```
+ EXIT_CODE=124
+ [[ 124 -eq 124 ]]
+ echo 'ERROR: ssh command timed out'
ERROR: ssh command timed out
+ return 124

real	60m0.198s
user	0m0.138s
sys	0m0.162s
```

PR #2230, opened 2024-04-26
ceph-dev-new-build: add timestamps to the build log

leonid-s-usov:
A quick change to add timestamps to the execution logs


ChristinaMeno (approved):
Good use of PS!

PR #2116, opened 2023-02-16
ceph-windows: Re-enable Windows event logs collection

ionutbalutoiu:
Cleanup `collect-event-logs.ps1` script:

* Rely on the default `$ErrorActionPreference` value (which is `Continue`).
* Add new function `SanitizeName` that it's used to sanitize the name of the log files.
  The function code existed before, but it was duplicated in the script.
* General PowerShell code cleanup.

The Windows event log collection was failing due `ssh_exec` abruptly
closing the connection, because the default timeout of `30 secs` was exceeded.

This change re-enables the Windows event log collection by increasing the
SSH timeout to `30 minutes`.

Additionally, the `ssh_exec` function is updated to print an appropriate
error message when SSH timeout is exceeded.

Signed-off-by: Ionut Balutoiu <ibalutoiu@cloudbasesolutions.com>

PR #2232, opened 2024-05-02
windows: increase log collection timeout

petrutlucian94:
Some failed Windows jobs are missing the build artifacts and it seems like the log collection function times out.

https://jenkins.ceph.com/job/ceph-windows-pull-requests/39380/

We'll increase the timeout from 30s to 10m.

### DIFF
--- a/scripts/ceph-windows/run_tests
+++ b/scripts/ceph-windows/run_tests
@@ -38,7 +38,7 @@ scp_upload $CEPH_KEYRING /ProgramData/ceph/keyring
 #
 # Setup the Ceph Windows build in the Windows VM
 #
-SSH_TIMEOUT=5m scp_upload $WORKSPACE/ceph.zip /ceph.zip
+scp_upload $WORKSPACE/ceph.zip /ceph.zip
 SSH_TIMEOUT=10m ssh_exec powershell.exe "\$ProgressPreference='SilentlyContinue'; Expand-Archive -Path /ceph.zip -DestinationPath / -Force"
 ssh_exec powershell.exe "Add-MpPreference -ExclusionPath 'C:\ceph'"
 ssh_exec powershell.exe "New-Service -Name ceph-rbd -BinaryPathName 'c:\ceph\rbd-wnbd.exe service'"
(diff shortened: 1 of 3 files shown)


scripts/ceph-windows/run_tests
```diff
@@ -82,7 +82,7 @@ function collect_artifacts() {
     ssh_exec /wnbd/wnbd-client.exe version
     ssh_exec curl.exe --retry-max-time 30 --retry 10 -L -o /workspace/collect-event-logs.ps1 $COLLECT_EVENT_LOGS_SCRIPT_URL
     SSH_TIMEOUT=30m ssh_exec powershell.exe /workspace/collect-event-logs.ps1 -LogDirectory /workspace/eventlogs
-    scp_download /workspace/eventlogs $WORKSPACE/artifacts/client/eventlogs
+    SSH_TIMEOUT=10m scp_download /workspace/eventlogs $WORKSPACE/artifacts/client/eventlogs
```
idryomov, line 85: Are Windows event logs significantly bigger than other logs collected in this function?

It would be nice to:
- extend `scp_upload()` and `scp_download()` the same way as `ssh_exec()` was extended in https://github.com/ceph/ceph-build/pull/2116
- confirm that we are indeed hitting a timeout before proceeding with this PR

petrutlucian94: > Are Windows event logs significantly bigger than other logs collected in this function?

The logs may get larger if there's an error that gets logged repeatedly. I saw that the log collection step failed only when there were failing tests.

> extend scp_upload() and scp_download() the same way as ssh_exec() was extended in ceph-windows: Re-enable Windows event logs collection #2116

Good point, thanks.

> Rather than singling out some call sites, we might want to increase the default timeout in scp_upload() and scp_download(). Currently it's the same as in ssh_exec(), which doesn't seem optimal -- uploading or downloading something is expected to take longer than executing a random command.

Makes sense, I've increased these timeouts to 10m.

idryomov: > Makes sense, I've increased these timeouts to 10m.

Then `SSH_TIMEOUT=10m` on this line isn't needed anymore?  And similarly there are two instances of `SSH_TIMEOUT=5m` specified for `scp_upload()` in other places -- probably should get rid of those as well.

petrutlucian94: I was just being cautious in case the default timeouts get decreased but that probably won't be the case. I'm dropping these right away.

idryomov (approved):

PR #2238, opened 2024-05-15
build_utils.sh: remove pbuilder's --mirror option for Debian

dmick:
The existing mirror is defunct, and hard-stopping bookworm builds. For now, let's just revert to not specifying a mirror (as we have in the past for other repo configurations) and solve any problems that result if they pop up.


yuriw (approved):

ljflores (approved):
Conceptually seems okay. Let's see how things go without a mirror and readjust accordingly.

PR #2239, opened 2024-05-16
build_utils.sh: re: last commit: one must specify a mirror

dmick:
To pbuild Debian on Ubuntu (as we do), one must specify a mirror that contains the Debian environments.  Use the most-generic one.


dmick:
testing in a private build tree

ljflores (approved):

PR #2241, opened 2024-05-16
Allow all the ceph -setup jobs to run on arm64.

dmick:
They only do file operations to build a source tarball and debian source package things, so the architecture isn't important.


dmick:
I've tested this by manually changing the configuration and found no problems with the setup results.

leonid-s-usov (commented):
LGTM, but I don't have enough competence here to approve.

PR #2242, opened 2024-05-17
ceph-pull-requests-arm64: extend timeout to 4h

Svelar:
Because performance of omani series machine are not so well which caused make check often failed, especially when cache hit was low, extend timeout to 4h seems reasonable.

Ref build duration: https://jenkins.ceph.com/job/ceph-pull-requests-arm64/buildTimeTrend

### DIFF
--- a/ceph-pull-requests-arm64/build/build
+++ b/ceph-pull-requests-arm64/build/build
@@ -12,6 +12,6 @@ export CHECK_MAKEOPTS="-j${n_test_jobs}"
 export BUILD_MAKEOPTS="-j${n_build_jobs}"
 export WITH_SEASTAR=true
 export WITH_RBD_RWL=true
-timeout 3h ./run-make-check.sh
+timeout 4h ./run-make-check.sh
 sleep 5
 ps -ef | grep -v jnlp | grep ceph || true


ljflores (approved):

dmick:
I think it's crazy that we have a "make check", required for every build and every PR, that takes over three hours anywhere.  Perhaps it's time to start reconsidering what all is in make check and if it's really necessary to be there (as opposed to in teuthology)

Svelar:
> I think it's crazy that we have a "make check", required for every build and every PR, that takes over three hours anywhere. Perhaps it's time to start reconsidering what all is in make check and if it's really necessary to be there (as opposed to in teuthology)

IMO CI is a common way to check PR's quality, but yes, three even four hours seem like ridiculous. For this PR is to adjust existing issue I observed, looking forward to any solution that can fundamentally change this situation.

tchaikov:
> I think it's crazy that we have a "make check", required for every build and every PR, that takes over three hours anywhere. Perhaps it's time to start reconsidering what all is in make check and if it's really necessary to be there (as opposed to in teuthology)

it is. IIRC, we once reduce the time to around 1 hour. 

but to make the test great again, i guess there some work to do. here is what i can think of

1. to measure the time (and even the cpu / memory resource) consumed by each test.
2. to see if we run them in an efficient paralleled way
3. to see if some of them can be removed or moved into teuthology

dmick (approved):
we really need to cut down make check.  approving, grudgingly

PR #2245, opened 2024-05-25
ceph-build-pull-requests: Drop ansible syntax checks, add retest trigger

dmick:


dmick:
I'm not especially thrilled with this, but the code is just too old and there's limited value in the ansible checks IMO.  As I said in the comment, maybe we can fork and update Mayeu.rabbitmq the way Andrew did years ago for sensu

PR #2253, opened 2024-06-05
Builders need to run java-11-openjdk in order to connect to the master

akraitman:


dmick (approved):
presumably it cleans up older javas, and agent.jar runs OK with v11?

akraitman:
@dmick I just noticed that it was the right place so I changed to be under "universal_rpms"

dmick (approved):

PR #2069, opened 2022-09-12
nfs-ganesha: build with tcmalloc

guits:
since https://github.com/ceph/ceph/pull/47628 any development builds of nfs-ganesha are broken.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

PR #2257, opened 2024-06-15
Disable centos8 kernel and ganesha

dmick:


nfs-ganesha/config/definitions/nfs-ganesha.yml
```diff
@@ -52,8 +52,8 @@
 
       - string:
           name: DISTROS
-          description: "A list of distros to build for. Available options are: bionic, centos8"
-          default: "centos8 bionic"
```
idryomov, line 56: Noting that this has been out of sync with `AVAILABLE_DIST` since https://github.com/ceph/ceph-build/pull/2069.

idryomov (approved):
Ack for kernel

PR #2259, opened 2024-06-30
Fix up quay-pruner

dmick:
A number of issues and performance enhancements.  Primarily it was not finding recent images at all because we'd switched to 'stream8' rather than '8', and so it was not searching for the right tags.  There were almost 20000 saved images.  That's back down to about 450.  Also, there's no 'image_id' attribute for images anymore, so we must use the sha1 digest to identify same images/different tags.

Also added some performance enhancements; removed an inner loop-inside-loop for every image by building a digest-to-image map, snarfing up all the shaman results in one request so that we can do all the comparisons in local memory, and a few more progress messages.

### DIFF
--- a/quay-pruner/build/prune-quay.py
+++ b/quay-pruner/build/prune-quay.py
@@ -66,41 +66,55 @@ def parse_quay_tag(tag):
     return ref, short_sha1, el, arch
 
 
+shaman_data = None
+
 def query_shaman(ref, sha1, el):
 
-    params = {
-        'project': 'ceph',
-        'flavor': 'default',
-        'status': 'ready',
-    }
+    global shaman_data
+
+    error = False
+    if shaman_data is None:
+        params = {
+            'project': 'ceph',
+            'flavor': 'default',
+            'status': 'ready',
+        }
+        try:
+            response = requests.get(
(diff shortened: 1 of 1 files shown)


quay-pruner/build/prune-quay.py
```diff
@@ -66,41 +66,56 @@ def parse_quay_tag(tag):
     return ref, short_sha1, el, arch
 
 
+shaman_data = None
```
zmc, line 69: Would you consider using something like [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)?

zmc (approved):
Memoization seems like it would be helpful but it's not a blocker

quay-pruner/build/prune-quay.py
dmick: I added functools.cache().  It's a bit cleaner and gives me the opportunity to separate the caching and the searching.

dmick:
Removed a few now-unnecessary (and in one case, incorrect in the face of sha1 tag duplication) mechanisms, and used functools.cache() for implementing the shaman data cache

zmc (approved):

PR #2260, opened 2024-07-01
sepia-fog-images: Unlock on failure

zmc:
Tested here: https://jenkins.ceph.com/job/sepia-fog-images/816/console

The tracebacks are ugly but aren't causing problems; their root cause is that the job expects teuthology to _not_ have any fog configuration present.

### DIFF
--- a/sepia-fog-images/build/failure
+++ b/sepia-fog-images/build/failure
@@ -12,6 +12,13 @@ funPowerCycle () {
   fi
 }
 
+# Should we use teuthology-lock to lock systems?
+if [ "$DEFINEDHOSTS" == "" ]; then
+  use_teuthologylock=true
+else
+  use_teuthologylock=false
+fi
+
 # Clone or update teuthology
 if [ ! -d teuthology ]; then
   git clone https://github.com/ceph/teuthology


dmick:
So I hesitate to proceed with this just because I remember that the job intentionally left machines that failed the capture locked for investigation, and I need to understand the whole job (again) to understand when or where that's appropriate or not

zmc:
> So I hesitate to proceed with this just because I remember that the job intentionally left machines that failed the capture locked for investigation, and I need to understand the whole job (again) to understand when or where that's appropriate or not

The trouble is these machines sometimes get left locked for months. I don't think the analysis is happening. Maybe I should add a boolean parameter to the job that can be set to false to leave the nodes locked.

dmick:
Oh I agree.  Someone needs to fix the fog capture job.

zmc:
Right, and this is an attempt to do so; there are currently four smithis still held by the last run of it, which was three days ago

dmick:
No, I mean the actual failure that is causing the machines to remain locked, which is serving as the only reminder/motivation to go fix that

dmick:
this is probably better than never having time to investigate the failures, I guess.

PR #2263, opened 2024-07-04
ceph-volume: update labels

guits:
These labels must be updated to make ceph-volume jobs use `centos9` workers.


asm0deuz (approved):
LGTM

Issue #36, opened 2014-12-05
Wrong URL for repo rpm


t0mmyt:
gen_yum_repo.sh
Line 87: baseurl=http://ceph.com/rpms/$dist/\$basearch

The URL is ceph.cpm/rpm, not rpms.  The repository rpm is broken.

alfredodeza:
gen_yum_repo.sh is no longer used, it still exists in the repo as a reference only.

PR #2269, opened 2024-07-16
adding clang-tidy as a linter tool

Suyashd999:
The Ceph project has a huge codebase, and it faces the risk of containing suboptimal code that could jeopardize storage reliability or induce performance bottlenecks or cause resource inefficiencies. Identifying and rectifying such code issues is important to maintain the integrity and efficiency of the Ceph storage system. clang-tidy, a powerful static analysis tool, offers a systematic approach to uncover critical issues within the codebase and generate comprehensive reports highlighting potential vulnerabilities.

Integrating clang-tidy into the Jenkins CI pipeline of Ceph would be very beneficial for the development, as any new possible vulnerabilities would be identified early on. The goal is to enhance code quality, maintain performance, and ensure the reliability of the storage system by continuously monitoring and addressing potential issues through automated static analysis.

### DIFF
--- a/ceph-pr-clang-tidy/build/build
+++ b/ceph-pr-clang-tidy/build/build
@@ -1,0 +1,1 @@
+sudo apt-get install -y clang-tidy
--- a/ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
+++ b/ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
@@ -1,0 +1,61 @@
+- job:
+    name: ceph-pr-clang-tidy
+    node: small
+    project-type: freestyle
+    defaults: global
+    display-name: 'ceph: Clang-tidy checks'
+    concurrent: true
+    quiet-period: 5
+    block-downstream: false
+    block-upstream: false
+    retry-count: 3
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          artifact-days-to-keep: 15
+      - github:
+          url: https://github.com/ceph/ceph/
+
+  parameters:
+  - string:
+      name: ghprbPullId
+      description: "the GitHub pull id, like '72' in 'ceph/pull/72'"
+      default: origin/main
+
+  scm:
+  - git:
+      url: https://github.com/ceph/ceph.git
+      name: origin
+      branches:
+        - origin/pr/${ghprbPullId}/merge
+      refspec: +refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*
+      skip-tag: true
+      shallow-clone: true
+      honor-refspec: true
+      timeout: 20
+
+  triggers:
+  - github-pull-request:
+      allow-whitelist-orgs-as-admins: true
+      org-list:
+        - ceph
+      trigger-phrase: 'jenkins test clang-tidy'
+      only-trigger-phrase: false
+      github-hooks: true
+      permit-all: true
+      auto-close-on-fail: false
+      status-context: "Signed-off-by"
+      started-status: "checking if bugs exist in file"
+      success-status: "no bugs found"
+      failure-status: "bugs found"
+
+  builders:
+  - shell:
+      !include-raw:
+        - ../../../scripts/build_utils.sh
+        - ../../build/build
+
+  publishers:
+  - junit:
+      results: report.xml
+      allow-empty-results: true


ceph-pr-clang-tidy/build/build
```diff
@@ -0,0 +1 @@
+sudo apt-get install -y clang-tidy
```
dmick, line 1: of course this only works if the build machine is an apt-based system, which it may not be using your current node labels.  It might be more appropriate to install the package inside the pbuilder image in that case, too, although it probably doesn't matter too much; there are ways of adding extra packages to the pbuilder image

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
```diff
@@ -0,0 +1,61 @@
...
+    properties:
+      - build-discarder:
+          days-to-keep: 15
+          artifact-days-to-keep: 15
+      - github:
+          url: https://github.com/ceph/ceph/
+
+  parameters:
+  - string:
+      name: ghprbPullId
```
dmick, line 21: I suspect this is probably provided by the ghprb plugin, and doesn't need to/shouldn't be a settable parameter

ceph-pr-clang-tidy/build/build
```diff
@@ -0,0 +1,35 @@
...
+{
+  echo "Files being checked:"
+  cat "$file_list"
+  echo
+  parallel -m clang-tidy -checks="-*,bugprone-use-after-move" -p "$WORKSPACE/build" {} < "$file_list"
+} | tee "$WORKSPACE/build/$output_file"
+
+sudo chmod +x ../src/script/clang-tidy-to-junit.py
+
+cat "$WORKSPACE/build/$output_file" | ../src/script/clang-tidy-to-junit.py $WORKSPACE/src >"$WORKSPACE/report.xml"
```
Suyashd999, line 35: The file `clang-tidy-to-junit.py` would be added to the `scripts` in a different PR.

dmick: You want the PR to be complete, and not break anything when integrated.  Go ahead and add extra script; I would say, for now, into the build directory of this job, as it's not clear it's of generic use yet.

Suyashd999: Done with adding the script.


I have taken the script from here: https://github.com/PSPDFKit-labs/clang-tidy-to-junit

ceph-pr-clang-tidy/build/build
dmick: We still need an answer for this; does clang-tidy exist for RHEL-like systems?  or do we want to restrict this job to only apt systems?

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
dmick: What's the response to this?

ceph-pr-clang-tidy/build/build
```diff
@@ -0,0 +1,35 @@
...
+{
+  echo "Files being checked:"
+  cat "$file_list"
+  echo
+  parallel -m clang-tidy -checks="-*,bugprone-use-after-move" -p "$WORKSPACE/build" {} < "$file_list"
+} | tee "$WORKSPACE/build/$output_file"
+
+sudo chmod +x clang-tidy-to-junit.py
+
+cat "$WORKSPACE/build/$output_file" | clang-tidy-to-junit.py $WORKSPACE/src >"$WORKSPACE/report.xml"
```
dmick, line 35: unnecessary cat; you can use < to redirect stdin

dmick (changes requested):
Let's resolve the outstanding comments

ceph-pr-clang-tidy/build/build
Suyashd999: It is available but for now lets keep the job to apt systems only

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
Suyashd999: I removed `ghprbPullId` as a settable parameter. However in others Jenkins jobs it is set as a settable parameter.

Reference: 
1. 
https://github.com/ceph/ceph-build/blob/e8d18c7aef15fa0424d096ba73a6e0d701a5215a/ceph-pull-requests-arm64/config/definitions/ceph-pull-requests-arm64.yml#L13-L16

2. 
https://github.com/ceph/ceph-build/blob/e8d18c7aef15fa0424d096ba73a6e0d701a5215a/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml#L137-L140

3. 
https://github.com/ceph/ceph-build/blob/e8d18c7aef15fa0424d096ba73a6e0d701a5215a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml#L27-L30

ceph-pr-clang-tidy/build/build
ronen-fr: How is this accomplished? won't the command be executed (and fail) on all systems?

dmick: The job specifies the labels for the jenkins builder it wants; we'll need to add something like 'jammy' as a label, that should suffice

Suyashd999: Where should I add this 'jammy' label?

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
Suyashd999: @dmick and what about this?

ceph-pr-clang-tidy/build/build
dmick: job.node is the list of labels the builder must satisfy.  You probably want jammy && small

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
dmick: We can leave it in for now; I don't think it's necessary but it might be useful for testing

dmick:
Let's change the name of LICENSE to LICENSE.clang-tidy-to-junit.py to clarify what the license applies to

dmick (changes requested):
address the couple more comments and then we'll set up a test job

ceph-pr-clang-tidy/config/ceph-pr-clang-tidy.yml
Suyashd999: @dmick I have added  `node: jammy && small
`

ceph-pr-clang-tidy/build/build
Suyashd999: > job.node is the list of labels the builder must satisfy. You probably want jammy && small



@dmick I have added `node: jammy && small`

Suyashd999:
> address the couple more comments and then we'll set up a test job

Hi @dmick! I think we are ready for a test job.

Suyashd999:
> > address the couple more comments and then we'll set up a test job
> 
> Hi @dmick! I think we are ready for a test job.
> 
> However there is one problem. Clang-tidy requires `compile_commands.json` file for analysis for which we have to pass `-DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON` to CMAKE. I have create a PR for that: #2270
> 
> This PR will also be need to be merged.
> 
> After this we are good to go.

Hey @dmick can you please check this? Thank you

Suyashd999:
> Maybe it's best to incorporate this as an optional step in the make-check job? make check already builds ceph, and will have a built ceph tree in the workspace (unlike a standalone jenkins job, which would have to redo the build process)

Hi @dmick !
After researching a bit I found out that we can use the Jenkins plugin ['Copy Artifact'](https://plugins.jenkins.io/copyartifact/)  which allows us to copy files and directories from different Jenkins jobs.

What we can do is allow `make check` to finish its execution and start our clang-tidy job. In our job it'll copy all the files from the job `make check` and then start clang-tidy.
I was able to make the copy artifact work in my local system. The `build` job after its successful completion would trigger our clang-tidy job. The clang-tidy job copies all the files from the `build` job in the form of a `tar` file and extracts it.

I think this is the best approach.

Please let me know what you think.

dmick:
That sounds like a good way to address the concerns.  IIRC this involves a change to the make check job to add permissions to allow clang-tidy to use the artifacts.

Suyashd999:
> That sounds like a good way to address the concerns. IIRC this involves a change to the make check job to add permissions to allow clang-tidy to use the artifacts.

Yes you are right. You will also need to install the plugin ['Copy Artifact'](https://plugins.jenkins.io/copyartifact/) Please do so

dmick:
Copy artifact is installed.  It's used in the ceph-build jobs (search for "copyartifact")

Suyashd999:
> Copy artifact is installed. It's used in the ceph-build jobs (search for "copyartifact")
@dmick 
Okay. I'll need to change the `make-check` job to archive the files right? And where would I need to add the code to automatically trigger the clang-tidy job after `make-check` job?

dmick:
Yes, the make check job will both archive the files and trigger the clang-tidy job

Suyashd999:
> Yes, the make check job will both archive the files and trigger the clang-tidy job

Hi @dmick. I have created a PR for archiving Ceph and triggering the job `ceph-pr-clang-tidy`. Can you please take a look at this? https://github.com/ceph/ceph-build/pull/2276 
Thanks!

Suyashd999:
> > Yes, the make check job will both archive the files and trigger the clang-tidy job
> 
> Hi @dmick. I have created a PR for archiving Ceph and triggering the job `ceph-pr-clang-tidy`. Can you please take a look at this? #2276 Thanks!

Hey @dmick! could you please look into this? Thank you!

Suyashd999:
Hi! @dmick can you check this please?

> > Yes, the make check job will both archive the files and trigger the clang-tidy job
> 
> Hi @dmick. I have created a PR for archiving Ceph and triggering the job `ceph-pr-clang-tidy`. Can you please take a look at this? #2276 Thanks!

ceph-pr-clang-tidy/build/build
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+set -ex
+
+# Clone the Ceph repository
+git clone https://github.com/ceph/ceph.git "$WORKSPACE/ceph"
```
yuvalif, line 5: can you do a shallow clone of only the relevant branch?
this would save considerable amount of time

Suyashd999:
@yuvalif done can you please take a look now, Thank you

ceph-pr-clang-tidy/build/build
```diff
@@ -0,0 +1,58 @@
...
+sudo apt-get install -y curl
+sudo apt-get install -y python3-routes
+sudo apt-get install -y clang-tidy parallel
+
+# Configure the build
+./do_cmake.sh -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON
+cd build
+
+# Replace the boost and include directories from the tar file
+tar -xzf "$WORKSPACE/ceph_build.tar.gz" -C "$WORKSPACE"
```
yuvalif, line 19: where is the tar file being generated?
do we control the content of the file and make sure we only add `boost` and `include` dirs are there?

yuvalif:
> @yuvalif done can you please take a look now, Thank you

do you have new time measurements after all of the latest improvements?

ceph-pr-clang-tidy/build/build
Suyashd999: The tar file is generated in the Jenkins job `make check`. Once the job finishes its execution it will zip the `boost` and `include` directories. 

Yes we control the content of tar file, it is being controlled in the pull request: https://github.com/ceph/ceph-build/pull/2276

Suyashd999:
> > @yuvalif done can you please take a look now, Thank you
> 
> do you have new time measurements after all of the latest improvements?

image

@yuvalif 
Apologies for the delay in reply, after various testing, I replicated the Jenkins jobs in my local system and started up our clang-tidy-jenkins job. The job took 24 minutes to complete. 

It was running the clang-tidy check `bugprone-use-after-move` on **ALL** files in the RGW 

The job was divided into various stages to exactly monitor the time taken by each stage.

**1. Copy artifact stage:**
image
Time: **5.1 seconds**

**2. Installing clang-tidy:**
image
Time: **12 seconds**

**3. Shallow clone ceph repo & recursive submodule update:**
image
Time: **7 minutes 20 seconds**

**4. Running `./do_cmake` with export commands:**
image
Time: **16 seconds**

**5. Replace the `boost` and `include` directories from the :`**
image
Time: **4.4 seconds**

**6. Running clang-tidy on ALL RGW files:**
image
Time: **15 minutes**


**Total time: 24 minutes**

@yuvalif so what should I do next

yuvalif:
@Suyashd999 thank you for the detailed report!
I think that we will have to go fo the option of running clang-tidy only on the files modified in the test.
this should reduce the part that takes most time. for testing the run time, you can pick arbitrary 10 RGW files (that should represent an "average" PR).

another question. when you do the "copy artifacts" stage, why can't you copy the actual code base as well?
this may take less time than doing the shallow git clone + submodule update?

Suyashd999:
> another question. when you do the "copy artifacts" stage, why can't you copy the actual code base as well? this may take less time than doing the shallow git clone + submodule update?

@yuvalif 
I tried copying the actual code base in the Jenkins job to save time but there are two issues, the problem arises because of 1. submodules and 2. `compile_commands.json` . 

1. To get `compile_commands.json` we need to run `./do_cmake.sh -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON`,
The first line of `do_cmake.sh` is `git submodule update --init --recursive --progress --recommend-shallow` So we need to wait for the submodules to be updated.

2. The submodules problem can by bypassed by not running `do_cmake` and copying the required `compile_commands.json` file from our `make check` (ceph-pull-requests) job but `compile_commands.json` uses absolute path instead of relative path. So using `compile_commands.json` from other job doesn't work. 

As of now with the configuration of clang-tidy being only run on those files which have been modified is working great. It takes roughly **7-10 minutes** to complete when the number of RGW files vary from 1-10.

It takes 4-5 minutes for the submodules to be updated and 2-4 minutes for the clang-tidy to finish execution. If we are able to discard the submodules step then the job would be done within couple of minutes

yuvalif:
> 1. To get `compile_commands.json` we need to run `./do_cmake.sh -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=ON`,
>    The first line of `do_cmake.sh` is `git submodule update --init --recursive --progress --recommend-shallow` So we need to wait for the submodules to be updated.
> 2. The submodules problem can by bypassed by not running `do_cmake` and copying the required `compile_commands.json` file from our `make check` (ceph-pull-requests) job but `compile_commands.json` uses absolute path instead of relative path. So using `compile_commands.json` from other job doesn't work.


you dont need to run `do_cmake`  to get `compile_commands.json` . inside the `build` directory run `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..`

it does exactly the same, only without the submodules

Suyashd999:
@yuvalif 
Great success :tada: 
image

With your insights I made the changes 

The entire Jenkins clang-tidy job only took **2 min 32
 seconds** :tada: 

The Jenkins job was run on ceph's latest commits 

I have attached the logs, for testing purposes the git diff was modified to work on the last commit in rgw which was https://github.com/ceph/ceph/commit/50fa90b944424fb0eae2d9c6d55064b99be7971b 

Please look into this,
[#36.txt](https://github.com/user-attachments/files/18589300/36.txt)

yuvalif:
> @yuvalif Great success 🎉 image
> 
> With your insights I made the changes
> 
> The entire Jenkins clang-tidy job only took **2 min 32 seconds** 🎉
> 
> The Jenkins job was run on ceph's latest commits
> 
> I have attached the logs, for testing purposes the git diff was modified to work on the last commit in rgw which was [ceph/ceph@50fa90b](https://github.com/ceph/ceph/commit/50fa90b944424fb0eae2d9c6d55064b99be7971b)
> 
> Please look into this, [#36.txt](https://github.com/user-attachments/files/18589300/36.txt)

thanks @Suyashd999 these are great news. this commit was fairly small (one file) but even if the whole process takes less than 10 min (for larget commits) I think we are good.

@dmick do you think these run times make sense? are you good with merging that?

PR #2274, opened 2024-08-19
sepia-fog-images: Work around ansible-galaxy installation problem

dmick:
The mount module is now ansible.posix.mount, and ansible-playbook can't seem to find it when executed from the parent directory. Change to teuthology/ to make it work, in lieu of being able to find any documentation at all about how it ought to be working

### DIFF
--- a/sepia-fog-images/build/build
+++ b/sepia-fog-images/build/build
@@ -186,7 +186,12 @@ while [[ $(echo $remaininghosts | wc -w) != 0 ]]; do
       # Set DHCP back
       ssh ubuntu@store01.front.sepia.ceph.com "sudo /usr/local/sbin/set-next-server.sh $host fog"
       # Prep the host for FOG image capture
+      #
+      # this pushd/popd should not be necessary, but I'm trying to get ansible to find
+      # its galaxy modules in teuthology/.ansible
+      pushd $WORKSPACE/teuthology
       ansible-playbook $WORKSPACE/ceph-cm-ansible/tools/prep-fog-capture.yml -e ansible_ssh_user=ubuntu --limit="$host*"
+      popd
       remaininghosts=${remaininghosts//$host/}
     else
       # This gets noisy


zmc (approved):

zmc:
Ah, I see now the issue you were having Friday. I imagine this is the bit of docs that would have been helpful: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#collections-paths

On my system, running `ansible-galaxy collection list` from a random directory picks up 103 modules; if I run it from an activated teuthology venv, I see only the eight that teuthology requires. If I run it from a teuthology repo root without activating the venv, I see 103+8. If I set `ANSIBLE_COLLECTIONS_PATH` to the repo's path, I also see 103+8.

dmick:
The way the job was running was from the parent of teuthology.  If I activate teh teuthology venv and run "ansible-galaxy collection list" from the parent of teuthology, it says

`"ERROR! - None of the provided paths were usable. Please specify a valid path with --collections-path"
`
which is slightly amazing to me.  If I instead run it from teuthology/, I get the 8 (which would be sufficient).

I don't even understand the documentation above.  It seems to be suffering from some templating errors?...  according to ansible-config, the defaults seem to be:

```
$ ansible-config dump | grep -i collections
COLLECTIONS_ON_ANSIBLE_VERSION_MISMATCH(default) = warning
COLLECTIONS_PATHS(default) = ['/home/jenkins-build/.ansible/collections', '/usr/share/ansible/collections']
COLLECTIONS_SCAN_SYS_PATH(default) = True
GALAXY_COLLECTIONS_PATH_WARNING(default) = True
```
so I guess it's just not handling the virtualenv the way one might like/expect.

If I add ANSIBLE_COLLECTIONS_PATH=$ANSIBLE_COLLECTIONS_PATH:$(pwd)/teuthology/.ansible, the collections list works as expected.

zmc:
Ah, it looks like the confusing(-to-me) behavior re: the working directory is caused by ansible looking for `ansible.cfg` in `$PWD`, and teuthology's does set the collections path. So, an alternative could be `ANSIBLE_CONFIG=/path/to/teuthology.git/ansible.cfg ansible-galaxy collection
 list`

dmick:
OK, that works too.  Closing this in favor of that better solution

PR #2276, opened 2024-08-24
archiving `ceph` directory for the job 'ceph-pr-clang-tidy'

Suyashd999:
The job `ceph-pr-clang-tidy` requires Ceph's build but since the job `ceph-pull-requests` is already building Ceph it is best to archive & copy the directory and work further upon it. It is done to save time and resources required to again build Ceph.

artifact required by: https://github.com/ceph/ceph-build/pull/2269

### DIFF
--- a/ceph-pull-requests/build/build
+++ b/ceph-pull-requests/build/build
@@ -12,3 +12,5 @@ export WITH_RBD_RWL=true
 timeout 3h ./run-make-check.sh
 sleep 5
 ps -ef | grep -v jnlp | grep ceph || true
+
+sudo tar -czf ceph_build.tar.gz *
--- a/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
+++ b/ceph-pull-requests/config/definitions/ceph-pull-requests.yml
@@ -64,6 +64,8 @@
         !include-raw:
           - ../../../scripts/build_utils.sh
           - ../../build/build
+    - trigger-builds:
+        - project: 'ceph-pr-clang-tidy'
 
     publishers:
       - cobertura:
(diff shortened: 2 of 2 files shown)


Suyashd999:
@dmick 
> 1. I think there's also some sort of permissions that need to be granted to the tidy job by the make check job to allow it to use the artifacts (see for example ceph-setup and job.properties.copyartifact)

1. Yes we must allow `Permission to Copy Artifact`, I'll check it in the file mentioned by you.

> 2) how long does ceph-tidy take?

2. The clang-tidy would take a maximum of 15-20 minutes to complete execution. 

> Do we want it to run on every ceph make-check build?

Yes we must run after every make-check build.

dmick:
Well, not "must", it, like everything, is a choice.  Let me bring this up to the clt.

Suyashd999:
> Well, not "must", it, like everything, is a choice. Let me bring this up to the clt.

Okay no problem

Suyashd999:
>  think there's also some sort of permissions that need to be granted to the tidy job by the make check job to allow it to use the artifacts (see for example ceph-setup and job.properties.copyartifact)

Hi @dmick! I have added the code to give permission to our job `ceph-pr-clang-tidy` to copy artifacts from `ceph-pull-requests`
Please do check.

Suyashd999:
> Well, not "must", it, like everything, is a choice. Let me bring this up to the clt.

@dmick 
any update from clt?

Suyashd999:
does everything look good ? @dmick

dmick:
I'm sorry we haven't made more progress with this.  You've hit on an area of Ceph development that is not very flexible to new work right now.  We're going to continue trying to move the discussion along.

Suyashd999:
> I'm sorry we haven't made more progress with this. You've hit on an area of Ceph development that is not very flexible to new work right now. We're going to continue trying to move the discussion along.
@dmick 
I see.. Please let me know what should I do from my end.

PR #2282, opened 2024-09-23
windows: ignore "apt-get update" errors

idryomov:
... as they are ignored in scripts/build_utils.sh since commit 3517941ef2aa ("script/build_utils.sh: don't fail if any mirror is not accessible").  This is prompted by an issue with libboost repo, which started returning 403 Forbidden errors.

While at it, skip fetching language and translation packs.


petrutlucian94:
@idryomov thanks for addressing this!

petrutlucian94 (approved):

PR #2285, opened 2024-10-31
ceph-dev-new-setup: Support tar.xz

zmc:


dmick (approved):
ew, you made me learn about bash arrays.  this will not be forgiven.

PR #2292, opened 2024-11-13
Clean up problematic rpm macro, install skopeo package for centos9

dmick:
couple small changes to the ansible for Jenkins builders:

1) don't set dist in rpmmacros; it was causing el9 packages to be labeled el8, and isn't necessary at all
2) add skopeo to installed packages (for container builds).

The package code ought to be cleaned up so that 'epel_packages' is what it says it is.


dmick:
@zmc @djgalloway  ping

zmc (approved):

PR #2298, opened 2024-12-05
build_utils.sh, various PR build scripts: skip if container/

dmick:
Parallel to the docs-only test, skip PR checks (api, make check, etc.) if the only changes are to the container/ subdirectory


dmick:
I've tested this by manually updating the jobs and seeing test PRs for files only in container/ succeed without doing work in API and make check, while other PRs are still running the full make check / API tests.

djgalloway (approved):

PR #2303, opened 2024-12-09
ceph-container: remove legacy jobs

guits:
These jobs are no longer needed, see ceph/ceph-container#2256

PR #2305, opened 2024-12-09
remove all jobs referring to ceph-container.

dmick:


dmick:
I slightly prefer this to https://github.com/ceph/ceph-build/pull/2303, just because it keeps them more easily accessible so I can deal with things that I've missed in the port to using ceph.git:container/

djgalloway (approved):

ktdreyer (approved):

PR #2308, opened 2024-12-15
feat: Build arm64 on bookworm

jinnatar:
Related to https://tracker.ceph.com/issues/69083

I have no idea if your CI will crash & burn, but I also can't test it, so let's see what happens.

### DIFF
--- a/ceph-build/config/definitions/ceph-build.yml
+++ b/ceph-build/config/definitions/ceph-build.yml
@@ -12,7 +12,7 @@
     execution-strategy:
        combination-filter: |
          DIST == AVAILABLE_DIST && ARCH == AVAILABLE_ARCH &&
-         (ARCH == "x86_64" || (ARCH == "arm64" && ["bionic", "focal", "jammy", "centos9"].contains(DIST)))
+         (ARCH == "x86_64" || (ARCH == "arm64" && ["bionic", "focal", "jammy", "centos9", "bookworm"].contains(DIST)))
     axes:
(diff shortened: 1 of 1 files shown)


jinnatar:
@djgalloway: Might you perhaps be able to review and/or point me in a direction that allows this CI change to run so we have an idea what else if anything needs changing to succeed?

Backstory is in https://tracker.ceph.com/issues/69083 but the short version is that the mirror _seems_ to contain arm64 packages for Debian, but it's actually in a very inconsistent state and missing most of the core packages. This PR is a feeble attempt to take it from 10% -> 100%, the other alternative being a different kind of change to take it from 10% -> 0% if the intent is to not publish for arm64 on Debian.

djgalloway:
Hi @jinnatar.  Appreciate your contribution here.  The Jenkins job definition you've modified is for our upstream signed packages that end up on download.ceph.com.  What I'd like to see happen is this change tested and working in a development repo first.  To do that, you would need to make this modification to all the job definition blocks in the ceph-dev and ceph-dev-new directories first.  Then if all looks correct, a new PR to modify the upstream release job.

jinnatar:
I'll have a go at updating the `ceph-dev` & `ceph-dev-new` dirs, but at a glance their configs are positively ancient so it'll take some fumbling to make progress. (Which is why I didn't think them relevant before, they seemed abandoned given that Debian is missing completely, when you in fact do package for Debian).

PR #2309, opened 2024-12-17
Revert "ceph-website-prs/config/definitions: remove org filter for site compile check

ljflores:
This reverts commit a227cbd01c628e9d98cf5cb3d96f0a299c4d12c3. We decided that this is too much of a security risk and are looking into better options.


djgalloway:
You can remove the whitelist in a second commit if you want.  Or add whomever already created some PRs to it.

PR #2310, opened 2024-12-27
Add ceph-dev-pipeline

zmc:
This may ultimately replace the existing ceph-*build jobs; for now, it is a
work in-progress.

### DIFF
--- a/ceph-dev-new-build/build/setup_rpm
+++ b/ceph-dev-new-build/build/setup_rpm
@@ -44,10 +44,6 @@ pwd
 
 setup_rpm_build_deps
 
-if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && "$RELEASE" =~ 8|9 ]] ; then
-      podman login -u $CONTAINER_REPO_USERNAME -p $CONTAINER_REPO_PASSWORD $CONTAINER_REPO_HOSTNAME/$CONTAINER_REPO_ORGANIZATION
-fi
-
 cd $WORKSPACE
 
 pkgs=( "chacractl>=0.0.21" )
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -1,0 +1,170 @@
+ceph_build_repo = "https://github.com/ceph/ceph-build"
+ceph_build_branch = "jenkinsfile"
+base_node_label = "gigantic"
+ubuntu_releases = [
+  "noble", // 24.04
+  "jammy", // 22.04
+  "focal", // 20.04
+]
+debian_releases = [
+  "bookworm", // v12 (2023)
+  "bullseye", // v11 (2021)
+]
+
+pipeline {
+  agent any
+  stages {
+    stage("source distribution") {
+      steps {
+        script {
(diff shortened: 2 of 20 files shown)


zmc:
An example run: https://jenkins.ceph.com/job/preserve-ceph-dev-pipeline/141/

scripts/build_container
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+# vim: ts=4 sw=4 expandtab
+set -ex
+PS4="\$(date --rfc-3339=seconds) + "
```
djgalloway, line 4: TIL... 👏

djgalloway:
Incredible work.  I have no qualms, suggestions, or nits.

zmc:
I wanted to set up label-based triggers to run this pipeline on ceph PRs, but it seems like I'd need to move the Jenkinsfile to the ceph repository to make that happen.

zmc:
I think this is complete enough to be merged (after removing the temporary commit and updating branch names). We also ought to document this new pipeline, but I couldn't find any documentation on the `ceph-dev-{new-,}` jobs to add to.

zmc:
Once we merge this it might make sense to branch it to see about integrating https://github.com/ceph/ceph/pull/59841/

djgalloway (approved):

ceph-dev-new-build/build/setup_rpm
```diff
@@ -44,10 +44,6 @@ pwd
 
 setup_rpm_build_deps
 
-if [[ $CI_CONTAINER == "true" && $DISTRO == "centos" && "$RELEASE" =~ 8|9 ]] ; then
```
dmick, line 47: The hope of having this section here was to fail the build early if the credentials were set incorrectly (granted, there might have been a comment to that effect, sorry).  Are you thinking that being able to run container builds separately from previously-built packages will be seamless enough that this isn't necessary anymore?

dmick (commented):
The one comment about podman login is all I really see to discuss at this point.  This is great work, and although I'm not necessarily looking forward to maintaining Groovy, this is clearly the right Jenkins direction.

ceph-dev-new-build/build/setup_rpm
zmc: Ah, no that wasn't clear. But yeah, since this pipeline does support building containers from shaman repos, I'm thinking it's not as important to check the credentials early anymore. They don't break regularly, right...?

dmick (approved):

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -0,0 +1,229 @@
...
+base_node_label = "gigantic"
+ubuntu_releases = [
+  "noble", // 24.04
+  "jammy", // 22.04
+  "focal", // 20.04
+]
+debian_releases = [
+  "bookworm", // v12 (2023)
+  "bullseye", // v11 (2021)
+]
```
djgalloway, line 12: Embarrassing question... Why is there no RPM distro list?

zmc: Since those distros don't use codenames, we can use a regex like `^(centos|rhel|fedora).*` - we don't need lists.

PR #2313, opened 2025-01-10
ceph-dev, ceph-dev-new: Add bookworm build

jinnatar:
Prereq for #2308

Likely still WIP as I don't have Jenkins credentials for a local build and need Jenkins to trigger to see if any of it works.

### DIFF
--- a/ceph-dev-new/config/definitions/ceph-dev-new.yml
+++ b/ceph-dev-new/config/definitions/ceph-dev-new.yml
@@ -25,8 +25,8 @@
 
       - string:
           name: DISTROS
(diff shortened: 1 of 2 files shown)


jinnatar:
Welp, this did not automagically trigger any Jenkins build, there is no documentation how to manually trigger / request one and I'm pretty sure none of this is actually fixable or even troubleshootable by someone outside the Ceph org.

So I guess I'll patiently wait for the Ceph org do fix this one.

djgalloway:
We don't test dev builds of Debian packages in our upstream lab / integration test framework.  Just Ubuntu / CentOS / Rocky.

djgalloway:
We're building umbrella for bookworm in ceph-dev-pipeline now

jinnatar:
It's a bit late for that given trixie is already out?

djgalloway:
> It's a bit late for that given trixie is already out?

Bookworm and trixie.

PR #2314, opened 2025-01-10
Do not run make check or API tests on smithi due to OOM

djgalloway:
Saw many jobs fail on smithi004.  dmesg shows lots of OOM kills.

image


dmick (approved):
Trying to verify briefly, I found 12 smithi tags in ceph-sepia-secrets, but can only find 5 smithis on jenkins.ceph.com and 2 on 2.jenkins.ceph.com, and I don't know whats happening.  But ruling out 5 currently probably isn't that big a deal I guess.

PR #2320, opened 2025-02-03
ceph-website: setup nodejs for building the website

nizamial09:


scripts/ceph-website/install-deps.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -ex
+
+# install node.js
+NODEJS_MAJOR_VERSION=20
+DISTRO="$(lsb_release -cs)"
+if [[ ! $(command -v node) || $(node --version | grep -oE "v([0-9])+" | cut -c 2-) < ${NODEJS_MAJOR_VERSION} ]]; then
+    NODEJS_KEYRING=/usr/share/keyrings/nodesource.gpg
```
dmick, line 8: I would add an 'echo' here explaining what the script found and what it's doing ("nodejs version %d too old, installing %d instead")

scripts/ceph-website/install-deps.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
dmick, line 1: The only overall comment is that it's a drag to have to revisit this as we change node versions, and it seems like it's going to just keep getting out of date again.  I don't know if there's a better solution for setting what version of node to use...some feature-based test or something...but if there is something that's better than "it's pinned to 20 until the next time something breaks" that would be preferable.  It wouldn't surprise me if that's not feasible, but I thought I'd throw it out there.

nizamial09: That's a nice idea. Maybe we could leverage [nvm-wrapper](https://www.jenkins.io/doc/pipeline/steps/nvm-wrapper/) in the jenkins. Ideally the plugin should have supported getting the [version from .nvmrc](https://github.com/ceph/ceph.io/blob/main/.nvmrc) which is already there in the ceph.io which I just saw. But since this plugin doesn't have the native .nvmrc support, probably we'll have to just read the node version from the file and install the node. It could be something we can try. 

But I don't know how to install and configure a plugin in the Jenkins Job

nizamial09: now that i said all this, probably we could just for now install the nvm in the machine and configure it and that should be fine. Let me have a look.

nizamial09: just used nvm here.

ceph-website/build/build
```diff
@@ -3,6 +3,14 @@ set -ex
 
 env
 
+set +e
```
nizamial09, line 6: `set +e` because I saw script abruptly ends here while running the export command.

dmick: I don't follow; was export causing an error somehow?

nizamial09: yeah, it was an issue with nvm i suppose. I was following this thread which had a couple of suggestions but none of it worked for me here

dmick: seems pretty hacky, but okay.....

nizamial09: https://github.com/nvm-sh/nvm/issues/1985, thought i pasted the issue i followed before but i forgot to

dmick (approved):

PR #2324, opened 2025-02-14
update all jobs to jenkins-job-builder v6; add 'choose from 7 days old' to ceph-dev-new-trigger

dmick:
jjb v6 has several non-backward-compatible changes:

- {macro} expansion is done unconditionally now, so all usages of ${var} change to ${{var}}
- !include-raw splits into several options; the one we always want is !include-raw-verbatim (no macro expansion)

This changeset

- changes all jobs to conform to jjb6 syntax
- uses a new prerelease version of jenkins-job-builder (integrated into main, but not yet released) from the git repo
- uses the functionality in that prerelease version to set 'maximum-age' of a ref to build for ceph-dev-new-trigger to 7 days

Note that some of the jobs are template jobs, and already had ${{}} in places, because they already needed to distinguish between macro usage and variable substitution.

Users who wish to run jenkins-job-builder locally will need to update to the prerelease version.  This can be done with "pip install git+https://opendev.org/jjb/jenkins-job-builder@60f031638"

### DIFF
--- a/attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
+++ b/attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
@@ -138,4 +138,4 @@
               username: CONTAINER_REPO_USERNAME
               password: CONTAINER_REPO_PASSWORD
       - build-name:
-          name: "#${BUILD_NUMBER} ${BRANCH}, ${DISTROS}, ${ARCH}, ${FLAVOR}"
+          name: "#${{BUILD_NUMBER}} ${{BRANCH}}, ${{DISTROS}}, ${{ARCH}}, ${{FLAVOR}}"
--- a/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
+++ b/ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
@@ -33,7 +33,7 @@
       - conditional-step:
           condition-kind: regex-match
           regex: .*reef.*
-          label: '${GIT_BRANCH}'
+          label: '${{GIT_BRANCH}}'
           on-evaluation-failure: dont-run
           steps:
             - shell:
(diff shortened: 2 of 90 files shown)


attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
```diff
@@ -138,4 +138,4 @@
               username: CONTAINER_REPO_USERNAME
               password: CONTAINER_REPO_PASSWORD
       - build-name:
-          name: "#${BUILD_NUMBER} ${BRANCH}, ${DISTROS}, ${ARCH}, ${FLAVOR}"
+          name: "#${{BUILD_NUMBER}} ${{BRANCH}}, ${{DISTROS}}, ${{ARCH}}, ${{FLAVOR}}"
```
djgalloway, line 141: I find it really interesting git decided which brackets were the new ones randomly...

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
```diff
@@ -84,23 +84,23 @@
       - conditional-step:
           condition-kind: shell
           condition-command: |
-            echo "${GIT_BRANCH}" | grep -v '\(reef\|squid\|centos9-only\|crimson-only\|jaeger\)'
+            echo "${{GIT_BRANCH}}" | grep -v '\(luminous\|mimic\|nautilus\|octopus\|pacific\|quincy\|reef\|squid\|centos9-only\|crimson-only\|jaeger\)'
```
djgalloway, line 87: 👍

djgalloway: But maybe a comment about why?

djgalloway:
LGTM aside from the one comment

attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
dmick: the github view shows them all to me ?

ceph-dev-new-trigger/config/definitions/ceph-dev-new-trigger.yml
dmick: er...actually this was a mistake, and I think it should not be there

dmick:
The check failed because it needs the new version of jenkins-job-builder in place before the check can work.  I'll have to update that manually first

attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
djgalloway: Look ARCH vs. FLAVOR

image

djgalloway (approved):

dmick:
Had to also change ceph-build-pull-requests to use the new jjb, and update it manually to get it to run this check

PR #2326, opened 2025-02-15
Rework changed-files test (doc-only, container-only)

dmick:
Take an array of patterns to compare each changed file against. For docs, use the current suggestions from
https://tracker.ceph.com/issues/69650


dmick:
Tested by copying ceph-pr-api and gutting it, using this version of scripts/build_utils.sh, and watching logs.  Found some false positives and so reworked the logic; it looks good now.  Pushed a change to README.md which was successfully marked DOCS_ONLY.  I think it's ready to go.

djgalloway (approved):

PR #2327, opened 2025-02-18
ansible: Add tag to python interpreter var setting tasks

djgalloway:
I'm working on a combination of terraform and ansible to quickly spin up instances in IBM Cloud so they can join Jenkins as builders.

Part of this process involves creating a virtualenv and running ansible from it.  python-jenkins needs to be installed in the virtualenv and setting the python interpreters inside the playbook breaks that.

This PR will let me `--skip-tags interpreter`


dmick (approved):
sure, why not

PR #2329, opened 2025-02-25
ceph-dev*: intermediate transition to WITH_CRIMSON

Matan-B:
See: https://github.com/ceph/ceph/pull/61672
This PR changes any WITH_SEASTAR / --with-seastar to WITH_CRIMSON / --with crimson respectively. We should let our builders add the new flags and once the PR is merged we could clean up the unused WITH_SEASTAR.

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -874,6 +874,7 @@ ceph_build_args_from_flavor() {
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
+        CEPH_EXTRA_RPMBUILD_ARGS="${CEPH_EXTRA_RPMBUILD_ARGS} --with crimson"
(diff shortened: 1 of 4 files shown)


scripts/build_utils.sh
```diff
@@ -874,6 +874,7 @@ ceph_build_args_from_flavor() {
         ;;
     crimson)
         CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
+        CEPH_EXTRA_RPMBUILD_ARGS="${CEPH_EXTRA_RPMBUILD_ARGS} --with crimson"
```
rzarzynski, line 877: This construct looks weird / confusing. Isn't `+=` working here?

Matan-B: I tried `+=` first but I saw already existing users here that go with this approach so I preferred keeping it that way. This should be removed once the PR mentioned is merged.

rzarzynski (approved):

dmick (approved):

PR #2331, opened 2025-02-27
Install and configure podman on Ubuntu builders

djgalloway:
This is the follow-up and preferred solution to https://github.com/ceph/ceph-build/pull/2330

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -64,10 +64,6 @@
           # jenkins-job-builder job:
           - libyaml-dev
           - jq
-          # ceph-docs job:
-          - doxygen
-          - ditaa
-          - ant
           - tmpreaper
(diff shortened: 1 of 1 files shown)


ansible/examples/builder.yml
```diff
@@ -486,6 +487,43 @@
         update_cache: yes
       when: ansible_os_family == "Suse"
 
+    ## UBUNTU PODMAN TASKS
```
dmick, line 490: are these OK to run on non-ubuntu?  If not should they have a when: ansible_os_family == "Ubuntu" or whatever?

djgalloway: They are actually safe to run on RPM-based builders.  I updated the block's comment.

dmick (approved):

PR #2332, opened 2025-02-28
ansible: Fixes to podman subuid/subgid checks

djgalloway:
I think this was working during my testing because sub{u,g}id_check.rc /wasn't/ 0.  I had manually modified the system so it wasn't in the same state as a fresh cloud instance.

I manually ran this playbook on an instance and confirmed it gets past the podman setup now.

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -501,9 +501,9 @@
       changed_when: false
 
     - name: Find highest used subuid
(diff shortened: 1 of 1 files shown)


dmick (approved):
Looked a little more carefully and this looks right: "command" doesn't allow shell constructs like pipe, "shell" does, and "rc == 0" from the greps is when the greps have found the string

PR #2334, opened 2025-03-10
ceph-dev-new-setup: Require recent podman

zmc:
And also run `restorecon` before using it.


djgalloway (commented):
What if podman is `-lt 4`?

zmc:
> What if podman is `-lt 4`?

It'd fall to the `else` clause and use docker

djgalloway:
Ahhhh, I understand the logic now.

What are your thoughts on also checking if `docker` is installed and just echo'ing an error about a misconfigured builder if neither command exist?

zmc:
If neither are installed, I'd argue we'd probably rather install the preferred one and then use it

djgalloway:
Is the version installed on the current Jenkins builders > 4?

zmc:
> Is the version installed on the current Jenkins builders > 4?

CentOS 9 has 5.3; Ubuntu 24.04 has 4.9. Jammy has 3.x; this is part of why I'd like to see us upgrade our Ubuntu hosts.

PR #2335, opened 2025-03-13
Add GPL v2 LICENSE to the repo

jmundack:
Adding LICENSE file to the repo.

picked the GPL v2 based on the "one of the" licenses in the main ceph repo + what chatgpt claims is the main LICENSE for the project

Steps followed from https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository to create the LICENSE file

Fixes: https://github.com/ceph/ceph-build/issues/1384

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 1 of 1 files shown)


cbodley:
from https://github.com/ceph/ceph/blob/main/COPYING#L6-L8

ceph is dual-licensed as LGPL-2.1 and LGPL-3. it includes lots of libraries with other compatible licenses. the only mention i see of GPL is the `bin/git-archive-all.sh` script

PR #2342, opened 2025-03-29
Add ceph-release-containers job

dmick:
Builds prerelease versions of ceph container images: two arch-specific images and one 'manifest-list' image.  Pushed all to quay.ceph.io/ceph, in a private set of repos (prerelease-amd64, prerelease-arm64, and prerelease) for testing and signoff.  Once moved, the script container/make-manifest-list.py can be used with the --promote flag to promote them from quay.ceph.io to quay.io for public release.


zmc (approved):
No notes, really. Nice work!

PR #2345, opened 2025-04-01
ceph-release-containers: tweak the parameter defaults

dmick:


ceph-release-containers/config/definitions/ceph-release-containers.yml
```diff
@@ -34,13 +34,13 @@
 
       - string:
           name: NO_PUSH
```
djgalloway, line 36: What even uses this parameter?

dmick: it's just there for testing.

PR #2356, opened 2025-04-17
Add scripts/setup_container_runtime.sh

zmc:
This supersedes #2334 and #2354. Tested here: https://jenkins.ceph.com/job/ceph-dev-new-setup/92635/

### DIFF
--- a/ceph-dev-new-setup/build/build
+++ b/ceph-dev-new-setup/build/build
@@ -103,13 +103,13 @@ echo current version $cephver
 
 srcdir=`pwd`
 
-PODMAN_MAJOR_VERSION=$(podman version -f json | jq -r '.Client.Version|split(".")[0]')
-if command -v podman && [ "$PODMAN_MAJOR_VERSION" -ge 4 ]; then
-    PODMAN=podman
-    PODMAN_STORAGE_DIR="$HOME/.local/share/containers/storage/"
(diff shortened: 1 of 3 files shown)


dmick (changes requested):
lot of hard-acquired knowledge buried in here; good stuff

ceph-dev-new-setup/build/build
```diff
@@ -103,12 +103,11 @@ echo current version $cephver
...
 
+setup_container_runtime
 if command -v podman; then
-    PODMAN=podman
-    PODMAN_STORAGE_DIR="$HOME/.local/share/containers/storage/"
-    test -d $PODMAN_STORAGE_DIR && sudo chgrp -R $(groups | cut -d' ' -f1) $PODMAN_STORAGE_DIR
+  PODMAN=podman
 else
-    PODMAN="sudo docker"
+  PODMAN=docker
```
dmick, line 110: how do we no longer need sudo with docker?

zmc: did we need it in the first place?

dmick: it used to be the case that only root could contact dockerd, sometime in the past

dmick: ...or the user needs to be part of the docker group, seems to still be the case

zmc: It checks for the group now

dmick (approved):

PR #2357, opened 2025-04-21
Add ceph-source-dist job

zmc:
The intent is to replace the various `ceph-*setup` jobs.

### DIFF
--- a/ceph-source-dist/build/Jenkinsfile
+++ b/ceph-source-dist/build/Jenkinsfile
@@ -1,0 +1,111 @@
+pipeline {
+  agent {
+    label "gigantic"
+  }
+  stages {
+    stage("set environment variables") {
+      steps {
+        script {
+          switch (env.FLAVOR) {
+            case "default":
+              env.CEPH_EXTRA_RPMBUILD_ARGS="--with tcmalloc"
+              env.CEPH_EXTRA_CMAKE_ARGS+=" -DALLOCATOR=tcmalloc"
+              env.CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_SYSTEM_BOOST=OFF -DWITH_BOOST_VALGRIND=ON"
+              env.DEB_BUILD_PROFILES=""
+              break
+            case "crimson":
+              env.CEPH_EXTRA_RPMBUILD_ARGS="--with seastar"
+              env.CEPH_EXTRA_CMAKE_ARGS+=" -DCMAKE_BUILD_TYPE=Debug"
+              env.DEB_BUILD_PROFILES="pkg.ceph.crimson"
+              break
+            case "jaeger":
+              env.CEPH_EXTRA_RPMBUILD_ARGS="--with jaeger"
+              env.DEB_BUILD_PROFILES="pkg.ceph.jaeger"
+              break
+            default:
+              println "FLAVOR={env.FLAVOR} is invalid"
+              assert false
+          }
+        }
+      }
+    }
+    stage("repository") {
+      steps {
+        dir("ceph") {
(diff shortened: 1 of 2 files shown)


scripts/make_source_dist.sh
```diff
@@ -0,0 +1,94 @@
...
+else
+  PODMAN="sudo docker"
+fi
+
+printf "FROM ubuntu:24.04\nRUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y dpkg-dev devscripts && apt-get clean && rm -rf /var/lib/apt/lists/*" | $PODMAN build -t ubuntu_builder -
+$PODMAN run --rm -v $PWD:/ceph:z -e DEBEMAIL=contact@ceph.com ubuntu_builder:latest bash -c "cd /ceph && dch -v $ceph_version_git-1 'autobuilder' < /dev/null"
+
+rm -rf dist release
+mkdir -p release
+echo $ceph_version_git > release/version
```
dmick, line 18: Seems like we lost some things: submodule update, 'dirty repo' check, maybe more; was that intentional?

scripts/make_source_dist.sh
```diff
@@ -0,0 +1,94 @@
...
+  compress_flags="${flag}cf"
+  if [ "$ceph_version_tarball" != "" ]; then break; fi
+done
+echo tarball vers $ceph_version_tarball
+
+echo extracting
+mkdir -p release/$ceph_version_git
+cd release/$ceph_version_git
+
+# FIXME: is the rpm patch stuff below ever necessary?
```
dmick, line 41: I can't find any use since 2016; I'm OK with removing it

scripts/make_source_dist.sh
dmick: ah, submodule update is done in make-dist itself, I see now

scripts/make_source_dist.sh
```diff
@@ -0,0 +1,94 @@
...
+# mkdir -p release/$ceph_version_git/rpm
+# cp rpm/*.patch release/$ceph_version_git/rpm || true
+
+tar $extract_flags $WORKSPACE/ceph/ceph-$ceph_version_tarball.$extension
+
+[ "$ceph_version_tarball" != "$ceph_version_git" ] && mv ceph-$ceph_version_tarball ceph-$ceph_version_git
+
+tar zcf ceph_$ceph_version_git.orig.tar.gz ceph-$ceph_version_git
+# FIXME delete this if it's not needed
+ln ceph_$ceph_version_git.orig.tar.gz ceph-$ceph_version_git.tar.gz
```
dmick, line 51: I believe Debian packaging uses orig.tar.gz and the specfile uses tar.gz, although I thought that the Debian orig.tar.gz was not supposed to include the debian/ dir, and all the references I can find say that, but, ours does and apparently always has.  I guess maybe it's a "native" package and not a "non-native" package, and so it's assumed it has debian/ included.  in which case naming it 'orig' makes little sense.  I don't *(&#$ know.

dmick (approved):
I've quickly come to the conclusion that correctness is unknowable.  If it works, ship it.  Respond to the comments or change something if you like.

scripts/make_source_dist.sh
zmc: Turns out we are only using the bz2! https://jenkins.ceph.com/job/ceph-dev-pipeline/288/pipeline-console/?selected-node=234

zmc:
@dmick I think this is sactually ready at this point. See the pipeline build I did with it: https://jenkins.ceph.com/job/ceph-dev-pipeline/297

PR #2360, opened 2025-04-23
scripts: docs_pr_only() considers PendingReleaseNotes

cbodley:
avoid building from source and running the various 'make check's when the pr only touches PendingReleaseNotes


cbodley:
noticed the unnecessary checks in https://github.com/ceph/ceph/pull/62936

dmick (approved):

PR #2362, opened 2025-05-07
Remove Github Pull Request Builder from required PR Check jobs

djgalloway:
This is the start of extricating Github Pull Request Builder.  GHPRB has some functionality that we've relied on forever: sending Github Status Checks back to Github PRs.  This can easily be accomplished using the API now.

The ceph-pull-requests, ceph-pull-requests-arm64, ceph-windows-pull-requests, and ceph-pr-api jobs will now be triggered by new Github Workflows (collection of Github Actions): https://github.com/ceph/ceph/pull/63168.

The other added benefit of this change is we no longer have to check for docs-only or container-only PRs on the Jenkins side of things.  Github will take care of checking and send fake status checks back to applicable PRs.

### DIFF
--- a/ceph-pr-api/build/build
+++ b/ceph-pr-api/build/build
@@ -1,13 +1,8 @@
 #!/bin/bash -e
 
-docs_pr_only
-container_pr_only
-if [[ "$DOCS_ONLY" = true || "$CONTAINER_ONLY" = true ]]; then
-    echo "Only the doc/ or container/ dir changed.  No need to run make check or API tests."
-    mkdir -p $WORKSPACE/build/out
-    echo "File created to avoid Jenkins' Artifact Archiving plugin from hanging" > $WORKSPACE/build/out/mgr.foo.log
-    exit 0
-fi
+echo "This job was triggered by $TRIGGER_METHOD via $TRIGGERED_BY"
+
+report_github_check_status pending "ceph API tests" "ceph API tests running" $GH_PULL_REQUEST_SHA $BUILD_URL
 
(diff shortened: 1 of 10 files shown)


scripts/build_utils.sh
```diff
@@ -1850,3 +1850,42 @@ function set_centos_python3_version() {
     sudo dnf reinstall -y $EXPECTED_PYTHON3_VERSION || sudo dnf install -y $EXPECTED_PYTHON3_VERSION
     sudo ln -fs /usr/bin/$EXPECTED_PYTHON3_VERSION /usr/bin/python3
 }
+
+function report_github_pr_check_status() {
```
zmc, line 1854: can this go in a separate file?

djgalloway: In `scripts/`?   I suppose.  I'm not sure what the point of a massive `build_utils.sh` file was anyway.

zmc: yeah, I'd suggest adding a new file in `scripts/`, maybe even named after this function. A snippet like this can make it both executable and source-able: https://github.com/ceph/ceph-build/blob/pipeline-container/scripts/update_shaman.sh#L48-L51

dmick: The point was simplicity; "one library"  Bash is not a good modular language.

djgalloway: Done

djgalloway:
We're not going to use github actions to trigger.  Prefer https://github.com/ceph/ceph-build/pull/2674.

PR #2363, opened 2025-05-07
ceph-dev-pipeline: Build in containers

zmc:


scripts/setup_container_runtime.sh
```diff
@@ -33,7 +33,7 @@ function setup_container_runtime () {
       if [ -d $PODMAN_STORAGE_DIR ]; then
         sudo chgrp -R $(groups | cut -d' ' -f1) $PODMAN_STORAGE_DIR
         if [ $(podman unshare du -s --block-size=1G $PODMAN_STORAGE_DIR | awk '{print $1}') -ge 50 ]; then
-          time podman image prune --filter=until="$(echo '24*7*2' | bc)h" --all --force
+          time podman image prune --filter=until="$(echo '24*7' | bc)h" --all --force
```
phlogistonjohn, line 36: I don't know what embedded shell this supports, but if it's bash you could do: `$((24*7))` instead of invoking bc.

djgalloway (approved):

scripts/setup_container_runtime.sh
zmc: ah you're right! I always forget about that capability. It is bash that's being used here.

PR #2365, opened 2025-05-14
FIx up builder.yml

dmick:
Allow automatically fetching api secret from secrets inventory based on 'api_uri'; fix up subuid setting; add comments

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -4,7 +4,12 @@
 ## Tested on: CentOS 7, CentOS 8, Xenial, Bionic, Focal, Leap 15.1 using ansible 2.8.5
 ##
 ## Example:
-## ansible-playbook -vvv -M ./library/ builder.yml -e '{"labels": "x86_64 xenial etc", "token": "XXXXX", "jenkins_credentials_uuid": "jenkins-build", "api_uri": "https://jenkins.ceph.com"}' -e permanent=true -e ansible_ssh_user=ubuntu --limit braggi01*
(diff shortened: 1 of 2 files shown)


ansible/examples/builder.yml
```diff
@@ -33,8 +33,10 @@
       tags:
         always
 
+    # get appropriate secret (clumsy variable names because you can't start
+    # a variable with a digit
     - set_fact:
-        token: "{{ jenkins_api_token }}"
+        token: "{{ ('2' in api_uri) | ternary(api_token_2_jenkins, api_token_jenkins)}}"
```
djgalloway, line 45: Might it be cleaner to store these in individual host-specific `{{ secrets_path }}/{{ host_name }}/jenkins_api_token.yml` files?

djgalloway: It's cleaner 😆.  IDK, I guess it doesn't matter if we never plan on having more than the 2 Jenkins instances.

djgalloway (approved):

ansible/examples/builder.yml
dmick: could also have hostname as a key in the dict I suppose.  Maybe having individual files would make it easier to edit in place etc. because more things handle the "straight encrypted file" than handle "encrypted strings".

dmick: OK, I reworked it, and also removed slave.yml, a symlink I had from long ago that somehow sometime snuck back in

files are simple 'ansible-fault encrypt' now, named by hostname

djgalloway: I like that better.  Thx!

PR #1805, opened 2021-04-28
scripts/build_utils.sh: build Debug builds for crimson flavor 

tchaikov:

PR #2367, opened 2025-05-18
scripts/build_utils: remove Debug build from Crimson

Matan-B:
See: https://github.com/ceph/ceph-build/pull/2167/files Builds should be set to Debug only when ending in "-debug".

https://github.com/ceph/ceph-build/pull/1805 set Crimson to always use Debug. That made sense in earlier development phases.
However, today we should align with Classical builds and use Debug only when needed.


Matan-B:
For context, https://github.com/ceph/ceph/pull/61637 will be used to set to RelDebWithInfo

cyx1231st (approved):

PR #2368, opened 2025-05-19
ceph-build: remove container build step

dmick:
This was removed from build_rpm, but then when a reorg happened to separate the function into a separate script, the fact that ceph-build doesn't do container builds anymore was missed.

Release container builds now happen after the manual signing step using job ceph-release-containers


dmick:
probably my fault reviewing, @zmc, but tagging you so you're aware

djgalloway (approved):

PR #2369, opened 2025-05-27
Seperate Crimson flavors to crimson-debug/release

Matan-B:
The following change would allow us to:

* Keep testing "Debug" builds (crimson-debug) by default.
* When crimson-only is used, crimson-release would also be built.
* `crimson-release` builds would be used in the weekly nightlies and on
  selected PR gatings (when crimson-debug is not enough).

Note: Once crimson-release is stable enough we could switch crimson-debug
      to actaully be of: "RelWithDebInfo + !NDEBUG + asserts" instead of "Debug".
      For now, let's keep it as Debug since we know it's stable.

See: https://github.com/ceph/ceph/pull/63557

### DIFF
--- a/ceph-dev-cron/config/definitions/ceph-dev-cron.yml
+++ b/ceph-dev-cron/config/definitions/ceph-dev-cron.yml
@@ -98,7 +98,7 @@
                     BRANCH=${{GIT_BRANCH}}
                     FORCE=True
                     DISTROS=centos9
-                    FLAVOR=crimson
+                    FLAVOR=crimson-debug
                     ARCHS=x86_64
       # build main on:
       # default: jammy centos9 windows
@@ -124,7 +124,7 @@
                     BRANCH=${{GIT_BRANCH}}
                     FORCE=True
                     DISTROS=centos9
-                    FLAVOR=crimson
+                    FLAVOR=crimson-debug
                     ARCHS=x86_64
 
     wrappers:
--- a/ceph-dev-cron/config/definitions/ceph-dev-cron.yml
+++ b/ceph-dev-cron/config/definitions/ceph-dev-cron.yml
@@ -102,7 +102,8 @@
                     ARCHS=x86_64
       # build main on:
       # default: jammy centos9 windows
-      # crimson: centos9
+      # crimson-debug: centos9
+      # crimson-release: centos9
       - conditional-step:
           condition-kind: regex-match
           regex: .*main.*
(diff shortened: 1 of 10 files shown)


cyx1231st:
I assume after this change, we need to name the branch pushed to ceph-ci as "...-crimson-only" instead of "...-debug", right?

Matan-B:
@zmc, @djgalloway, @dmick - Can you please help with an ack from your side pre merging? Thanks!

dmick:
How does ceph handle debug vs non-debug builds with default flavors?  Why are crimson builds different?

attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
djgalloway, line 1: There's no need to modify a job in the 'attic'

attic/ceph-dev-trigger/config/definitions/ceph-dev-trigger.yml
djgalloway, line 1: There's no need to modify a job in the 'attic'

athanatos:
> How does ceph handle debug vs non-debug builds with default flavors? Why are crimson builds different?

The consensus at the moment is that the other components aren't as interested in compiling in asserts during teuthology testing.

By contrast, crimson, and particularly seastore, has a bunch of asserts that we want in during testing but not in the release build (performance impact is measurable).

In summary, different needs.

dmick:
Are you saying that ceph proper builds with debug on for CI and release builds?

Matan-B:
> Are you saying that ceph proper builds with debug on for CI and release builds?

Assuming ceph proper refers to non-crimson builds, no. Debug is not used in is CI and release builds unless explicitly requested. Branches ending with `*-debug` suffix are the only way to create Debug builds.

--

For clarification, This PR solely affects Crimson builds due its specific needs.
Any concerns regarding other builds types can be further discussed out of this PR scope.

attic/ceph-dev-container-only/config/definitions/ceph-dev-container-only.yml
Matan-B: The only reason I see for updating jobs in "attic" is to be consistent across all users in the repo. Otherwise, it would be hard, without additional context, to understand why some users (i.e flavor type lists) are different than others across the repo.

If the attic is not planned to be updated anymore, should it be moved to git's ignore list?

dmick:
For clarification, this PR is clearly about crimson builds, and the point of bringing up other builds was to understand why crimson has different requirements compared to other styles of builds.

The desire for build change makes sense to me.  Please remove the changes to attic/ files and we'll proceed.

Matan-B:
/attic  file changes removed.
@dmick, @djgalloway - can you please review? Thanks!

PR #2373, opened 2025-06-04
remove 'jaeger' flavor now that it's on by default

cbodley:
jaeger was enabled by default in all ceph builds in https://github.com/ceph/ceph/pull/44684 which was present in reef and later releases. so no supported releases need a flavor to turn it on

ceph-dev-new-trigger hadn't been updated from 'DISTROS=centos8 focal' so is probably not being used


dmick (approved):
lgtm

cbodley:
rebased to resolve conflicts from https://github.com/ceph/ceph-build/pull/2369

Issue #2351, opened 2025-04-09
ceph-dev-pipeline: Locate ceph-dev-new-setup builds automatically


zmc:
c-d-p takes an optional param `SETUP_BUILD_ID` which can be pointed at a previously-completed ceph-dev-new-setup build, and use its artifacts instead of invoking its own c-d-n-s build. This saves a lot of time on rebuilds, but only if one remembers to manually insert the parameter.

It would be even better if c-d-p could locate one automatically via (most likely) calls to the Jenkins REST API.

PR #2378, opened 2025-06-11
[ceph-dev-pipeline] Get SETUP_BUILD_ID from last success build

deepssin:
Issue - https://github.com/ceph/ceph-build/issues/2351

Added `fingerprint: true` for audit purposes while copying artifacts.


deepssin:
this is a dup, already merged via another PR

PR #2379, opened 2025-06-12
Fix up ssh client and host key code; add comments on how to run

dmick:


ansible/examples/builder.yml
```diff
@@ -1,12 +1,17 @@
...
+# it or something.  (files/, templates/ are two such paths)
+#
 ## Instead of trying to keep 4 separate playbooks up to date, let's do it all here.
 ## The only difference from using multiple playbooks is we need to specify `-e libvirt=true` and/or `-e permanent=true` if the builder will be permanent/static.
 ## Tested on: CentOS 7, CentOS 8, Xenial, Bionic, Focal, Leap 15.1 using ansible 2.8.5
 ##
 ## Example:
 ## define labels in inventory "jenkins_labels" dict, keyed by fqdn
 ##
-## ansible-playbook -vvv -M ./library/ builder.yml, "api_uri": "https://jenkins.ceph.com"}' --limit braggi01*
```
djgalloway, line 9: 🤦

djgalloway:
Why remove the `github` ansible tags?

ansible/examples/builder.yml
dmick: I did that last commit after removing some other -e

dmick:
github tags: I had added some more, and when I removed them just decided to remove them all.  (I think I added them in a past commit too, mostly just for convenience of test.  I think the github host key has changed exactly once in all the time we've had Ceph on github)

ansible/examples/builder.yml
djgalloway: I assumed it was me

PR #2387, opened 2025-06-24
builder.yml: Add tasks to update builder

akraitman:
- Installing java-21 insted of java-17(EOL)
 - Stoping unattended-upgrades to avoid conflicts
 - Refreshes the apt package cache to ensure the system has the latest package metadata before performing upgrades
 - Upgrades all installed packages to their latest versions on debian based systems and cleans up unused dependencies and cached files
 - Updates the package cache and upgrades all packages to their latest versions on redhat based systems
 - Updates the package cache and upgrades all packages on suse based systems
 - Fixing the task - "get github host key from github if necessary" that checks github_host_key == "" but since github_host_key is undefined, ansible throws an error
 - Fixing the taks - "Install the systemd unit files for jenkins" by using playbook_dir for explicit path resolution

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -1,8 +1,4 @@
 ---
-# NOTE: there are references to relative paths here, and if you want
-# them to work, this playbook must be in the parent of examples/, so, symlink
-# it or something.  (files/, templates/ are two such paths)
-#
 ## Instead of trying to keep 4 separate playbooks up to date, let's do it all here.
 ## The only difference from using multiple playbooks is we need to specify `-e libvirt=true` and/or `-e permanent=true` if the builder will be permanent/static.
 ## Tested on: CentOS 7, CentOS 8, Xenial, Bionic, Focal, Leap 15.1 using ansible 2.8.5
@@ -33,8 +29,7 @@
     osc_pass: 'password'
     container_mirror: 'docker-mirror.front.sepia.ceph.com:5000'
     secrets_path: "{{ lookup('env', 'ANSIBLE_SECRETS_PATH') | default('/etc/ansible/secrets', true) }}"
-    java_version: 'java-17'
-
+    java_version: 'java-21'
 
   tasks:
     - name: "Include appropriate jenkins API token"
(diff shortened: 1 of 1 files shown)


ansible/examples/builder.yml
```diff
@@ -1016,7 +1074,7 @@
 
         - name: Install the systemd unit files for jenkins
           ansible.builtin.template:
-            src: "templates/systemd/jenkins.{{ item }}.j2"
+            src: "{{ playbook_dir }}/../templates/systemd/jenkins.{{ item }}.j2"
```
djgalloway, line 1077: Is this a magic variable?

ansible/examples/builder.yml
```diff
@@ -825,16 +880,19 @@
...
       set_fact:
         github_host_key: "{{ lookup('file', github_host_key_file) }}"
       when: github_host_key_file != ""
+      tags: always
 
     - name: get github host key from github if necessary
       set_fact:
         github_host_key: "{{ lookup('url', 'https://raw.githubusercontent.com/ceph/ceph-build/main/ansible/files/ssh/hostkeys/github.com.pub') }}"
-      when: github_host_key == ""
+      when: github_host_key_file == ""
```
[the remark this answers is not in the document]
akraitman: I have added I comment above github_host_key is undefined it fails the task, the intent is to fetch the key from GitHub when the local file is not found

ansible/examples/builder.yml
dmick: We should do the same addition to the files/ references, and remove the comment at the top starting with NOTE: that I just added

ansible/examples/builder.yml
```diff
@@ -44,6 +43,62 @@
...
 
+    - name: Stop unattended-upgrades to avoid conflicts
+      systemd:
+        name: unattended-upgrades
+        state: stopped
+      when: ansible_os_family == "Debian"
+      ignore_errors: yes
+
+    - name: Wait for APT lock to be released
+      command: "timeout 300 bash -c 'while fuser /var/lib/apt/lists/lock /var/cache/apt/archives/lock /var/lib/dpkg/lock-frontend >/dev/null 2>&1; do sleep 5; done'"
```
dmick, line 54: I see that this is recommended in a few places; this blog entry makes the point that it's got some problems, and maybe offers a better solution with apt config variables:  https://blog.sinjakli.co.uk/2021/10/25/waiting-for-apt-locks-without-the-hacky-bash-scripts/.  I see that the apt module supports dpkg_options: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_module.html#parameters   WDYT?

ansible/examples/builder.yml
```diff
@@ -44,6 +43,62 @@
       tags:
         always
 
+    - name: Stop unattended-upgrades to avoid conflicts
+      systemd:
+        name: unattended-upgrades
+        state: stopped
+      when: ansible_os_family == "Debian"
+      ignore_errors: yes
```
djgalloway, line 51: It's not even supposed to be installed in the first place https://github.com/ceph/ceph-build/blob/main/ansible/examples/builder.yml#L459-L464 😭

ansible/examples/builder.yml
akraitman: Done, did this change on two more tasks in the playbook

ansible/examples/builder.yml
akraitman: These options don’t address the lock issue, but for "Upgrade all packages (Debian)" task I think of adding force-confold and force-confdef, this will ensure old configuration files are kept (force-confold) and defaults are used for new configs (force-confdef), it will reduce interactive prompts

akraitman: For the lock task, I will use wait_for instead of using the fuser

djgalloway (commented):
My only suggestion is the commit message should be something more like `builder.yml: Add tasks to update builder packages`.

Going through commit logs and just seeing a commit message like "Updates" means whomever is reading the logs has to go and actually look at the code changes.

ansible/examples/builder.yml
akraitman: You're right, I thought it's just waiting without checking if the lock files exist or not, I have added the lock_timeout to the packages upgrade tasks and I also added those options force-confold,force-confdef as mentioned in my previous comment

dmick: ok, well, the lock_timeout parameter is different yet, and is not implemented with DPkg::Lock::Timeout, but maybe it is good enough.  The module implements its own timeout loop.

PR #2388, opened 2025-06-24
ceph-trigger-build: Correct crimson-only behavior

zmc:
We were accidentally building default instead of crimson-debug


djgalloway:
> We were accidentally building default instead of crimson-debug

I feel dumb asking this but how does this not build all three now?

zmc:
yes, the `'default'` in `defaults` is the default; it gets replaced by `'crimson-debug`' in the added line.

PR #2390, opened 2025-06-27
ceph-trigger-build: correct git trailer examples

zmc:
Underscores aren't valid; we missed this in review


djgalloway:
I was going off the Jenkinsfile https://github.com/ceph/ceph-build/blob/8cad21a94efe7b5bacdcfe7d87f0fdb5f87c27e7/ceph-trigger-build/build/Jenkinsfile#L12.  Do the dashes get translated somewhere?

PR #2391, opened 2025-07-02
ceph-trigger-build: Clarify build job selection

zmc:


djgalloway:
Maybe clarify https://github.com/ceph/ceph-build/pull/2391/files#diff-4c0a7678c6fb42dcb0503596e92324226b0f4900ec1264f5da36a679506caa19R13 as well.

"All parameters are optional.  However, the use of git trailers is only supported in combination with `CEPH-BUILD-JOB: ceph-dev-pipeline` for now."

Issue #2336, opened 2025-03-13
Clean up Jenkins jobs that have never been run


jmundack:
Looking at the current list of Jenkins jobs we have at https://jenkins.ceph.com/view/all/

And sorting them by last success/failure - there are a bunch of jenkins jobs that have never been run (They show up as "NA") - we should remove them

deepssin:
| Job-name | Status | Reviewed? | Action
|--------|--------|--------|--------|
| ceph-cbt: lint tests | Never run | Yes | Future Usage Plan |
| ceph-grafana-trigger | Never run | In-progress | |
| [ceph-iscsi-cli: Flake8](https://jenkins.ceph.com/view/all/job/ceph-iscsi-cli-flake8/) | Never run | No | |

PR #2392, opened 2025-07-04
Moving jobs that never have been run to attic

deepssin:
Solves issue - https://github.com/ceph/ceph-build/issues/2336


jmundack:
can we just remove these? (and if need be - we can always `git revert` and pull it back in?)

my concern with moving this to the attic is - now we have a "new" dumping ground for things :)

hoping we can just leverage `git` instead of moving things around

jmundack:
> So @jmundack: I see that you did the investigation of this issue to start in the GitHub issue linked here. I think the "never have been run" is not strictly correct; I know those jobs have run in the past, and I suspect the stats have been reset by a reboot or reinstall of the jenkins server.

@dmick  - looking at this [job ](https://jenkins.ceph.com/user/jmundack/my-views/view/Jobs%20that%20have%20never%20successfully%20build/job/ceph-deploy/) for example - it failed 5 years ago. The fact that we have some jobs with history from 5 years ago, i would expect it to be safe to say the the "never been run" jobs reported by Jenkins - has not been run at least in the last 5 years? if so, we should be good to kill it?

my take on this is - we need to have some lifecycle for cleaning these up otherwise we will accumulate a lot of crud with community members coming in and our of the project

deepssin:
Closing , as looks like there is no reliable way to get the info on what has never been run.

Issue #2338, opened 2025-03-13
Remove Jenkins jobs for older releases and deprecated repos


jmundack:
Looking at the current list of Jenkins jobs we have at https://jenkins.ceph.com/view/all/

And sorting them by last success/failure - there are jobs that havent been run in years - These jobs should be removed to reduce the overhead on jenkins AND our jenkins administrators

eg: Jobs that are specific to EOL'd releases or for repos that have been migrated/deprecated

PR #2394, opened 2025-07-08
Move the jobs that have not been run in last 1 year to attic

deepssin:
Issue reference - https://github.com/ceph/ceph-build/issues/2338


djgalloway:
We still care about most of those.  ceph-windows-test is probably the only one that could move.

PR #2396, opened 2025-07-08
Enable linger for the Jenkins user

zmc:
This fixes errors like:
RunRoot is pointing to a path (/run/user/$ID/containers) which is not writable. Most likely podman will fail. Error: creating events dirs: mkdir /run/user/$ID: permission denied


akraitman (approved):

dmick:
crap, merged already

PR #2398, opened 2025-07-09
removed build_utils dependency from ceph build pull request job

Adarsha1999:


Adarsha1999:
Hello @dmick 

1. Updated the name of the script based on the feedback
2. For review https://github.com/ceph/ceph-build/blob/edb2fc36a54dd0ac23e688a36b919207280756fb/scripts/build_utils.sh#L6-L249

PR #2403, opened 2025-07-16
ceph-dev-pipeline: Only build arm64 for centos9

zmc:
I somehow missed this behavior in ceph-dev-new-build

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -192,8 +192,8 @@ pipeline {
             values 'default', 'crimson-release', 'crimson-debug'
           }
         }
-        // crimson is only supported on centos9 x86_64
         excludes {
+          // crimson is only supported on centos9 x86_64
           exclude {
             axis {
               name 'FLAVOR'
@@ -212,6 +212,17 @@ pipeline {
             axis {
               name 'ARCH'
               notValues 'x86_64'
+            }
+          }
+          // Only build arm64 for centos9
+          exclude {
+            axis {
+              name 'ARCH'
+              values 'arm64'
+            }
+            axis {
+              name 'DIST'
+              notValues 'centos9'
             }
           }
         }


zmc:
@djgalloway so we want to keep building both, then?

djgalloway:
> @djgalloway so we want to keep building both, then?

I guess, if it's possible, don't build by default but allow the possibility via git trailers?

zmc:
> > @djgalloway so we want to keep building both, then?
> 
> I guess, if it's possible, don't build by default but allow the possibility via git trailers?

I like this idea but it would be very awkward because of how we express build targets via parameters - meaning, we build a matrix consisting of all possible combinations of values.

One option would be to add a flag that somehow controls whether "less supported" targets are built, but I think that could cause quite a bit of confusion.

Another option would be to change how `ceph-dev-pipeline`'s parameters are expressed. Taking the example of a default `main` build, we would go from:
```
DISTROS=centos9 jammy noble windows
ARCHS=x86_64 arm64
```
To something like:
```
TARGETS=centos9/x64_64 centos9/arm64 jammy/x64_64 jammy/arm64 noble/x64_64 noble/arm64 windows/x64_64 windows/arm64
```
Or maybe:
```
TARGETS=centos9/{x64_64,arm64} jammy/{x64_64,arm64} noble/{x64_64,arm64} windows/{x64_64,arm64}
```
The latter is more concise but introduces a scenario where a typo could lead to confusing behavior and/or error messages.

djgalloway:
I am checking with CSC on Monday whether we care about Ubuntu builds and we will follow up next week.

phlogistonjohn:
David, Dan and I did a little looking and it's failing within the build of the UADK library, something that appears to been added specifically for arm support. So to perhaps narrow the discussion a little - we may need to ask if the Ceph arm maintainers are interested in supporting the ubuntu builds on arm64, because they're the ones who added uadk support and will probably be the ones to have to fix it.

dmick:
echoing what I said in chat, I think we should leave these builds enabled for now until a Ceph developer of some stripe can test a fix.

dmick:
Looking at https://github.com/Linaro/uadk.git, it doesn't look like that use of __FILE__ is conditioned on DEBUG

zmc:
Closing because we want to keep building on arm64 and ultimately to fix these issues.

PR #2404, opened 2025-07-17
scripts/build_utils: default to gcc-14 for ubuntu noble

cbodley:
noble provides both gcc-13 and 14, but its gcc 13.2.0 suffers from an LTO bug that causes ceph daemons to crash horribly (tracked in https://tracker.ceph.com/issues/63867)

### DIFF
--- a/scripts/build_utils.sh
+++ b/scripts/build_utils.sh
@@ -804,7 +804,7 @@ setup_pbuilder_for_old_gcc() {
         jammy)
             old=11;;
         noble)
-            old=13;;
+            old=14;;
     esac
     setup_gcc_hook $old > $hookdir/D10update-gcc-alternatives
     chmod +x $hookdir/D10update-gcc-alternatives


djgalloway (approved):
Fine with me

cbodley:
after this merged, i checked a [build log](https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=noble,DIST=noble,MACHINE_SIZE=gigantic/92739//consoleFull) and found:
```
I: user script /var/cache/pbuilder/build/635150/tmp/hooks/D10update-gcc-alternatives starting
update-alternatives: error: no alternatives for gcc
update-alternatives: error: alternative path /usr/bin/gcc-14 doesn't exist
update-alternatives: using /usr/bin/gcc-13 to provide /usr/bin/gcc (gcc) in auto mode
I: user script /var/cache/pbuilder/build/635150/tmp/hooks/D10update-gcc-alternatives finished
...
-- The CXX compiler identification is GNU 13.2.0
```
i think something somewhere needs to install `g++-14` on these machines. any idea where best to do that?

cbodley:
> https://github.com/ceph/ceph/blob/main/install-deps.sh#L63

thanks, testing in https://github.com/ceph/ceph/pull/64582

djgalloway:
> > > https://github.com/ceph/ceph/blob/main/install-deps.sh#L63
> > 
> > 
> > thanks, testing in [ceph/ceph#64582](https://github.com/ceph/ceph/pull/64582)
> 
> @djgalloway i don't think the package builds use install-deps.sh? this didn't help with https://jenkins.ceph.com/job/ceph-dev-new-build/ARCH=x86_64,AVAILABLE_ARCH=x86_64,AVAILABLE_DIST=noble,DIST=noble,MACHINE_SIZE=gigantic/92752//consoleFull

Oh.  Shoot.  Well, let's look at this programmatically.

ceph-dev-new-build calls `setup_pbuilder` which calls scripts/build_utils.sh' `setup_pbuilder` function: https://github.com/ceph/ceph-build/blob/main/ceph-dev-new-build/build/setup_pbuilder#L15

`setup_pbuilder` with `setup_gcc` does all this fun stuff: https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L626-L642 including calling `setup_pbuilder_for_ppa` https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L813-L821

If `use_ppa` is true, then `setup_pbuilder_for_new_gcc` gets called.  Looks like `use_ppa` only gets set to true on focal: https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L671-L690.  So I guess that needs to change to include noble.

NB: This will also set add `-DWITH_STATIC_LIBSTDCXX=ON` to the cmake args: https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L1039

And `local gcc_ver` should get set to 14 for noble? https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L816

https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L807 should probably get set back to 13 since that is the version shipped with the distro.

cbodley:
thanks @djgalloway 

> If `use_ppa` is true, then `setup_pbuilder_for_new_gcc` gets called. Looks like `use_ppa` only gets set to true on focal: https://github.com/ceph/ceph-build/blob/main/scripts/build_utils.sh#L671-L690. So I guess that needs to change to include noble.

noble ships with [gcc-13](https://packages.ubuntu.com/noble/gcc-13) and [gcc-14](https://packages.ubuntu.com/noble/gcc-14), so we don't need to add the toolchain ppa repo for access to it

i think `setup_pbuilder_for_old_gcc()` and its use of update-alternatives is the right way to prefer 14 over 13, but something else needs to guarantee that 14 is installed

one way to force installation of gcc-14 and g++-14 would be to update [debian/control](https://github.com/ceph/ceph/blob/main/debian/control#L23) from `Build-Depends: g++ (>= 11)` to `g++ (>= 14)`. we can do that eventually on main once we drop jammy support, but we still need to backport all of this ubuntu noble stuff to tentacle/squid which are stuck supporting both

djgalloway:
> > something else needs to guarantee that 14 is installed
> 
> what's responsible for deploying these ubuntu noble machines for jenkins? would it make sense to preinstall g++-14 there? i imagine we'd need something else to cover the build-in-containers case

The packages get built inside pbuilder so the underlying OS' gcc version doesn't matter.

cbodley:
> The packages get built inside pbuilder so the underlying OS' gcc version doesn't matter.

i see, so `setup_pbuilder_for_old_gcc()` is just adding a hook for commands that run 'inside'

`setup_pbuilder_for_new_gcc()` has a `D05install-new-gcc` hook that enables the toolchain repo then runs `apt-get install -y g++-$version`

maybe `setup_pbuilder_for_old_gcc()` needs a similar hook that just runs `apt-get install -y g++-$version`? that might be specific to noble, but probably wouldn't break anything for other distro verions

djgalloway:
> > maybe `setup_pbuilder_for_old_gcc()` needs a similar hook that just runs `apt-get install -y g++-$version`? that might be specific to noble, but probably wouldn't break anything for other distro verions

"probably" 😬 

> does #2405 look sane?

I would say that's not really the appropriate place to put it since we're not really setting up pbuilder for "old" gcc.

I think what probably makes more sense is to pass `use_ppa` as a parameter to `setup_pbuilder_for_new_gcc()`.  Then if true, add the ppa, if not, don't.  Then run the `env DEBIAN_FRONTEND=noninteractive apt-get install -y g++-$version` unconditionally?

cbodley:
> I would say that's not really the appropriate place to put it since we're not really setting up pbuilder for "old" gcc.

@djgalloway isn't use of PPA the only distinction between "old" vs "new"?
```bash
setup_pbuilder_for_ppa() {
    local hookdir=$1
    if use_ppa; then
        local gcc_ver=11
        setup_pbuilder_for_new_gcc $hookdir $gcc_ver
    else
        setup_pbuilder_for_old_gcc $hookdir
    fi
}
```
so setup_pbuilder_for_old_gcc feels like the right place for this logic. maybe renaming to `setup_pbuilder_for_ppa_gcc` vs `setup_pbuilder_for_distro_gcc` would help clarify this?

cbodley:
> I just meant because gcc-14 isn't the old one. Maybe I'm nitpicking.

@djgalloway i agree that the naming is confusing. we could either change the names, or combine the two functions into one with the ppa bits guarded by `if use_ppa`. i tend to think they're better as separate functions

cbodley:
> I'll review whatever you think is best.

thanks @djgalloway - i think https://github.com/ceph/ceph-build/pull/2405 is sufficient, if it works

PR #2407, opened 2025-07-22
ceph-dev-pipeline/build/Jenkinsfile: add tests for rocky

dmick:
There were a couple of conditionals for all the el distros that needed to also compare for the string 'rocky'


dmick:
close in favor of #2408

PR #2412, opened 2025-07-25
Build specific commits with new build pipelines

zmc:
With these changes:

- `ceph-trigger-build` will pass a `SHA1` parameter to the triggered job, set to the ID of the HEAD commit
- `ceph-dev-pipeline` will pass that to the setup job
- `ceph-source-dist` will clone that specific commit - if specified - otherwise it will clone the branch as it does now
- `ceph-dev-pipeline` will check that source tarballs represent the requested SHA1 - if one is supplied


zmc:
An example build: https://jenkins.ceph.com/job/ceph-dev-pipeline/425/

djgalloway (approved):

PR #2415, opened 2025-07-31
ceph-dev-new-build: add setup_container_runtime.sh

zmc:


dmick (approved):
surprised it's not already there

PR #2416, opened 2025-07-31
ceph-build-pull-requests: Use shellcheck

zmc:


zmc:
Test build here: https://jenkins.ceph.com/job/ceph-build-pull-requests-zmc/2

This produces _a lot_ of output at the moment, because it is warning about the same errors in `built_utils.sh` for very many of our jobs.

scripts/shellcheck_job.py
```diff
@@ -0,0 +1,68 @@
...
+import argparse
+import pathlib
+import subprocess
+import sys
+import xmltodict
+
+from typing import List, Tuple
+
+
+def parse_args(args: List[str]) -> argparse.Namespace:
```
phlogistonjohn, line 11: This is very nitpicky but since you use `
def process_job(job_obj: dict) -> bool` elsewhere I suggest sticking with either pre-3.9 `from typing import List, Dict, ...` or post-3.9 and use `dict`, `list`, `tuple`, etc as your type hints.  It's not wrong to mix them, but w/in a single file if you use one post-3.9 type hint you might as well do it everwhere else in that file :-)

phlogistonjohn (commented):
generally lgtm, I see no major issues

scripts/shellcheck_job.py
zmc: I'm glad you pointed out that inconsistency :)

scripts/shellcheck_job.py
```diff
@@ -0,0 +1,66 @@
...
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument(
+        "job_xml",
+        nargs="*",
+        help="The job XML file(s) to process",
+    )
+    return parser.parse_args(args)
+
```
phlogistonjohn, line 19: This spacing isn't pep8 but I'll let it slide if that's what you prefer ;-)

phlogistonjohn (approved):
looks ok to me.

scripts/shellcheck_job.py
zmc: good catch, I ran `black` and repushed.

phlogistonjohn (approved):
lgtm

PR #2417, opened 2025-07-31
builder.yml: Install wget

djgalloway:
image


dmick:
real users use curl, but whatever :D

dmick (approved):

PR #2419, opened 2025-08-01
builder.yml: Install bc

djgalloway:
Required in scripts/setup_container_runtime.sh


dmick:
putting up or shutting up: https://github.com/ceph/ceph-build/pull/2421

PR #2421, opened 2025-08-01
scripts/setup_container_runtime.sh: remove call to bc, can just use bash

dmick:
instead of https://github.com/ceph/ceph-build/pull/2419


djgalloway:
Even better: https://github.com/ceph/ceph-build/pull/2420

dmick:
already addressed in 2420.  you gotta be quick on this team.

PR #2422, opened 2025-08-01
build_utils.sh: Fix shellcheck errors

zmc:


scripts/build_utils.sh
```diff
@@ -1004,10 +1004,11 @@ build_debs() {
             $venv/chacractl binary ${chacra_flags} create ${chacra_endpoint}
 
         # extract cephadm if it exists
-        if [ -f release/${vers}/cephadm_${vers}*.deb ] ; then
-            dpkg-deb --fsys-tarfile release/${vers}/cephadm_${vers}*.deb | tar -x -f - --strip-components=3 ./usr/sbin/cephadm
+        for file in release/${vers}/cephadm_${vers}*.deb; do
```
phlogistonjohn, line 1007: Is it logically OK for this loop to execute for >1 matching file, even if that's unlikely?

phlogistonjohn (commented):
generally looks ok to me. I do have a question about the if being replaced by a loop.

scripts/build_utils.sh
djgalloway: Yeah, isn't that just going to extract cephadm from the last one if there are >1?

zmc: I was unsure, so I had inserted a `break` below

phlogistonjohn: I missed that, thanks.

PR #2423, opened 2025-08-05
Remove admin lists from ghprb configs

djgalloway:
These will be replaced by proper RABC eventually


dmick:
This won't stop us from administering because of our org privs, right?

dmick:
(I assume you mean "that no one not in the Ceph org should be able...")

and ok, agree it's just cruft

dmick (approved):

djgalloway:
https://jenkins.ceph.com/job/ceph-build-pull-requests/2656/

PR #2424, opened 2025-08-08
Adding readme for ceph-pull-requests, ceph-windows-pull-requests and ceph-pr-api jobs

tintumathew10:
Adding readme for ceph-pull-requests, ceph-windows-pull-requests and ceph-pr-api jobs

### DIFF
--- a/ceph-pr-api/README.rst
+++ b/ceph-pr-api/README.rst
@@ -1,0 +1,501 @@
+ceph-pr-api
+===========
+The ceph-pr-api job, listed as "ceph API tests" under pull requests in the Ceph repository, 
+is automatically triggered for all PRs as part of the CI/CD validation process. 
+https://github.com/ceph/ceph
+
+The ceph-api job runs integration tests in a vstart cluster 
+(helper script provided by Ceph located in the src directory of the Ceph source tree), 
+testing components like RADOS, RBD, RGW, and CephFS. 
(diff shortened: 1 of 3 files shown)


djgalloway:
While I appreciate the concept and effort, I don't think copying, pasting, and then commenting in the job config in a README file is a good idea.  The job configs will change and fall out of alignment with the READMEs.  Instead, I might suggest adding comments in the _actual_ job configs pointing to Jenkins Job Builder documentation.

I do like the workflow charts of how the jobs work, however.  I am curious to hear my colleagues' opinions.

dmick:
My initial thoughts were as David's....duplicating docs is not the right answer and these would be incredibly fragile wrt choices in the jobs, which do get changed.

What was the motivation for this PR?

deepssin:
@dmick When we started exploring the jobs to understand how upstream Jenkins operates — what each job does and how they interact — we found it challenging to get a clear picture without diving deeply into the job definitions. This slowed onboarding and cross-team understanding, especially for newer contributors.

The intent of this documentation is not to duplicate the configs verbatim, but to offer a high-level, human-readable reference that speeds up knowledge transfer. I agree we can adjust the format or location to avoid fragility — for example, by adding description/notes alongside the job code and parsing them to generate the documentation.

Given the potential to reduce onboarding time and improve cross-team awareness, I’d like to work together on refining the approach so it addresses your concerns while preserving the core goal — making upstream Jenkins more approachable.

PR #2430, opened 2025-08-21
setup_container_runtime.sh: Reset podman

zmc:
... if `podman system info` fails.


scripts/setup_container_runtime.sh
```diff
@@ -2,6 +2,7 @@
 # vim: ts=4 sw=4 expandtab
 function setup_container_runtime () {
   loginctl enable-linger "$(id -nu)"
+  (podman system info > /dev/null) || podman system reset --force
```
phlogistonjohn, line 5: Why the parens around `podman system info > /dev/null`?

zmc: no special reason. removed.

PR #2431, opened 2025-08-21
ceph-dev-pipeline: Do not require SHA1 property

zmc:
dd70a303 inadvertently made the SHA1 property mandatory; correct that.


djgalloway (commented):
Won't `def sha1_trimmed = env.SHA1.trim().toLowerCase()` still throw an exception if SHA1 isn't defined?

dmick:
bad comment before.  trim() wasn't failing before when SHA1 was empty, so I think it would only be a problem if it weren't defined at all.  The error before was the substring operation on a then-blank sha1.

zmc:
> Won't `def sha1_trimmed = env.SHA1.trim().toLowerCase()` still throw an exception if SHA1 isn't defined?

I am pretty sure the failure was happening at:
```
def ceph_builder_tag_short = "${env.BRANCH}.${env.DIST}.${ARCH}"
```
based on:
```
Found unhandled java.lang.StringIndexOutOfBoundsException exception:

Range [0, 7) out of bounds for length 0
```

I'll do a test build

dmick (approved):

PR #2433, opened 2025-08-21
ceph-trigger-build: Do not trigger on tag pushes

zmc:
This pipeline doesn't support tags, and it doesn't seem like shaman does either.


dmick (approved):
pretty much the same solution I'd come up with

PR #2439, opened 2025-09-03
Run make check in containers

phlogistonjohn:
Convert the arm64 and standard (amd64) CI jobs to execute make check in containers. This allows us to separate the build and test environment from the underlying CI node OS.


phlogistonjohn:
While we've tested this a bunch with a custom "preserve" job in jenkins I'm leaving this as a draft for now until we're sure this is "safe" to merge.

zmc (approved):

phlogistonjohn:
I'm fine either way: now or early next week.

zmc:
I think it might be smart to merge sooner rather than later, and depending on potential issues either fix them or revert

PR #2442, opened 2025-09-11
ceph-build-pull-requests: Use uv; drop OS restriction

zmc:


scripts/setup_uv.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# vim: ts=4 sw=4 expandtab
+
+function setup_pipx () {
+  command -v pipx || (
+    command -v apt && sudo apt install -y pipx
```
dmick, line 6: well who the hell knew newline was a bash pipeline delimiter?  Live and learn

PR #2451, opened 2025-09-24
Skip more tests if nothing relevant changed

zmc:


dmick:
do you wanna add tp ceph-pull-requests-arm64 as well

zmc:
it's in here

dmick (approved):

PR #2452, opened 2025-09-25
Switch to ceph-dev-pipeline for ceph.git builds, add Rocky10

djgalloway:


zmc:
Thanks! And do you think we could also populate the build description with metadata about the triggered builds, similar to ceph-trigger-build? As it stands now it's not possible to discover what was triggered from this job.

zmc:
I guess that's a limitation of the freestyle job type. Maybe in the future we could add a cron trigger to ceph-trigger-build an obsolete this job, since the logic already exists there.

PR #2454, opened 2025-10-01
Update jenkins ansible builder.yml for noble

dmick:
Really it's more about "updating to use a virtualenv for installing python modules", but noble's python/pip pretty much require it.


dmick:
(also updateded braggi18, which is how I found out about the ceph-build-pull-requests issue)

dmick:
I have not tested on anything but noble, no.  Wouldn't hurt.

dmick:
Tested on jammy and centos9 with no problem.

PR #2458, opened 2025-10-06
cephadm-dashboard-e2e: install docker from ubuntu's repo

nizamial09:
- add a cleanup script to remove the docker installation.


nizamial09:
i haven't tested this in the jenkins yet by updating the job configuration yet. Only done it in my local VM (which worked fine). Will wait for an approval before testing it out in the jenkins machines.

nizamial09:
> 1. there are build hosts that have installed docker in the standard way from the standard repos, and rely on it, so "is docker present, if so remove it" is going to break those hosts

@dmick i've modified the condition to check whether docker is installed through the docker official repo (docker-ce) and then only remove it. Maybe its more safer than before. Again I tried this out in my local VM which succesfully worked.

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
djgalloway: Might as well update this

djgalloway (dismissed):

ceph-dashboard-cephadm-e2e/config/definitions/ceph-dashboard-cephadm-e2e.yml
nizamial09: we test the nightly job against main and the supported branches and I think >=reef is still being supported right? so I left other branches in the black-list

```
- luminous
            - mimic
            - nautilus
            - pacific
            - quincy
            - octopus
            ```

PR #2459, opened 2025-10-06
Add login to docker.io for ceph-pull-requests*.  Remove old comment.

dmick:


ceph-pull-requests-arm64/build/build
```diff
@@ -40,6 +40,8 @@ bwc() {
         "${@}"
 }
 
+podman login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} docker.io
```
jmundack, line 43: document(comment) the fact that we need to login to avoid rate limiting - so that someone else doesn't try to remove it as its not "needed"

dmick: it's in the commit msg

PR #2461, opened 2025-10-06
Add bookworm as a buildable distro version

dmick:
Now that we are using ceph-dev-pipeline to build release images, we need to be able to request bookworm


dmick:
this requires additions to ceph.git:src/script as well (see https://github.com/ceph/ceph/pull/65804)

zmc (approved):

PR #2470, opened 2025-10-15
ceph-pull-requests: Add cleanup script

zmc:
We mostly just need to correct permissions that podman may have changed, so the workspace can be deleted by the next job.


dmick:
retest

PR #2471, opened 2025-10-16
ceph-dev-pipeline: Use correct builder variant

zmc:


djgalloway:
This probably should have at least been communicated to a mailing list or perhaps somehow had a soaking period where it was optional.  Many failed builds over the past couple weeks with `build-with-container.py: error: unrecognized arguments: --image-variant=packages`.

PR #2473, opened 2025-10-20
ceph-dev-pipeline: Updates for bwc image variants

zmc:


zmc:
Tested here: https://jenkins.ceph.com/job/ceph-dev-pipeline/1377
The two centos9 builds look good to me, but the noble build failed with missing dependencies. I am also testing this PR as part of this: https://github.com/ceph/ceph/pull/65994
@phlogistonjohn - do you have a guess as to why noble is attempting to use crimson here?

phlogistonjohn (approved):
Seems OK to me. The first patch does more in this language that I don't know but it seems fine. The latter two I am more confident about and also seem OK.

dmick:
won't we want a "--image-variant=packages.minimal" somewhere (or whatever the flag name is for "deps for a default build without crimson or make-check")?

PR #2475, opened 2025-10-23
setup_container_runtime.sh: prune ext. containers

zmc:
`podman system prune` will sometimes fail with a message like: "image is in use
by a container: consider listing external containers and force-removing image"


zmc:
The reason I didn't want to always do a `prune --all --force` is that we'd never benefit from previously cached image builds

dmick (approved):

PR #2477, opened 2025-10-24
ceph-dev-pipeline: check shaman for "are there binaries"

dmick:
Checking chacra isn't the right check, since this build's chacra url is not necessarily the prior successful build's chacra url (there are multiple chacra instances).  Instead, check shaman to see if there is a built repo; if so, the build can be skipped.  build_container will search shaman again to find the repo url to use to build the container, if necessary.

(so far completely untested; put up here for discussion)

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -338,9 +338,15 @@ pipeline {
             steps {
               script {
                 def os = get_os_info(env.DIST)
-                def shaman_endpoint = "https://shaman.ceph.com/api/repos/ceph/${env.BRANCH}/${env.SHA1}/${os.name}/${os.version_name}"
+                def shaman_endpoint = "https://shaman.ceph.com/api/repos/ceph/${env.BRANCH}/${env.SHA1}/${os.name}/${os.version_name}/"
                 def chacra_repo_url = sh(
-                  script: '''curl -s -S --retry 4 ${shaman_endpoint} | jq -r 'first(.[] | select(.flavor == "env.FLAVOR" and (.archs | any(. == "env.ARCH")) and .status=="ready")) | .chacra_url' ''',
+                  script: """curl -s -S --retry 4 ${shaman_endpoint} | 
+                      jq -r '
+                        first(.[] | select(.flavor == env.FLAVOR
+                        and (.archs | any(. == env.ARCH))
+                        and .status=="ready"))
+                      | .chacra_url'
+                  """,
                   returnStdout: true
                 ).trim()
                 if ( chacra_repo_url != "" && env.FORCE != "true" ) {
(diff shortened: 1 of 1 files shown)


djgalloway (changes requested):
Tabs vs spaces?
image

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -333,18 +337,13 @@ pipeline {
...
                 def os = get_os_info(env.DIST)
-                def chacra_endpoint = "ceph/${env.BRANCH}/${env.SHA1}/${os.name}/${os.version_name}/${env.ARCH}/flavors/${env.FLAVOR}/"
-                def chacractl_rc = sh(
-                  script: "$HOME/.local/bin/chacractl exists binaries/${chacra_endpoint}",
-                  returnStatus: true,
-                )
-                if ( chacractl_rc == 0 && env.FORCE != "true" ) {
+                def shaman_endpoint = "https://shaman.ceph.com/api/repos/ceph/${env.BRANCH}/${env.SHA1}/${os.name}/${os.version_name}"
+                def chacra_repo_url = sh(
+                  script: '''curl ${shaman_endpoint} | jq -r 'first(.[] | select(.flavor == "env.FLAVOR" and (.archs | any(. == "env.ARCH")) and .status=="ready")) | .chacra_url'''',
```
djgalloway: Maybe some error handling / retries for curl?

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -333,18 +337,13 @@ pipeline {
             }
             steps {
               script {
-                sh './scripts/setup_chacractl.sh'
```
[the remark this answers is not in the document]
dmick: oh very good point, good catch

ceph-dev-pipeline/build/Jenkinsfile
dmick: ok will add -s -S (bar doesn't confront me but it's useless).  --fail...well the worst that happens is that we build packages we might not have had to build, so I think failing the job is too far

djgalloway:
Can you push this to 2.jenkins and run a test build

dmick:
oh absolutely I'll test; testing on a branch now: https://jenkins.ceph.com/job/ceph-dev-pipeline/1797/

dmick:
well I think I have the syntax for the curl | jq right now, but it's exposed another problem:

https://github.com/ceph/ceph-build/blob/main/scripts/build_container#L16-L24

when we skip the build with this mechanism, CI_COMPILE is not turned off, so the test above still tests for jenkins_build_url matching the current build URL, which is no more true than it is when CI_COMPILE is off.

I need to ponder what the extra test, if any, needs to be, with a fresher mind.

djgalloway:
https://github.com/ceph/ceph-build/commit/1d0aecec80a479834e0e88d205ff3e5a40117cb6

PR #2485, opened 2025-11-03
ceph-trigger-build: Fix JENKINS-37984

djgalloway:
I have no idea what changed or how we just got bit by this.  Jenkins wasn't updated.

https://issues.jenkins.io/projects/JENKINS/issues/JENKINS-37984


gregsfortytwo (approved):
I don’t really get why this matters, but it matches the Jenkins upstream report workaround and we need builds to run — LGTM.

PR #2493, opened 2025-11-10
[WIP] Include Crimson binary by default (Centos)

Matan-B:
### Pushed for **early** discussion purposes

Dependencies:
* https://github.com/ceph/ceph/pull/65782 
* https://github.com/ceph/ceph-build/pull/2494

---

Once https://github.com/ceph/ceph/pull/65782 is merged, we would be able to build
Crimson (ceph-osd-crimson) as part of the default build flavor with
rpm builds.

ceph-osd-crimson and ceph-osd-classic are selected based on their
priorities.
ceph-osd-classic has the highest priority of 100 when installed.
Crimson could only be used if the following is run:
```
update-alternatives --set ceph-osd /usr/bin/ceph-osd-crimson
```

---

crimson-debug flavor is remained since the project uses it for testing

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -200,7 +200,7 @@ pipeline {
           }
           axes {
             name 'FLAVOR'
-            values 'default', 'crimson-release', 'crimson-debug'
+            values 'default', 'crimson-debug'
           }
         }
         // crimson is only supported on centos9 x86_64
@@ -208,7 +208,7 @@ pipeline {
           exclude {
             axis {
               name 'FLAVOR'
-              values 'crimson-release', 'crimson-debug'
+              values 'crimson-debug'
             }
             axis {
               name 'DIST'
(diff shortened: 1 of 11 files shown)


ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -443,14 +442,15 @@ pipeline {
                 def deb_build_profiles = "";
                 switch (env.FLAVOR) {
                   case "default":
-                    ceph_extra_cmake_args += " -DALLOCATOR=tcmalloc"
+                    ceph_extra_cmake_args += " -DALLOCATOR=tcmalloc -DWITH_CRIMSON=true"
```
cbodley: maybe it's time for [CMakeLists.txt](https://github.com/ceph/ceph/blob/cf11bef497/CMakeLists.txt#L740) to enable this option by default?

[ceph.spec.in](https://github.com/ceph/ceph/blob/cf11bef497/ceph.spec.in#L94) could do similar with
```diff
-%bcond_with crimson
+%bcond_without crimson
```
that way ceph-build doesn't need to do anything special for crimson

Matan-B: Makes sense, cmake could also verify that the compiler used is capable of building crimson and the distro used. I'll follow up on this once the dependency packaging PR is merged. Thanks!

PR #2497, opened 2025-11-23
build flavors: Cleanup "crimson", introduce "debug"

Matan-B:
With https://github.com/ceph/ceph/pull/66229 merged,
Crimson is now included (though not used) by default in our RPM builds.
This means the existing default flavor can also be used for Crimson testing
by selecting Crimson as the default OSD package.

Notes:

* The previous workaround related to DWITH_STATIC_LIBSTDCXX is no longer
  relevant for Crimson (it was tied to older compiler issues).

* The crimson-only branch name selection is also cleaned up,
  as centos9-only can now be used instead.

* This change breaks Crimson Tentacle CI builds:
  The packaging update that includes Crimson in RPM builds was not backported to Tentacle.
  Tentacle builds would still require a dedicated flavor that enables WITH_CRIMSON
  However, since Crimson changes have not been backported to Tentacle (since the first RCA),
  there is no strong reason to keep building and testing the same Crimson HEAD.
  So we can use this opprtuinity to stop nightly Crimson/Tentacle builds and tests.

  See last Crimson tentacle run, (which is not expected to change):
  https://pulpito.ceph.com/teuthology-2025-11-22_22:56:11-crimson-rados-tentacle-distro-crimson-debug-smithi/

### DIFF
--- a/ceph-dev-new-setup/build/build
+++ b/ceph-dev-new-setup/build/build
@@ -86,6 +86,17 @@ else
   printf 'No cmake debug options added to branch %s.\n' "$BRANCH"
 fi
 
+# Tentacle is the last release that needs dedicated Crimson builds,
+# Later releases are able to use Crimson with the default build.
+# As the "Crimson flavor" is no longer available, we need a *temporary* way
+# to be able build Crimson for tentacle.
+# Note: This could be removed once Crimson we have Umbrella release builds.
+if [[ "$BRANCH" == *-crimson-tentacle ]]; then
+  CEPH_EXTRA_RPMBUILD_ARGS="--with crimson"
+  CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_CRIMSON=true"
+  printf 'Added Crimson Tentacle cmake configs to branch %s. CEPH_EXTRA_CMAKE_ARGS: %s\n' "$BRANCH" "$CEPH_EXTRA_CMAKE_ARGS"
+fi
+
 ceph_build_args_from_flavor ${FLAVOR}
 
 mkdir -p release


ceph-dev-new-setup/build/build
```diff
@@ -86,6 +86,18 @@ else
...
+# Tentacle is the last release that needs dedicated Crimson builds,
+# Later releases are able to use Crimson with the default build.
+# As the "Crimson flavor" is no longer available, we need a *temporary* way
+# to be able build Crimson for tentacle.
+# Note: This could be removed once Crimson we have Umbrella release builds.
+if [[ "$BRANCH" == *-crimson-tentacle ]]; then
+  CEPH_EXTRA_RPMBUILD_ARGS="--with crimson"
+  CEPH_EXTRA_CMAKE_ARGS+=" -DWITH_CRIMSON=true"
+  printf 'Detected Crimson Tentacle build \n'
+  printf 'Added debug cmake configs to branch %s. CEPH_EXTRA_CMAKE_ARGS: %s\n' "$BRANCH" "$CEPH_EXTRA_CMAKE_ARGS"
```
tchaikov: this message could be misleading. as we are not adding "debug" cmake options, but enabling crimson for the branch to be built.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -246,8 +238,6 @@
 - project:
     name: ceph-perf
     osd-flavor:
```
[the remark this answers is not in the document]
Matan-B: I've missed the fact that "ceph-perf" has its own osd-flavor definition. I've kept crimson/classic. Thanks!

ceph-dev-new-setup/build/build
```diff
@@ -86,6 +86,17 @@ else
   printf 'No cmake debug options added to branch %s.\n' "$BRANCH"
 fi
 
+# Tentacle is the last release that needs dedicated Crimson builds,
+# Later releases are able to use Crimson with the default build.
+# As the "Crimson flavor" is no longer available, we need a *temporary* way
+# to be able build Crimson for tentacle.
+# Note: This could be removed once Crimson we have Umbrella release builds.
```
batrick: We will still need to be able to build Tentacle until upgrade testing from T to V is no longer necessary (probably when V is EOL).

Matan-B: This section is only true for Crimson. We only build the last stable release (and main) as there's no upgrade testing/support yet. Starting U we will probably start working on upgradability. We still break on-disk data structures from time to time as we are still flexible with this - the plan is to stop with this approach starting U.
This is why we should be able to remove this bit once U is out as Crimson will no longer be supported in T _at all_. I'll mention this in this comment to avoid confusion.

djgalloway:
@tchaikov Do you want to review this again or is it good to go?

tchaikov:
@djgalloway hi David, the failure is not related to this change. and i don't think it's transient:
```
Building remotely on [ec2-44-223-66-26.compute-1.amazonaws.com](https://jenkins.ceph.com/computer/ec2-44-223-66-26.compute-1.amazonaws.com/) (buster jammy ec2 arm64 huge installed-os-jammy xenial) in workspace /home/jenkins-build/build/workspace/ceph-build-pull-requests
...
/tmp/jenkins7947793624570378365.sh: line 1891: warning: setlocale: LC_ALL: cannot change locale (en_us.UTF-8): No such file or directory
...
+ for reqs in $WORKSPACE/ansible/requirements/*
+ ./.venv/bin/ansible-galaxy install -r /home/jenkins-build/build/workspace/ceph-build-pull-requests/ansible/requirements/sensu-requirements.yml -p /home/jenkins-build/build/workspace/ceph-build-pull-requests/ansible/roles --force
ERROR: Ansible could not initialize the preferred locale: unsupported locale setting
```
see https://jenkins.ceph.com/job/ceph-build-pull-requests/2909/consoleFull#19485512547dc3600c-28d8-42e2-9653-291a0c275fdf

probably we need to generate this locale in ceph-cm-ansible:role/testnode ?

ceph-dev-new-setup/build/build
```diff
@@ -86,6 +86,17 @@ else
...
+if [[ "$BRANCH" == *-crimson-tentacle ]]; then
```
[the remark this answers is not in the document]
djgalloway: @batrick Are these suggestions 'nice to haves' or must haves?  I think CentOS9 builds are blocked by this PR not being merged.

PR #2498, opened 2025-12-05
All: Limit PR checks to Ceph org

djgalloway:
Long overdue


dmick:
I looked at this when you filed it, but I still don't understand what the change will do to the workflow.  Does this mean that PRs from org members run the checks, but PRs from non-org members must have the checks triggered manually by someone in response to some notification somewhere?

djgalloway:
> I looked at this when you filed it, but I still don't understand what the change will do to the workflow. Does this mean that PRs from org members run the checks, but PRs from non-org members must have the checks triggered manually by someone in response to some notification somewhere?

Correct

PR #2500, opened 2025-12-09
Change quay-quay-quay.... to quay.ceph.io

dmick:
The external name works from locations outside the existing sepia lab, and there's no good reason not to use the public name.


dmick:
(already tested in a hacked ceph-dev-pipeline build, and works fine.)

PR #2511, opened 2025-12-19
builder.yml: Ensure consistent subuid and subgid across builders

djgalloway:
This might explain a lot of the storage weirdness we've been observing.  If a container image layer is create in rootless on podman on one machine, that image layer is restored to another host with *different* subuid or subgid, errors can occur.

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -785,8 +785,14 @@
       become_user: "{{ jenkins_user }}"
       when: ansible_os_family == "Suse"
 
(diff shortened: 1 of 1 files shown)


dmick:
mentioned in slack, but adding here for completeness:

It seems to be the case that "container images" (like you'd push) do not contain any info about the subuid/subgid mapping used to create them; they have container-internal UIDs in them that will get mapped when instantiated into a container.  Do you remember what it was that broke exactly?

djgalloway:
https://github.com/ceph/ceph-build/pull/2572

PR #2539, opened 2026-02-17
Add release-tracker-workflow pipeline for Shaman-based RC testing

deepssin:
This adds a Jenkins pipeline to automate RC testing for release trackers: resolve or pass a Ceph SHA1, wait for it on Shaman, schedule teuthology suites (in parallel), aggregate results, and optionally post to Redmine and send email. There is no build step; the pipeline assumes the SHA1 is already available on shaman.
Note: The job needs to run on a teuthology agent (that means we need to add teuthology-node as an agent). The pipeline is parameterized (e.g. AGENT_LABEL, teuthology paths, Paddles/Pulpito URLs) so it can be adapted to different environments.

This is meant as a starting point for discussion—defaults and layout may need to change depending on how we want to run this in production (e.g. which Jenkins, which agents, credentials). Happy to adjust based on feedback.

### DIFF
--- a/release-tracker-workflow/README.rst
+++ b/release-tracker-workflow/README.rst
@@ -1,0 +1,44 @@
+Release Tracker Workflow
+========================
+
+A Jenkins pipeline that automates RC (release candidate) testing for Ceph release trackers:
+
+- Resolve or accept a Ceph SHA1 (optional **CEPH_SHA1**; must exist on Shaman).
+- Wait for the SHA1 on Shaman, then schedule teuthology suites (all triggered at once, then wait for all).
+- Aggregate pass/fail results and post to Redmine (tracker.ceph.com) and/or send email.
+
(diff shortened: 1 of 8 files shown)


deepssin:
> I guess I'm fine with this conceptually. Has it been tested anywhere? I'd like to see it working.

I am working on getting logs for this, will share ASAP

dmick:
fwiw, this works too:

`cat testsock.py`
```
import socket
import ssl

c = ssl.create_default_context()
h = 'paddles-paddles.apps.pok.os.sepia.ceph.com'
with socket.create_connection((h,443)) as sock:
    with c.wrap_socket(sock, server_hostname=h) as ssock:
        print(ssock.version())


```
strace shows that it opens /etc/ssl/certs/ca-certificates.crt

djgalloway (approved):
Pretty slick.  Fine with me.

PR #2566, opened 2026-04-07
ceph-perf: tighten PR regression gate

knrt10:
This describes the perf PR job as a release regression check, pin the benchmark definition to ceph-main, and surface the perf regression status in GitHub. This also reenables the ceph-perf-pull-request

Fixes: https://tracker.ceph.com/issues/75665

### DIFF
--- a/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
+++ b/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
@@ -49,6 +49,7 @@
     name: run-cbt
     builders:
       - shell: |
+          set -euxo pipefail
           cd {src-dir}
           archive_dir={archive-basedir}/$(git rev-parse --short HEAD)
           if test -d $archive_dir ; then
@@ -58,24 +59,19 @@
           export FOR_MAKE_CHECK=true
           cxx_compiler=g++
           c_compiler=gcc
-          for i in $(seq 15 -1 10); do
+          for i in $(seq 22 -1 19); do
               if type -t clang-$i > /dev/null; then
                   cxx_compiler="clang++-$i"
                   c_compiler="clang-$i"
                   break
               fi
           done
-          if test {osd-flavor} = "crimson" ; then
-              export WITH_CRIMSON=true
-              # TODO use clang-10 on ubuntu/focal
-              timeout 7200 src/script/run-make.sh \
-                --cmake-args "-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF" \
-                vstart-base crimson-osd
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml
-          else
-              timeout 7200 src/script/run-make.sh --cmake-args "-DCMAKE_BUILD_TYPE=Release -DWITH_TESTS=OFF" vstart-base
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml --classical
-          fi
+          export WITH_CRIMSON=true
+          cmake_args="-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF $CEPH_PERF_EXTRA_CMAKE_ARGS"
+          timeout 7200 src/script/run-make.sh \
+            --cmake-args "$cmake_args" \
+            vstart-base
+          src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt $WORKSPACE/cbt -a $archive_dir {benchmark-yaml}
 
 - builder:
     name: compare-cbt-results
@@ -105,18 +101,19 @@
             --summary ${{result}}                        \
             --text report.md
 
-- job-template:
-    name: 'ceph-perf-{osd-flavor}'
+- job:
+    name: ceph-perf
     project-type: freestyle
     defaults: global
     concurrent: true
-    # use lastest rhel and ubuntu for crimson for clang build
     node: performance
-    display-name: 'ceph: {osd-flavor} perf test'
+    display-name: 'ceph: perf regression test'
+    description: 'Release CBT regression check for Crimson: builds ceph-pr and ceph-main with WITH_CRIMSON=ON, runs the checked-in radosbench 4K read workload, and compares results via cbt/compare.py. The GitHub check fails when compare.py reports a statistically worse result, too much noise, or errors.'
     quiet-period: 5
     block-downstream: false
     block-upstream: false
     retry-count: 3
+    disabled: false
     check-app-id: "62865"
     check-install-id: "8465036"
     check-name: "perf-test"
@@ -138,22 +135,29 @@
       - string:
           name: ghprbPullId
           description: "the GitHub pull id, like '72' in 'ceph/pull/72'"
+      - string:
+          name: CEPH_PERF_EXTRA_CMAKE_ARGS
+          default: ''
+          description: 'Extra CMake arguments appended to the Crimson build (use this to disable additional components for a more minimal build).'
 
     triggers:
       - github-pull-request:
           allow-whitelist-orgs-as-admins: true
           org-list:
             - ceph
-          trigger-phrase: 'jenkins test {osd-flavor} perf'
+          trigger-phrase: 'jenkins test perf'
           skip-build-phrase: '^jenkins do not test.*'
           only-trigger-phrase: false
           white-list-labels:
             - performance
-            - crimson
           github-hooks: true
           permit-all: false
           auto-close-on-fail: false
           cancel-builds-on-update: true
+          status-context: 'ceph perf regression'
+          started-status: 'running ceph perf regression'
+          success-status: 'ceph perf regression passed'
+          failure-status: 'ceph perf regression failed'
 
     scm:
       - ceph-main
@@ -166,7 +170,7 @@
           . /etc/os-release || ID=ubuntu
           case $ID in
           debian|ubuntu)
-              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable clang-12
+              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable
               ;;
           centos|rhel)
               sudo dnf copr remove tchaikov/llvm-toolset-10 || true
@@ -209,12 +213,12 @@
 
       - run-cbt:
           src-dir: "ceph-main"
-          osd-flavor: '{osd-flavor}'
+          benchmark-yaml: "$WORKSPACE/ceph-main/src/test/crimson/cbt/radosbench_4K_read.yaml"
           # ideally cbt-results should be persited across jobs, so the test result can be reused
           archive-basedir: "$WORKSPACE/cbt-results"
       - run-cbt:
           src-dir: "ceph-pr"
-          osd-flavor: '{osd-flavor}'
+          benchmark-yaml: "$WORKSPACE/ceph-pr/src/test/crimson/cbt/radosbench_4K_read.yaml"
           # use the basedir of git checkout, so it can be wiped
           archive-basedir: "$WORKSPACE/ceph-pr"
       - compare-cbt-results:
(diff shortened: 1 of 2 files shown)


ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -154,6 +156,10 @@
           permit-all: false
           auto-close-on-fail: false
           cancel-builds-on-update: true
+          status-context: 'ceph perf regression ({osd-flavor})'
```
Matan-B: Do we still need `osd-flavor`? Crimson no longer use dedicated flavor builds.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -71,10 +71,10 @@
               timeout 7200 src/script/run-make.sh \
                 --cmake-args "-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF" \
                 vstart-base crimson-osd
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml
+              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir {benchmark-yaml}
```
Matan-B: run-cbt is outdated with crimson usage and still relies on cyanstore, which is no longer relevant for our purposes.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -71,10 +71,10 @@
...
           else
               timeout 7200 src/script/run-make.sh --cmake-args "-DCMAKE_BUILD_TYPE=Release -DWITH_TESTS=OFF" vstart-base
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml --classical
+              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir {benchmark-yaml} --classical
```
Matan-B: I don't think we need two build types, crimson usage can be selected based on the OSD starting params.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
Matan-B: compiler used here is outdated.

Matan-B (changes requested):
Few pieces still missing here. I’m not sure this would work with older compilers and exiting CBT support. I would suggest trying to run some of the tests locally first to see if the numbers make sense before we start blocking PRs with this job.
Is this check required once enabled?

We should probably also add some details to Crimson’s dev guide, how the test works, when a PR is blocked (under which cases)  and where the results are stored. This should help devs mitigate failures.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -58,24 +58,18 @@
           export FOR_MAKE_CHECK=true
           cxx_compiler=g++
           c_compiler=gcc
-          for i in $(seq 15 -1 10); do
+          for i in $(seq 18 -1 12); do
```
Matan-B: Crimson would FTBFS with clang less than 19

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -58,24 +58,18 @@
...
-                --cmake-args "-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF" \
-                vstart-base crimson-osd
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml
-          else
-              timeout 7200 src/script/run-make.sh --cmake-args "-DCMAKE_BUILD_TYPE=Release -DWITH_TESTS=OFF" vstart-base
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml --classical
-          fi
+          export WITH_CRIMSON=true
+          timeout 7200 src/script/run-make.sh \
+            --cmake-args "-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF" \
```
Matan-B: We should be able to disable more components if we would like to have a minimal build.

Matan-B (changes requested):
I'm not sure I understand if this is ready for review or was tested locally.
If it was, can you please share examples of how the results comparison work?


The clang version that was already mentioned in the previous review would still FTBFS.
CBT is still not updated for the relevant usage with Crimson.
Under which cases a PR is considered blocked (dev docs)?
It is also not clear what the strategy is here - is this intended to run only against Crimson PRs?

Please make sure there is a solid plan for how this is going to be implemented. Starting with a draft PR is also a good approach, but please be clear about the draft stage (what is missing, what is not yet tested/verified and what the next steps are or what is working so far). Otherwise, this PR does not seem ready for review.

My suggestion here would be to enable this check as a non-required PR check and iterate on it until it can be re-enabled as a required one. Alternatively, running parts of the jobs locally to make sure all is working as expected.
I would be happy to help in any way if you have any questions.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -58,24 +59,19 @@
...
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml
-          else
-              timeout 7200 src/script/run-make.sh --cmake-args "-DCMAKE_BUILD_TYPE=Release -DWITH_TESTS=OFF" vstart-base
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml --classical
-          fi
+          export WITH_CRIMSON=true
+          cmake_args="-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF $CEPH_PERF_EXTRA_CMAKE_ARGS"
+          timeout 7200 src/script/run-make.sh \
+            --cmake-args "$cmake_args" \
+            vstart-base
```
tchaikov: why did you drop `crimson-osd` from the build target list? please note `vstart-base` alone does not build `crimson-osd`.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -209,12 +213,12 @@
...
       - run-cbt:
           src-dir: "ceph-main"
-          osd-flavor: '{osd-flavor}'
+          benchmark-yaml: "$WORKSPACE/ceph-main/src/test/crimson/cbt/radosbench_4K_read.yaml"
           # ideally cbt-results should be persited across jobs, so the test result can be reused
           archive-basedir: "$WORKSPACE/cbt-results"
       - run-cbt:
           src-dir: "ceph-pr"
-          osd-flavor: '{osd-flavor}'
+          benchmark-yaml: "$WORKSPACE/ceph-pr/src/test/crimson/cbt/radosbench_4K_read.yaml"
```
tchaikov: i think both runs should use the same `yaml` file, so that it's an apple-to-apple comparison.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -58,24 +59,19 @@
...
-              # TODO use clang-10 on ubuntu/focal
-              timeout 7200 src/script/run-make.sh \
-                --cmake-args "-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF" \
-                vstart-base crimson-osd
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml
-          else
-              timeout 7200 src/script/run-make.sh --cmake-args "-DCMAKE_BUILD_TYPE=Release -DWITH_TESTS=OFF" vstart-base
-              src/script/run-cbt.sh --build-dir $PWD/build --source-dir $PWD --cbt ${{WORKSPACE}}/cbt -a $archive_dir src/test/crimson/cbt/radosbench_4K_read.yaml --classical
-          fi
+          export WITH_CRIMSON=true
```
tchaikov, line 84: does crimson require a certain version of gcc? i am asking because, if clang-19..22 is not installed, `cxx_compiler` stays `g++` . and IIRC, seastar is only compatible with the latest two major versions of GCC and Clang.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -166,7 +170,7 @@
           . /etc/os-release || ID=ubuntu
           case $ID in
           debian|ubuntu)
-              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable clang-12
+              sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y python3-yaml python3-lxml python3-prettytable
```
tchaikov: `clang-12` is dropped, but do we need to install a updated clang here?

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -49,32 +49,59 @@
...
+          git submodule update --init --recursive
+          python3 - "$PWD/src/test/crimson/cbt/t2c.py" <<'PY'
+          import pathlib
+          import sys
+          p = pathlib.Path(sys.argv[1])
+          s = p.read_text()
+          old = "teuthology_config = yaml.load(input)"
+          new = "teuthology_config = yaml.safe_load(input)"
+          if old in s:
+              p.write_text(s.replace(old, new, 1))
```
perezjosibm: I'd suggest to have this change in the actual source module rather than dynamically changing it during build. That change to t2c.py (teuthology to cbt yaml translation) might include some doc unit tests.

perezjosibm (approved):
LGTM, I'd suggest to move the dynamic change on the t2c.py into the main codebase instead, (probably accompanied by unit tests of the module itself).
Update to the documentation for the change introduced in this PR would be very much appreciated.
Many thanks

knrt10:
@perezjosibm done https://github.com/ceph/ceph/pull/69344

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -49,32 +49,49 @@
...
+                  if test "$gcc_major" -lt 13 || test "$gxx_major" -lt 13; then
+                      echo "Need clang-19+ or GCC/G++ 13+ for Crimson build (found gcc=$gcc_major g++=$gxx_major)" >&2
+                      exit 1
+                  fi
+                  cxx_compiler=g++
+                  c_compiler=gcc
+              fi
               export WITH_CRIMSON=true
-              # TODO use clang-10 on ubuntu/focal
+              cmake_args="-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF -DWITH_SPDK=OFF -DWITH_DPDK=OFF -DWITH_QATLIB=OFF -DWITH_QATZIP=OFF $CEPH_PERF_EXTRA_CMAKE_ARGS"
```
Matan-B: Is there any reason to introduce a Crimson separated build? We should be able to include Crimson by default as we already do in CI. The cbt test itself could vary (`run-cbt.sh --classical` or `--crimson`

Matan-B (commented):
Nice!

Is there new check marked as "required" to merge PR? I think we can merge a non required version so we could test the changes here and keep iterating if needed.

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
knrt10: Make sense, addressed change

perezjosibm (approved):
LGTM, many thanks!

PR #2568, opened 2026-04-13
Add teuthology-nightly-cadence (and trigger) plus teuthology-runner

deepssin:
This enables the nightly crons to run from a jenkins pipeline. 

Here are the benefits of putting this in pipeline instead of running from crontab - 

- Schedules live in Git (JJB YAML) with PR review, not only on a machine’s crontab
- Runs are visible in Jenkins: history, logs, and what triggered each build 
- Less tied to one host so no dependency on a single sepia account
- Easier reruns and experiments from the UI without SSH or editing live cron.

### DIFF
--- a/ceph-release-containers/build/Jenkinsfile
+++ b/ceph-release-containers/build/Jenkinsfile
@@ -102,6 +102,7 @@ pipeline {
                 ./make-manifest-list.py --version ${VERSION}
               else
                 ./make-manifest-list.py
+              fi
               '''
           }
         }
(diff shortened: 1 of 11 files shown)


djgalloway:
Makes sense to me conceptually.  Have you pushed the jobs manually to test in Jenkins?

deepssin:
> Makes sense to me conceptually. Have you pushed the jobs manually to test in Jenkins?

@djgalloway here is one sample log - https://jenkins.ceph.com/job/preserve-teuthology-nightly-cadence-trigger/9/console

djgalloway (approved):

PR #2578, opened 2026-05-11
jenkins-job-builder: pin setuptools for JJB on Python 3.12

deepssin:
Stevedore still imports pkg_resources when loading jenkins-jobs CLI extensions. setuptools>=82 can leave that module unavailable in fresh venvs on newer agents, causing No module named pkg_resources and No jjb.cli.subcommands extensions found.

Install setuptools<82 after bootstrap so jenkins-jobs works on both Python 3.10 and 3.12 workers until upstream drops pkg_resources.

### DIFF
--- a/jenkins-job-builder/build/build
+++ b/jenkins-job-builder/build/build
@@ -16,7 +16,10 @@ set -euxo pipefail
 # when jenkins-job-builder 6.5 or later is released, this requirement should change to
 # jenkins-job-builder>=6.5
 
-pkgs=( "dataclasses" "jenkins-job-builder>=6.4.3" )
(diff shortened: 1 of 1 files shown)


jenkins-job-builder/build/build
djgalloway: Can't you just add `"setuptools<82"` here or does it need to be installed after jjb?

deepssin: I have added to pkgs. Figured that the install is sequential so that `jjb` doesn't override the `setuptools` install so placed it after JJB in that list.

zmc (approved):

PR #2582, opened 2026-05-15
ceph-pr-commits: support target branch in docs commit summary

bluikko:
ceph/ceph@0a54fcdfc491ce2b2bb3ded77e319a7cff785e73 added a [new policy](https://github.com/ceph/ceph/blame/main/SubmittingPatches-backports.rst#L48) that backport PRs with not-cherry-picked commits need the commit message summary to begin with the target branch name.
This policy conflicted with docs-only commits that must have the same line start with `doc`, as enforced by this job.

Modify the regex used to allow an additional optional `targetbranch: ` string in front of the `doc` string in the beginning of the commit message summary, only if `target_branch` is something else than `main` and allow only the target branch name.

This is a rather naive attempt and the same result could have been done differently so feedback is appreciated.

In essence it allows commit message summary lines such as:
```console
tentacle: doc: fix foo
```

### DIFF
--- a/ceph-pr-commits/build/test_commits.py
+++ b/ceph-pr-commits/build/test_commits.py
@@ -39,7 +39,12 @@ class TestCommits(object):
 
     @pytest.mark.doc_test
     def test_doc_title(self):
-        doc_regex = '^doc'
+        if self.target_branch == 'main':
+            doc_regex = '^doc'
+        else:
+            # the title of a non-cherry-pick commit starts with the target branch name, like
+            # "squid: doc: ...." instead of "doc: ..."
+            doc_regex = f'^({self.target_branch}: )?doc'
         all_commits = f'git log -z --no-merges --pretty=format:%s origin/{self.target_branch}..{self.source_branch}'
         wrong_commits = list(filterfalse(
             re.compile(doc_regex).search,
--- a/ceph-pr-commits/build/test_commits.py
+++ b/ceph-pr-commits/build/test_commits.py
@@ -39,9 +39,8 @@ class TestCommits(object):
 
     @pytest.mark.doc_test
     def test_doc_title(self):
-        doc_regex = '\nDate:[^\n]+\n\n    doc'
-        all_commits = 'git log -z --no-merges origin/%s..%s' % (
-            self.target_branch, self.source_branch)
+        doc_regex = '^doc'
+        all_commits = f'git log -z --no-merges --pretty=format:%s origin/{self.target_branch}..{self.source_branch}'
         wrong_commits = list(filterfalse(
             re.compile(doc_regex).search,
             self.command(all_commits).split('\0')))


bluikko:
**Comprehensive testing of this change has not been done**. I have only done basic testing in a local Python REPL as I do not know how this could be tested!

ceph-pr-commits/build/test_commits.py
```diff
@@ -39,7 +39,8 @@ def setup_class(cls):
 
     @pytest.mark.doc_test
     def test_doc_title(self):
-        doc_regex = '\nDate:[^\n]+\n\n    doc'
+        doc_regex = '\nDate:[^\n]+\n\n    %sdoc' % ('' if
+            self.target_branch == 'main' else ('(%s: )?' % (self.target_branch)))
```
tchaikov: i'd suggest take this opportunity to simplify the command line and regex. so we have two commits:

1. to replace `git log -z --no-merges` with `git log --pretty=format:%s`, and replace `'\nDate:[^\n]+\n\n    doc'` with `^doc`. to simplify the "doc" prefix detection
2. to adapt to the new policy,
    ```py
    if self.target_branch == 'main':
      doc_regex = '^doc'
    else:
      # the title of a non-cherry-pick commit starts with the target branch name, like
      # "squid: doc: ...." instead of "doc: ..."
      doc_regex = f'^({self.target_branch}: )? doc'
    ```
please note, i add a comment to explain the rationales behind the branch prefix.

bluikko: I've pushed another commit following your suggestion. I removed the latter (extra) space in `: )? doc'`.

Thank you for the review.

ceph-pr-commits/build/test_commits.py
```diff
@@ -39,9 +39,13 @@ def setup_class(cls):
...
-        doc_regex = '\nDate:[^\n]+\n\n    %sdoc' % ('' if
-            self.target_branch == 'main' else ('(%s: )?' % (self.target_branch)))
-        all_commits = 'git log -z --no-merges origin/%s..%s' % (
+        if self.target_branch == 'main':
+            doc_regex = '^doc'
+        else:
+            # the title of a non-cherry-pick commit starts with the target branch name, like
+            # "squid: doc: ...." instead of "doc: ..."
+            doc_regex = f'^({self.target_branch}: )?doc'
+        all_commits = 'git log -z --no-merges --pretty=format:%s origin/%s..%s' % (
```
tchaikov: this does not work. as this will throw:
```
TypeError: not enough arguments for format string
```
i'd suggest use f-string to format the string. and transpose these two commits for better reviewability.

bluikko: Right, of course. I did not think about the first `%s`.

bluikko: I've cleaned up the two commits. Thank you for your time & feedback so far.

tchaikov: > transpose these two commits for better reviewability.

could you reconsider this suggestion?

tchaikov (approved):
lgtm. but we can still reorder these two commits for a cleaner git history. this is far from a blocker.

tchaikov:
i tested this change manually with two test cases:
```
WORKSPACE=$HOME/dev ghprbTargetBranch=main pytest 
```
1. after adding a commit whose title starts with "doc" on top of origin/main, the test passes
2. after adding a commit whose title does not start with "doc" on top of origin main, the test fails.



> lgtm. but we can still reorder these two commits for a cleaner git history. this is far from a blocker.

@bluikko i don't insist on this part. so, if you would like get it merged without this change, i will merge your pull request as it is, otherwise, i will rereview and retest with your change after you update this pull request.

ceph-pr-commits/build/test_commits.py
```diff
@@ -39,7 +39,12 @@ def setup_class(cls):
 
     @pytest.mark.doc_test
     def test_doc_title(self):
-        doc_regex = '\nDate:[^\n]+\n\n    doc'
+        if self.target_branch == 'main':
+            doc_regex = '\nDate:[^\n]+\n\n    doc'
+        else:
+            # the title of a non-cherry-pick commit starts with the target branch name, like
+            # "squid: doc: ...." instead of "doc: ..."
+            doc_regex = f'^\nDate:[^\n]+\n\n    ({self.target_branch}: )?doc'
```
tchaikov: this change is buggy, you added a '^' before '\\n'. this breaks all non-main branches. ideally, every commit should be self-contained and correct.

tchaikov (changes requested):
i withdraw my approval, because i realized that e6dd8df3ab16aa3ba8c51e999251845209c502b2 was buggy -- even the 2nd commit fixes it.

ceph-pr-commits/build/test_commits.py
bluikko: I apologize for that, copy-paste error. I'll transpose the commits too.

bluikko: I've followed your suggestions.  
It does make much more sense to first commit improvement to git output and then add the branch name support; this way the extra `^` would've likely not happened either. Very embarrassing, I tried to half-ass it by just adding another commit instead of rewriting the git history and of course made things worse.

Thanks again.

tchaikov (approved):
lgtm

PR #2584, opened 2026-05-15
update_shaman: Send pipeline matrix URL when updating shaman build url

djgalloway:
Fixes: https://tracker.ceph.com/issues/72370

### DIFF
--- a/scripts/update_shaman.sh
+++ b/scripts/update_shaman.sh
@@ -1,17 +1,65 @@
 #!/bin/bash
 # vim: ts=4 sw=4 expandtab
 
-submit_build_status() {
+get_pipeline_node_url() {
+    local distro="$1"
+    local arch="$2"
+    local flavor="${FLAVOR:-default}"
 
-    # A helper script to post (create) the status of a build in shaman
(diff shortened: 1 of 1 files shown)


scripts/update_shaman.sh
```diff
@@ -1,17 +1,65 @@
...
     distro=$4
     distro_version=$5
     distro_arch=$6
+
+    local pipeline_url log_url
+    pipeline_url=$(get_pipeline_node_url "$distro" "$distro_arch")
+
+    # If we were able to deduce an individual matrix branch node ID,
+    # use that for log_url, else use the older consoleFull endpoint.
+    if echo "$pipeline_url" | grep -q "selected-node"; then
```
phlogistonjohn: If you are OK with bashisms you could also do something like:
```
if [[ $pipeline_url =~ selected-node ]] ; then
```

It's a tad more efficient, and a tad easier to read (IMO), but not important.

phlogistonjohn (approved):

scripts/update_shaman.sh
djgalloway: Implemented :)

PR #2585, opened 2026-05-15
ceph-dev-pipeline: Validate given BRANCH param against derived BRANCH

djgalloway:
Handles an unlikely edge case but handled nonetheless.  This could be problematic if, for example, a user triggered ceph-dev-pipeline by specifying::

    SHA1=1234abcd
    BRANCH=main

But that SHA1 actually belongs to a wip branch.  Now both chacra and shaman think the tip of main is 1234abcd.

Fixes: https://tracker.ceph.com/issues/73658

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -177,7 +177,7 @@ def doSourceDistribution() {
 
 // Stage 4 (copy artifacts): copy artifacts from the setup job, verify SHA1, set VERSION/build description, and extract the source tarball.
 def doCopyArtifactsStage() {
-  def artifact_filter = "dist/sha1,dist/version,dist/other_envvars,dist/ceph-*.tar.bz2"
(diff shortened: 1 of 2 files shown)


djgalloway:
Works

```
++ git rev-parse HEAD
+ echo SHA1=5b8ba02b0047ca4d778cc5ae1717639d727e94a9
+ '[' '' = true ']'
+ '[' -n 5b8ba02b0047ca4d778cc5ae1717639d727e94a9 ']'
+ mapfile -t tip_branches
++ git for-each-ref --points-at HEAD '--format=%(refname:strip=3)' refs/remotes/origin
+ '[' 1 -eq 0 ']'
+ printf '%s
' main
+ grep -qxF not-a-real-branch
++ git rev-parse HEAD
+ echo 'BRANCH=not-a-real-branch does not point at SHA1=5b8ba02b0047ca4d778cc5ae1717639d727e94a9; using main'
BRANCH=not-a-real-branch does not point at SHA1=5b8ba02b0047ca4d778cc5ae1717639d727e94a9; using main
+ BRANCH=main
+ echo BRANCH=main
```

https://jenkins.ceph.com/job/ceph-source-dist/6126/console

PR #2570, opened 2026-04-23
feat: add Pulp upload scripts and pipeline integration

vamahaja:
This PR will introduce an optional path to publish built RPM/DEB artifacts to Pulp alongside the existing Chacra flow.
  - Introduce a `PULP_UPLOAD` boolean parameter (default false) on the `ceph-dev` job.
  - Add `scripts/setup_pulp.sh` to install `pulp-cli` (with pulp-cli-deb) via uv and write client config from Jenkins credentials.
  - Add `scripts/pulp_upload.sh` to upload discovered packages, attach them to the target repository, create a publication, and publish a distribution with project/version/ref/arch/sha labels.
  - Update the Jenkinsfile to install/configure the Pulp client when forcing a rebuild with Pulp upload while Chacra already has artifacts, and to invoke `pulp_upload.sh` after `chacra_upload.sh` when `THROWAWAY` is not `true` and `PULP_UPLOAD` is `true`.

PR #2591, opened 2026-05-25
fix(pulp): correct DEB uploads, distribution labels, and server URL

vamahaja:
Fix several issues in the Pulp upload pipeline introduced with the initial Pulp integration (#2570).

Distribution labels:
- Set `distro_version` from `OS_VERSION` (append ".stream" for CentOS 8+) instead of `OS_VERSION_NAME` so RPM distros label correctly.
- Add branch label and rename flavor --> flavors to match Pulp expectations.

DEB uploads:
- Resolve the upload directory via `pool/main`, with `WORKDIR` and `deb_tree` fallbacks when the APT pool layout is absent.
- Upload architecture-specific and "all" .deb packages; de-duplicate by base name when the same package appears in multiple paths.

RPM uploads:
- Name the `SRPM` Pulp repository `${REPO_NAME}-source` (not `-SRPMS`) to align with existing chacra source-repo conventions.

Setup:
- Point `PULP_SERVER_URL` at https://pulp.front.sepia.ceph.com.

Document:
- Add a script and method level header describing purpose.

### DIFF
--- a/scripts/pulp_upload.sh
+++ b/scripts/pulp_upload.sh
@@ -1,4 +1,25 @@
 #!/bin/bash
+
+# Upload Ceph build artifacts to a Pulp repository and publish distributions.
+
+# Intended for CI jobs after package builds complete. Discovers RPM or DEB
+# packages under WORKSPACE/dist/ceph, uploads them to Pulp, attaches the
+# uploaded content to per-arch repositories, and creates publications and
+# distributions with metadata labels for Shaman discovery.
+
+# Required environment variables:
+#   WORKSPACE       - Jenkins workspace root containing dist/ceph artifacts
+#   CEPH_VERSION    - Full Ceph version string (used to derive release branch)
+#   SHA1            - Git commit SHA for this build
+#   OS_NAME         - Target OS (e.g. centos, ubuntu)
+#   OS_VERSION      - Target OS version
+#   OS_VERSION_NAME - Ubuntu codename (used when OS_NAME=ubuntu)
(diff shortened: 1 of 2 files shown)


scripts/pulp_upload.sh
```diff
@@ -20,6 +20,26 @@ resolve_os_version_for_repo() {
...
+        printf '%s\n' "${pool_main}"
+        return 0
+    fi
+    local workdir="${deb_tree}/WORKDIR"
+    if [ -d "${workdir}" ]; then
+        log "pool/main not found; falling back to ${workdir}"
+        printf '%s\n' "${workdir}"
+        return 0
+    fi
+    log "WARNING: no pool/main or WORKDIR under ${deb_tree}; searching ${deb_tree}"
```
djgalloway, line 68: Should this be just a warning or be fatal?

djgalloway:
Good with this other than my one question

scripts/pulp_upload.sh
vamahaja: it should be just warning, in case of above two checks failing it will fall back to default value `deb_tree`

vamahaja:
Changes are working as expected, here is the testrun 

- https://jenkins.ceph.com/job/ceph-dev-pipeline/5766/

djgalloway:
Please add some comments in `pulp_upload.sh` to explain what the functions are for.  Input, output, etc.

vamahaja:
> Please add some comments in `pulp_upload.sh` to explain what the functions are for. Input, output, etc.

done

djgalloway (approved):

PR #2595, opened 2026-05-27
ceph-dev-pipeline: Fix NotSerializableException when iterating readPr…

djgalloway:
…operties result

Jenkins periodically saves its state to disk so it can resume builds after a restart or when waiting for an agent. Some Java objects cannot be saved this way. We were looping over a properties file in a way that created one of those unsaveable objects, which caused Jenkins to crash when it happened to try saving at that moment.

The fix changes the loop to only work with plain strings, which Jenkins has no trouble saving.


yuriw (approved):

dmick (approved):
I'd have added a comment, but it makes sense to me

PR #2599, opened 2026-05-29
ceph-pull-requests: add parameter for DISTRO_BASE

phlogistonjohn:
This new parameter can be used to manually start a "make check" job via with jenkins ui with a distro base other than the default value of "jammy". This will allow us to test on a newer distro base and eventually let (some) users self-service test their fixes for these issues.


djgalloway (approved):

dmick:
There aren't other dependencies, right, this can be merged?

phlogistonjohn:
> There aren't other dependencies, right, this can be merged?

Yes, IMO

djgalloway:
I only didn't merge it in case @dmick had something to say about the variable name itself 😄

dmick:
Yeah, I probably would have gone with something like BUILD_CONTAINER_BASE, but no worries

PR #2601, opened 2026-05-29
ceph-devstack: Update default repo

zmc:


djgalloway (approved):

ceph-devstack/config/definitions/ceph-devstack.yml
```diff
@@ -41,7 +41,7 @@
             - amathuria
```
djgalloway: Do we want to change this?

PR #2606, opened 2026-06-10
ceph-dev-pipeline, ceph-trigger-build: Update for u & v releases

zmc:


ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -381,7 +381,7 @@ pipeline {
                   podman pull ${env.CEPH_BUILDER_IMAGE}:${ceph_builder_tag_short} || \
                   true
                 """
-                def withCrimson = !(env.BRANCH in ['tentacle', 'squid', 'reef'])
+                def withCrimson = !(env.BRANCH in ['umbrella', 'tentacle', 'squid', 'reef'])
```
djgalloway, line 385: This will make umbrella _not_ build with crimson.  This was added to build with crimson on main by default but none of the release branches.

cbodley: cc @Matan-B: we've been disabling crimson for stable release builds - should this change for umbrella, or should we disable it here?

Matan-B: Crimson _must_ be supported in umbrella (for future 10.0 support), let's revert this.
@djgalloway @cbodley thanks for the alert!

Matan-B: cc @shraddhaag

ceph-trigger-build/build/Jenkinsfile
```diff
@@ -20,7 +20,7 @@ def pretty(obj) {
 // These are the default parameter values for the pipeline
 @Field def defaults = [
   'CEPH_BUILD_JOB': 'ceph-dev-pipeline',
-  'DISTROS': 'rocky10 centos9 jammy noble windows',
+  'DISTROS': 'rocky10 centos9 noble resolute windows',
```
cbodley, line 23: note that we haven't removed jammy from qa suites on main yet, so we'll still need jammy builds for now

i just raised https://github.com/ceph/ceph/pull/69422 to remove it from our qa suites, but it may take a while to merge

zmc: these are just the defaults for builds without a match below - given that, do you still want to keep jammy there?

ceph-dev-pipeline/build/Jenkinsfile
zmc: I'm happy to update this to whatever folks think is appropriate!

ceph-trigger-build/build/Jenkinsfile
cbodley: is there some other logic that adds jammy back to main and tentacle and umbrella? the only other mentions of jammy i see in this file are for reef and squid

zmc: not at the moment, so dropping this line

zmc:
It looks like we don't want the umbrella-related changes, so closing in favor of #2608 as I don't want the branch name on this PR to be misleading

PR #2607, opened 2026-06-11
ceph-dev-cron: add jammy to umbrella

cbodley:
umbrella will still support ubuntu jammy, and qa suites there depend on jammy packages. reenable jammy in ceph-dev-cron builds in case we want to schedule suites directly against the tip of umbrella

https://docs.ceph.com/en/latest/start/os-recommendations/#platforms

### DIFF
--- a/ceph-dev-cron/build/Jenkinsfile
+++ b/ceph-dev-cron/build/Jenkinsfile
@@ -22,7 +22,7 @@ node('built-in') {
           ]
       ],
       umbrella: [
-          distros: 'noble rocky10 centos9 windows',
+          distros: 'noble jammy rocky10 centos9 windows',
           extras : [
               [distros:'centos9 rocky10', flavors:'debug', archs:'x86_64']
           ]


cbodley:
when comparing our tentacle builds with umbrella builds, i also noticed that umbrella doesn't have any coverage of `archs:arm64` or `distros:bookworm` or `flavors:debug`

https://shaman.ceph.com/builds/ceph/umbrella/64c25741846ec007cd4a26dca9792a756beed596/
https://shaman.ceph.com/builds/ceph/tentacle/024d0210bc26738cabe6327628f7437ae89d304a/

i only added jammy in this pr because i know we need it for teuthology runs - we don't test arm64 or bookworm builds there

but it's possible that the crimson suite still relies on debug builds? should i add that here too? cc @Matan-B

cbodley:
p.s. maybe i'm looking in the wrong place, but i'm not sure what's actually triggering arm64 and bookworm builds for tentacle given this config:
```yaml
      tentacle: [
          distros: 'noble jammy rocky10 centos9 windows',
          extras : [
              [distros:'centos9 rocky10', flavors:'debug', archs:'x86_64']
          ]
```

Matan-B:
> when comparing our tentacle builds with umbrella builds, i also noticed that umbrella doesn't have any coverage of `archs:arm64` or `distros:bookworm` or `flavors:debug`
> 
> https://shaman.ceph.com/builds/ceph/umbrella/64c25741846ec007cd4a26dca9792a756beed596/ https://shaman.ceph.com/builds/ceph/tentacle/024d0210bc26738cabe6327628f7437ae89d304a/
> 
> i only added jammy in this pr because i know we need it for teuthology runs - we don't test arm64 or bookworm builds there
> 
> but it's possible that the crimson suite still relies on debug builds? should i add that here too? cc @Matan-B

Crimson suite will begin supporting deb with Noble (also in U). So jammy could stay as is without debug. fwiw, I think debug builds could be useful not only for Crimson but this seems out-of-scope.
Thank you for making sure!

CC @shraddhaag

djgalloway:
> p.s. maybe i'm looking in the wrong place, but i'm not sure what's actually triggering arm64 and bookworm builds for tentacle given this config:
> 
> ```yaml
>       tentacle: [
>           distros: 'noble jammy rocky10 centos9 windows',
>           extras : [
>               [distros:'centos9 rocky10', flavors:'debug', archs:'x86_64']
>           ]
> ```

Probably actual Ceph releases

djgalloway (approved):

PR #2608, opened 2026-06-11
ceph-dev-pipeline, ceph-trigger-build: Update for vampire

zmc:


djgalloway:
@batrick do we want debug builds too or nah?

ceph-trigger-build/build/Jenkinsfile
```diff
@@ -65,6 +65,10 @@ def params_from_branch(initialParams) {
         params[0]['FLAVOR'] += ' debug'
       }
       break
+    case ~/.*vampire.*/:
+      params[-1]['DISTROS'] = 'rocky10 noble resolute'
+      params[-1]['ARCHS'] += ' arm64'
```
batrick, line 70: Looks right to me.

Might want to cite: https://docs.ceph.com/en/latest/start/os-recommendations/#platforms

in your commit message.

batrick (approved):

batrick:
> @batrick do we want debug builds too or nah?

I would think that should be opt-in with a git trailer?

ceph-trigger-build/build/Jenkinsfile
batrick: Shouldn't `main` be updated too? But looking at the code I dont' see how.

I had some tickets created for 26.04 here:

https://tracker.ceph.com/issues/77310

if we can keep redmine updated with the various parts that would be nice for helping with the next iteration of this exercise.

zmc: I'm fine with adding main but it sounds like @cbodley also wants us to keep building jammy for main. Are we ok with the increased build volume?

batrick: I think maybe that's because of `make check` still running on jammy?

djgalloway: @cbodley ?

cbodley: i think we should keep building jammy packages for main until jammy is removed from our qa suites - i had raised https://github.com/ceph/ceph/pull/69422 to do that

cbodley: similarly for centos 9 in https://github.com/ceph/ceph-build/pull/2619 and https://github.com/ceph/ceph/pull/69406

djgalloway:
Closing in favor of #2726, which cherry-picks the ceph-trigger-build commit from here as-is and adds the rest of the vampire wiring (build_utils, ceph-dev-cron, PR pipeline whitelists, signing/sync scripts, quay-pruner). The ceph-dev-pipeline resolute commit already landed via 59b542e5. Thanks @zmc!

PR #2610, opened 2026-06-15
builder-reimage: improve reimage flow and node handling

jitendrasahu1803:
## Summary

This change improves the builder reimage pipeline by tightening validation, fixing failure handling, and adding safer Jenkins and MAAS node management around the reimage workflow.

## Changes included

- fail early when `installed-os-*` label is missing in inventory
- stop the pipeline correctly if MAAS deploy fails
- prevent post-reimage Ansible stages from running after reimage failure
- abort running MAAS deployment when the Jenkins job is aborted
- sync Jenkins node labels with inventory
- validate Jenkins node state before reimage
- mark node temporarily offline before reimage when needed
- restore node back online after completion only if this job marked it offline

## Outcome

- better validation before reimage starts
- safer handling of MAAS deployment failures
- cleaner Jenkins node lifecycle during reimage
- improved consistency between inventory and Jenkins node state

### DIFF
--- a/builder-reimage/config/definitions/builder-reimage.yml
+++ b/builder-reimage/config/definitions/builder-reimage.yml
@@ -33,7 +33,7 @@
       - git:
           url: https://github.com/ceph/ceph-build.git
           branches:
-            - main  
+            - main
           clean: true
 
     pipeline-scm:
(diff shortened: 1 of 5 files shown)


builder-reimage/config/definitions/builder-reimage.yml
```diff
@@ -1,5 +1,5 @@
 - job:
-    name: builder-reimage
+    name: preserve-builder-reimage
```
djgalloway: Leftover

djgalloway:
Where do we delete `/tmp/jenkins_git_key`?  We need to make sure we do that unconditionally.

builder-reimage/config/definitions/builder-reimage.yml
jitendrasahu1803: Fixed

djgalloway (approved):

PR #2611, opened 2026-06-15
setup_container_runtime.sh: Cleanup buildah containers

djgalloway:
Unsure where these are even coming from but let's clean 'em up.

Fixes: https://tracker.ceph.com/issues/77414


phlogistonjohn (approved):
lgtm

PR #2619, opened 2026-06-18
Drop centos9 from main builds

djgalloway:
Fixes: https://tracker.ceph.com/issues/77316


cbodley:
our qa suites on main still depend on these packages/containers, so we'll need to remove those first - similar to my https://github.com/ceph/ceph/pull/69422 which removes ubuntu 22

edit: i see that @batrick already prepared this in https://github.com/ceph/ceph/pull/69406 :)

PR #2612, opened 2026-06-16
introduce debug builds for noble

shraddhaag:
This commit adds debug builds for the distro noble. With this change, we will also get debug noble builds. This is especially useful for testing crimson in QA suites.

PR #2621, opened 2026-06-22
ceph-dev-pipeline: extract matrix stage script bodies into helper funcs

shraddhaag:
## Description
Problem: We were hitting JVM's 64KB bytecode limit when adding new distros to the matrix stage script. Example: https://github.com/ceph/ceph-build/pull/2612#issuecomment-4737982837. 

Solution: Add helper functions for each stage in the matrix stage script, such that the original block is significantly reduced, thus ensuring we do not hit the JVM's 64KB limit again. This will also ensure we can keep extending the matrix with more distros without worrying about this issue.

## Testing 
I have been able to successfully parse the XML using: `jenkins-jobs --conf ~/.jenkins_jobs.ini test ceph-dev-pipeline/config/definitions/ceph-dev-pipeline.yml`. I am unable to actually test with this as I do not have permissions to update the pipelines: `shraddhaag is missing the Job/Configure permission`. If someone can help test this out, that would be great!

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -141,33 +141,384 @@ def get_ceph_release_repo_text(base_url) {
   return text.toString()
 }
 
+// Stage 1 (source distribution): trigger the ceph-source-dist job and record its number/URL OR reuse SETUP_BUILD_ID.
+def doSourceDistribution() {
+  if ( ! env.SETUP_BUILD_ID ) {
+    def setup_build = build(
+      job: env.SETUP_JOB,
+      parameters: [
+        string(name: "BRANCH", value: env.BRANCH),
+        // Below are just for ceph-source-dist
+        string(name: "SHA1", value: env.SHA1),
+        string(name: "CEPH_REPO", value: env.CEPH_REPO),
+        string(name: "CEPH_BUILD_BRANCH", value: env.CEPH_BUILD_BRANCH),
+        // Below are only for actual releases
+        string(name: 'RELEASE_TYPE',  value: env.RELEASE_TYPE ?: ''),
+        string(name: 'RELEASE_BUILD', value: env.RELEASE_BUILD ?: ''),
+        string(name: 'VERSION',       value: env.VERSION ?: '')
+      ]
+    )
+    env.SETUP_BUILD_ID = setup_build.getNumber()
+  }
+  println "SETUP_BUILD_ID=${env.SETUP_BUILD_ID}"
+  env.SETUP_BUILD_URL = new URI([env.JENKINS_URL, "job", env.SETUP_JOB, env.SETUP_BUILD_ID].join("/")).normalize()
+  println "${env.SETUP_BUILD_URL}"
+}
+
+
+// Stage 2 (node): seed build_matrix for this cell, log the assigned node, and install the container runtime. Kept inline (no helper) so it can read the matrix-axis bindings DIST/ARCH/FLAVOR directly.
+// Stage 3 (checkout ceph-build): clone the ceph-build repo into the workspace. Kept inline because it's a single declarative `checkout` step with no script-block logic.
+
+// Stage 4 (copy artifacts): copy artifacts from the setup job, verify SHA1, set VERSION/build description, and extract the source tarball.
+def doCopyArtifactsStage() {
+  def artifact_filter = "dist/sha1,dist/version,dist/other_envvars,dist/ceph-*.tar.bz2"
+  def os = get_os_info(env.DIST)
+  if ( env.CI_COMPILE && os.pkg_type == "deb" ) {
+    artifact_filter += ",dist/ceph_*.diff.gz,dist/ceph_*.dsc"
+  }
+  println artifact_filter
+  copyArtifacts(
+    projectName: env.SETUP_JOB,
+    selector: specific(buildNumber: env.SETUP_BUILD_ID),
+    filter: artifact_filter,
+  )
+  sh 'sudo journalctl --show-cursor -n 0 --no-pager | tail -n1 | cut -d\" \" -f3 > $WORKSPACE/cursor'
+  def sha1_trimmed = env.SHA1.trim().toLowerCase()
+  def sha1_props = readProperties file: "${env.WORKSPACE}/dist/sha1"
+  def sha1_from_artifact = sha1_props.SHA1.trim().toLowerCase()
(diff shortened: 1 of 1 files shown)


djgalloway:
I would like to see very obvious comments above each function.  Additionally, I think it'd be beneficial for each function to have a number preceding it.

Something like,

```
// ===============
// 10 - Source Distribution
// ===============
def 10-doSourceDistribution() {
}

// ===============
// 20 - Copy Artifacts
// ===============
def 20-doCopyArtifactsStage() {
}
```

etc.

djgalloway:
This worked as a concept: https://jenkins.ceph.com/job/preserve-ceph-dev-pipeline-dgalloway/37/

shraddhaag:
>I would like to see very obvious comments above each function. 

done! 

>Additionally, I think it'd be beneficial for each function to have a number preceding it.

I think function names starting with numbers is not allowed by most languages, including Groovy. I can perhaps do something like `stage1DoSourceDistribution` if the idea is to explicitly mention in the function name the order in which they are invoked. 

If not, could you please help me understand why we want numeric to start the function name?

djgalloway:
> > I would like to see very obvious comments above each function.
> 
> done!

Can you make them more obvious?  Include a stage number.

> > Additionally, I think it'd be beneficial for each function to have a number preceding it.
> 
> I think function names starting with numbers is not allowed by most languages, including Groovy. I can perhaps do something like `stage1DoSourceDistribution` if the idea is to explicitly mention in the function name the order in which they are invoked.
> 
> If not, could you please help me understand why we want numeric to start the function name?

Ah, right, fair enough.  Good point.  Skip it.  I think "StageN" is worse.  Loud comments should be enough.

djgalloway:
We have a number of very large changes incoming to ceph-dev-pipeline so I would ask that we hold off on merging this until those are sorted.  Then we can split the new stages back out.

zmc:
I very much like the idea behind this PR! One change I would like to see, for readability / maintainability / safety reasons is the new functions taking arguments instead of depending on environment variables to be set.

shraddhaag:
> We have a number of very large changes incoming to ceph-dev-pipeline so I would ask that we hold off on merging this until those are sorted. Then we can split the new stages back out.

@djgalloway I saw that https://github.com/ceph/ceph-build/pull/2623 was merged which made substantial changes to ceph-dev-pipeline. I've rebased against main and extracted the new added logic into its own contained helper function. 

If we could review + test this PR one last time that will be great. I am blocked on this to enable debug builds for noble (https://github.com/ceph/ceph-build/pull/2622). 

> I very much like the idea behind this PR! One change I would like to see, for readability / maintainability / safety reasons is the new functions taking arguments instead of depending on environment variables to be set.

@zmc Thanks Zack! I definitely like the idea of explicit argument declaration. I went ahead and implemented it but turns out it is quite a major change touching code which seems out of scope of this PR. I've created a new PR with the respective changes: https://github.com/ceph/ceph-build/pull/2635 which we can test + merge once this one is in. If you'd prefer for both commit to land together in this PR itself, please let me know.

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -141,33 +141,380 @@ def get_ceph_release_repo_text(base_url) {
...
+        string(name: 'VERSION',       value: env.VERSION ?: '')
+      ]
+    )
+    env.SETUP_BUILD_ID = setup_build.getNumber()
+  }
+  println "SETUP_BUILD_ID=${env.SETUP_BUILD_ID}"
+  env.SETUP_BUILD_URL = new URI([env.JENKINS_URL, "job", env.SETUP_JOB, env.SETUP_BUILD_ID].join("/")).normalize()
+  println "${env.SETUP_BUILD_URL}"
+}
+
```
djgalloway, line 167: Can you add some comments here about what the stages are and that they are in-line in the pipeline below?

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -141,33 +141,380 @@ def get_ceph_release_repo_text(base_url) {
   return text.toString()
 }
 
+// Stage 1 (source distribution): trigger the source-dist setup job (or reuse SETUP_BUILD_ID) and record its number/URL.
```
djgalloway: How about

```
// Stage 1 (source distribution): trigger the ceph-source-dist job and record its number/URL OR reuse SETUP_BUILD_ID.
```

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -141,33 +141,380 @@ def get_ceph_release_repo_text(base_url) {
...
+  sh "cat dist/sha1 dist/version"
+  sh '''#!/bin/bash
+    set -ex
+    cd dist
+    mkdir ceph
+    tar --strip-components=1 -C ceph -xjf ceph-${VERSION}.tar.bz2 ceph-${VERSION}/{container,ceph.spec,ceph.spec.in,debian,Dockerfile.build,do_cmake.sh,install-deps.sh,run-make-check.sh,make-debs.sh,make-dist,make-srpm.sh,src/script}
+  '''
+}
+
+// Stage 5 (check for built packages): check chacra for an existing build (and skip if found, unless FORCE) and set up the Pulp client if needed.
```
djgalloway: `s/skip/skip compilation`

ceph-dev-pipeline/build/Jenkinsfile
shraddhaag: I had added the details for this inline where the logic exists. Also added the comments here.

ceph-dev-pipeline/build/Jenkinsfile
shraddhaag: done!

ceph-dev-pipeline/build/Jenkinsfile
shraddhaag: done!

dmick: this, by the way, still has the bug that it needs to be checking shaman and not chacra, since the chacra that it searches can't be known without asking shaman (and the local chacra for this build is not the right one except by coincidence).

I have a PR but I need to get back to it; just so you know, if testing doesn't go the way you expect.

dmick (approved):
I too wish for there to be more argument passing, and accept the argument that it's involved and could/should wait for a followup PR.  Otherwise, if it tests OK, it's ok, and hopefully will last for a while. 

I did look deeply enough to see that some variables had to change (adding env., adding def for local scope) and just hope that all those were caught with testing, because inspection is just crazymaking.

PR #2624, opened 2026-06-22
ceph-iscsi: Support building for el10

djgalloway:


dmick (approved):
Looks good.  I probably would have named CONTAINER_IMAGE something like BUILD_IN_CONTAINER, but with luck no one else will ever look at this again anyway

PR #2626, opened 2026-06-23
ceph-perf: pin pyjwt for github-check and detect clang-18

knrt10:
github-check fails on fresh venvs with PyJWT 2.13+ because github3 passes app_id as int to jwt.encode(), which now requires iss to be a string. Pin pyjwt<2.13 before installing githubcheck.

Also extend the clang version search down to 18 for Noble perf nodes, and quote app-id/install-id in github-check shell invocations.

### DIFF
--- a/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
+++ b/ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
@@ -58,31 +58,12 @@
           fi
           export NPROC=$(nproc)
           export FOR_MAKE_CHECK=true
+          export WITH_CRIMSON=true
           # Same Crimson-inclusive build as ceph-pull-requests (WITH_CRIMSON=true).
           # Classic vs crimson perf is selected at CBT runtime via run-cbt.sh.
-          # Crimson/Seastar needs a recent compiler: prefer clang-19+, and
-          # only fall back to GCC/G++ when the host has version 13+.
-          cxx_compiler=
-          c_compiler=
-          for i in $(seq 25 -1 19); do
-              if type -t clang-$i > /dev/null; then
-                  cxx_compiler="clang++-$i"
-                  c_compiler="clang-$i"
-                  break
-              fi
-          done
-          if test -z "$cxx_compiler"; then
-              gcc_major=$(gcc -dumpfullversion -dumpversion | cut -d. -f1)
-              gxx_major=$(g++ -dumpfullversion -dumpversion | cut -d. -f1)
-              if test "$gcc_major" -lt 13 || test "$gxx_major" -lt 13; then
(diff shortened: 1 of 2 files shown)


ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -64,7 +64,7 @@
           # only fall back to GCC/G++ when the host has version 13+.
           cxx_compiler=
           c_compiler=
-          for i in $(seq 25 -1 19); do
+          for i in $(seq 25 -1 18); do
```
Matan-B: iirc minimum supported clang version is 19 (coroutine support). Take a look at how the make check nodes are installing clang 19 (from source).

ceph-perf-pull-requests/config/definitions/ceph-perf-pull-requests.yml
```diff
@@ -58,31 +58,12 @@
...
-              if test "$gcc_major" -lt 13 || test "$gxx_major" -lt 13; then
-                  echo "Need clang-19+ or GCC/G++ 13+ for Crimson build (found gcc=$gcc_major g++=$gxx_major)" >&2
-                  exit 1
-              fi
-              cxx_compiler=g++
-              c_compiler=gcc
-          fi
-          export WITH_CRIMSON=true
-          cmake_args="-DCMAKE_CXX_COMPILER=$cxx_compiler -DCMAKE_C_COMPILER=$c_compiler -DCMAKE_BUILD_TYPE=Release -DWITH_CRIMSON=ON -DWITH_TESTS=OFF -DWITH_SPDK=OFF -DWITH_DPDK=OFF -DWITH_QATLIB=OFF -DWITH_QATZIP=OFF $CEPH_PERF_EXTRA_CMAKE_ARGS"
+          # Compiler: run-make.sh installs clang-19 on Jenkins (src/script/run-make.sh
```
Matan-B, line 64: Where is run-make.sh being run in these nodes?

knrt10: run-make.sh runs in the `run-cbt` builder [(line 67)](https://github.com/ceph/ceph-build/pull/2626/changes#diff-cd71ad4c32a6514db532136683b665cf0e23255b4021629c05b600470092504eR67), Twice per job. from `ceph-main/` and `ceph-pr/` on the performance-labeled agents. The setup step above it only installs CBT/github-check deps, it doesn't touch the compiler.

These aren't the same physical nodes as make-check. What's shared is the script. On first build, `run-make.sh prepare()` then `get_llvm()` installs `clang-19` via `llvm.sh` when `JENKINS_HOME` is set.

Matan-B (approved):

PR #2627, opened 2026-06-23
tcmu-runner: Support building for el10

djgalloway:


phlogistonjohn (approved):
I'm not an expert on tcmu runner but the container stuff and shell scripting largely looks good to me.

PR #2628, opened 2026-06-23
enhancement: split `Chacra` and `Pulp` upload into independent pipeline stages

vamahaja:
Refactor the ceph-dev-pipeline build job so Chacra and Pulp uploads are independently configurable and run as separate Jenkins stages.

Changes consists of:
- Add `CHACRA_UPLOAD` parameter (default: `true`) and enable `PULP_UPLOAD` by default, allowing each artifact destination to be toggled separately.
- Split `check for built packages` into `configure Chacra`, `configure Pulp`, and `validate artifact checks stages`, with Pulp repository existence checks mirroring the existing Chacra chacractl check.
- Split `upload packages` into `prepare upload`, `upload to Chacra`, and `upload to Pulp stages`; Shaman status updates are scoped to the Chacra stage.
- Extract `build_ceph_release_rpm()` helper so ceph-release RPM generation is shared for both Chacra and Pulp repo URLs.
- Require `PULP_SERVER_URL`, `PULP_USERNAME`, and `PULP_PASSWORD` in `setup_pulp.sh`; move the Pulp server URL into the Jenkinsfile.

### DIFF
--- a/ceph-dev-pipeline/config/definitions/ceph-dev-pipeline.yml
+++ b/ceph-dev-pipeline/config/definitions/ceph-dev-pipeline.yml
@@ -72,7 +72,7 @@
       - bool:
           name: PULP_UPLOAD
           description: "Whether to upload packages to Pulp"
-          default: false
+          default: true
 
       - bool:
           name: PULP_REGISTRY_UPLOAD


scripts/pulp_upload.sh
djgalloway, line 1: If we're trying to create a Pulp repository that already exists (the name already exists), we should WARN that the repo already exists, then:

1. check if it's the same sha1 (so I guess in pulp land, this is called distribution?)
  a. if it is the same sha1,
    i. if `FORCE=true`, delete the distribution and recreate it
    ii. if `FORCE=false`, check for each individual package and only upload missing packages
  b. if different sha1, proceed with uploading packages like normal
2. if different sha1, proceed as normal

This PR currently indiscriminately wipes out any existing distribution that, while unlikely, may be in use.

vamahaja: @djgalloway  Thanks for the feedback! Breaking down how the current implementation handles these scenarios:

Regarding the check for the same sha1 (distribution):

1. If it is the same sha1:
    - `FORCE=true` (delete and recreate): This is already working exactly as you suggested.
    - `FORCE=false` (check individual packages and upload missing): Currently, when `FORCE=false`, the pipeline skips building packages altogether if it finds the distribution is already present for that specific `sha1`. Modifying this to check for individual missing packages would be a very tedious task, as it requires multiple steps to fetch and validate packages specifically mapped through `distribution` -> `publication` -> `content`. Furthermore, Pulp already works smartly in the background to avoid redundant uploads by maintaining references to existing packages rather than duplicating content.
2. If it's a different sha1:
    - Proceed with uploading packages like normal: This is already working exactly the same way.

ceph-dev-pipeline/config/definitions/ceph-dev-pipeline.yml
```diff
@@ -72,7 +72,7 @@
       - bool:
           name: PULP_UPLOAD
           description: "Whether to upload packages to Pulp"
-          default: false
+          default: true
```
djgalloway, line 75: Did you mean to flip this?

vamahaja: Yes, setting default to `true`. This will upload all packages to pulp.

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -233,32 +234,101 @@ def doCopyArtifactsStage() {
...
+  check_pulp_repo += "project=ceph,"
+  check_pulp_repo += "branch=${env.BRANCH},"
+  check_pulp_repo += "arch=${env.ARCH},"
+  check_pulp_repo += "flavor=${env.FLAVOR},"
+  check_pulp_repo += "sha1=${env.SHA1},"
+  check_pulp_repo += "distro=${os.name},"
+  check_pulp_repo += "distro_version=${os.version_name}"
+  def pulp_repo_output = sh(script: check_pulp_repo, returnStdout: true).trim()
+
+  env.PULP_REPO_EXISTS = pulp_repo_output ? "true" : "false"
```
djgalloway: This check will always return true because it's testing whether the output is empty: pulp ... repository list prints [] when nothing matches, which is a non-empty string, so PULP_REPO_EXISTS is always "true". With PULP_UPLOAD now defaulting to true and FORCE=false, every matrix cell would skip compilation. Verified against pulp.front.sepia.ceph.com with labels that can't match anything:

```
pulp rpm repository list --label-select="project=ceph,branch=main,arch=x86_64,flavor=default,sha1=abcd,distro=centos,distro_version=9"
[]
```

The label query also can't ever match what pulp_upload.sh actually stores: the label key is flavors (query uses flavor=), and distro_version is stored as e.g. 9.stream/10.1 (query uses 9/10 from get_os_info). And repositories are branch-scoped and reused across builds — the per-sha1 labels live on distributions.

Since distributions are per-sha1 and deterministically named (dist-ceph-<branch>-<os>-<version>-<arch>-<short_sha1>), I'd suggest checking for the distribution by name and using the exit code instead of parsing output. Note the naming details when mirroring pulp_upload.sh: ubuntu uses the codename (noble) rather than 24.04, rpm arches use aarch64 (the matrix axis is arm64), and deb lookups use --name while rpm uses --distribution.

```
pulp rpm distribution show --distribution "dist-ceph-tentacle-rocky-10-x86_64-9cbd0048"
echo "rc=$?"
{
  ...
  "pulp_labels": {
    "flavors": "default",
    "distro_version": "10.1",
    ...
  },
  "name": "dist-ceph-tentacle-rocky-10-x86_64-9cbd0048",
  ...
}
rc=0

pulp rpm distribution show --distribution "dist-ceph-tentacle-rocky-10-x86_64-deadbeef"
echo "rc=$?"
Error: Could not find rpm distribution with {'name': 'dist-ceph-tentacle-rocky-10-x86_64-deadbeef'}.
rc=1
```

If the naming ever drifts from pulp_upload.sh, this fails toward rebuilding instead of silently skipping.

scripts/pulp_upload.sh
zmc: This response reads like it was generated by an LLM and if that's the case, let's please not do that to each other. It also misses what seems like the key point to me, which is that it deletes existing distros. I think that deserves some thought.

vamahaja:
Changes are already merged as part of PR - https://github.com/ceph/ceph-build/pull/2645.

PR #2636, opened 2026-06-29
Fixed issues in ppc-ut job.

anshuman-agarwala:
Fixed PullID description and usage.
Removed cobertura publisher since we're not running dashboard suites on ppc64.


hasan4791:
@anshuman-agarwala Also we need to rename ppc64 occurrences with ppc64le as they refer to different endian architectures. We need to change the label in the workers also.

zmc (approved):

PR #2639, opened 2026-07-02
ceph-dev-pipeline: report builder-container failures to shaman

tchaikov:
The "builder container" stage builds the ceph-build image used to compile and package ceph, but never calls update_shaman.sh: no started, no failed. Only the later "build" and "upload packages" stages report status.

When the image build itself fails, say a newly added build dependency turns out not to be installable on one of the distros in the build matrix, Jenkins marks the branch red, but shaman hears nothing about it and keeps showing the last known good build. A distro that's actually broken looks the same as one that's passing.

Add the same started/failed calls the later stages already use, so builder-container failures reach shaman too.


djgalloway:
I like this idea.  Can you just rebase please?

tchaikov:
@djgalloway hi David, thanks! sure. rebased.

djgalloway (approved):

PR #2644, opened 2026-07-08
pulp-cleanup: add nightly Jenkins job for stale Pulp resource cleanup

vamahaja:
## Summary
- Add a nightly `pulp-cleanup-nightly` Jenkins job that purges stale `ceph` project Pulp repositories and distributions using `scripts/purge-policy.yaml`.
- Cleanup excludes named policy refs via `label-select`, deletes remaining resources past the default retention days (distributions before repositories), then runs orphan cleanup.
- Update `pulp_upload.sh` so repository version retention uses per `BRANCH` & `keep_minimum` from the purge policy, with fallback to the default.

## Job config
- Job: `pulp-cleanup-nightly`, triggered `@midnight`.
- Parameters: `CEPH_BUILD_BRANCH` (default `main`), `CEPH` bool (default `true`) to enable `ceph` project cleanup.
- Pipeline sets up the Pulp client, then runs `cleanup.sh` when `CEPH` is `true`.

## Purge policy (`scripts/purge-policy.yaml`)
- Project root: `ceph`
- Default retention: `days: 14`, `keep_minimum: 3`
- Per-ref overrides for release and special branches
- Cleanup uses named refs as exclusions; uploads use `keep_minimum` for repository version retention

### DIFF
--- a/pulp-cleanup/build/Jenkinsfile
+++ b/pulp-cleanup/build/Jenkinsfile
@@ -1,0 +1,68 @@
+import groovy.transform.Field
+
+@Field String pulpServerUrl = "https://pulp.front.sepia.ceph.com"
+
+
+pipeline {
+    agent {
+        label 'small'
+    }
+    
+    options {
+        timeout(time: 1, unit: 'HOURS')
+        timestamps()
+    }
+    
+    stages {
+        stage('Checkout') {
+            steps {
+                checkout scm
+            }
+        }
+
+        stage('Setup Pulp Client') {
+            steps {
+                script {
+                    withCredentials([
+                        usernamePassword(
+                            credentialsId: 'pulp-admin-auth',
+                            usernameVariable: 'PULP_USERNAME',
+                            passwordVariable: 'PULP_PASSWORD'
+                        )
+                    ]) {
+                        echo "Setting up Pulp Client..."
+                        sh """#!/bin/bash -ex
+                            export PULP_SERVER_URL="${pulpServerUrl}" && ./scripts/setup_pulp.sh
+                        """
+                    }
+                }
+            }
+        }
+
+        stage('Setup yq') {
+            steps {
+                script {
+                    echo "Setting up yq..."
+                    sh """#!/bin/bash -ex
+                        source ./scripts/setup_yq.sh
+                    """
+                }
+            }
+        }
+
+        stage('Execute Pulp Cleanup for Ceph') {
+            when {
+                expression {
+                    params.CEPH
+                }
+            }
+            steps {
+                script {
+                    def project = "ceph"
+                    echo "Starting Pulp Cleanup for project: ${project}"
+                    sh "./pulp-cleanup/build/cleanup.sh ${project}"
+                }
+            }
+        }
+    }
+}
--- a/pulp-cleanup/build/Jenkinsfile
+++ b/pulp-cleanup/build/Jenkinsfile
@@ -1,0 +1,62 @@
+import groovy.transform.Field
+
+@Field String pulpServerUrl = "https://pulp.front.sepia.ceph.com"
+
+
+pipeline {
+    agent {
+        label 'small'
+    }
+    
+    options {
+        timeout(time: 1, unit: 'HOURS')
+        timestamps()
+    }
+    
+    stages {
+        stage('Checkout') {
+            steps {
+                checkout scm
+            }
+        }
+
+        stage('Setup Pulp Client') {
+            steps {
+                script {
+                    withCredentials([
+                        usernamePassword(
+                            credentialsId: 'pulp-admin-auth',
+                            usernameVariable: 'PULP_USERNAME',
+                            passwordVariable: 'PULP_PASSWORD'
+                        )
+                    ]) {
+                        echo "Setting up Pulp Client..."
+                        sh """#!/bin/bash -ex
+                            export PULP_SERVER_URL="${pulpServerUrl}" && ./scripts/setup_pulp.sh
+                        """
+                    }
+                }
+            }
+        }
+
+        stage('Execute Pulp Cleanup for Ceph') {
+            when {
+                expression {
+                    params.CEPH
+                }
+            }
+            steps {
+                script {
+                    def project = "ceph"
+                    def dryRun = params.DRY_RUN ?: false
+                    echo "Starting Pulp Cleanup for project: ${project}"
+                    echo "DRY_RUN: ${dryRun}"
+                    sh """#!/bin/bash -ex
+                        export DRY_RUN=${dryRun}
+                        ./pulp-cleanup/build/cleanup.sh ${project}
+                    """
+                }
+            }
+        }
+    }
+}
--- a/pulp-cleanup/build/cleanup.sh
+++ b/pulp-cleanup/build/cleanup.sh
@@ -1,0 +1,190 @@
+#!/bin/bash
+
+# Purge stale Pulp repositories and distributions per purge-policy.yaml.
+
+set -euo pipefail
+
+readonly PROJECT="$1"
+readonly PURGE_POLICY_FILE="${WORKSPACE}/scripts/purge-policy.yaml"
+
+readonly LABEL="ref"
+readonly PULP_LIST_LIMIT=1000
+readonly PULP_TYPES=(rpm deb)
+readonly PULP_RESOURCES=(distribution repository)
+
+# Default protection time in hours
+readonly PROTECTION_TIME=24
+
+log() {
+    echo "[pulp_cleanup] $*" >&2
+}
+
+get_policy_refs() {
+    local project="$1"
+    local label="$2"
+
+    yq -r ".${project}.${label} | keys[]" "${PURGE_POLICY_FILE}"
+}
+
+read_default_purge_policy() {
+    local project="$1"
+    local -n _days="$2"
+    local policy_line
+
+    policy_line=$(
+        yq -r \
+            ".${project}.default | [(.days // \"\")] | @tsv" \
+            "${PURGE_POLICY_FILE}"
+    )
+    IFS=$'\t' read -r _days <<< "${policy_line}"
+}
+
+fetch_pulp_resource_json() {
+    local type="$1"
+    local resource="$2"
+    local project="$3"
+    local label_select="$4"
+
+    pulp "${type}" "${resource}" list \
+        --limit "${PULP_LIST_LIMIT}" \
+        --ordering '-pulp_created' \
+        --label-select project="${project}","${label_select}" \
+        2>&1 | sed -n '/^\[/,$p'
+}
+
+filter_resource_names_by_age() {
+    local pulp_json="$1"
+    local days="$2"
+
+    if [ -z "${days}" ]; then
+        echo "${pulp_json}" | jq -r '.[].name'
+        return 0
+    fi
+
+    echo "${pulp_json}" | jq -r --arg days "${days}" '
+        .[]
+        | select(
+            (.pulp_created | split(".")[0] + "Z" | fromdateiso8601)
+            < (now - ($days | tonumber) * 86400)
+          )
+        | .name
+    '
+}
+
+list_stale_pulp_resources() {
+    local type="$1"
+    local resource="$2"
+    local project="$3"
+    local label_select="$4"
+    local days="$5"
+    local pulp_json resource_list
+
+    pulp_json=$(
+        fetch_pulp_resource_json \
+            "${type}" "${resource}" "${project}" "${label_select}"
+    )
+    resource_list=$(
+        filter_resource_names_by_age "${pulp_json}" "${days}"
+    )
+    echo "${resource_list}"
+}
+
+destroy_pulp_resources() {
+    local type="$1"
+    local resource="$2"
+    local resource_list="$3"
+    local name
+
+    if [ -z "${resource_list}" ]; then
+        return 0
+    fi
+
+    while IFS= read -r name; do
+        [ -z "${name}" ] && continue
+        log "Destroying ${type} ${resource}: ${name}"
+        if ! pulp "${type}" "${resource}" destroy \
+                --name "${name}"; then
+            log "ERROR: Failed to destroy ${type} ${resource}: ${name}"
+            return 1
+        fi
+    done <<< "${resource_list}"
+}
+
+collect_stale_resources_for_ref() {
+    local project="$1"
+    local label_select="$2"
+    local days="$3"
+    local type resource key
+
+    declare -n _stale_resources="$4"
+
+    for type in "${PULP_TYPES[@]}"; do
+        for resource in "${PULP_RESOURCES[@]}"; do
+            key="${type}_${resource}"
+            _stale_resources[$key]=$(
+                list_stale_pulp_resources \
+                    "${type}" "${resource}" "${project}" \
+                    "${label_select}" "${days}"
+            )
+            log "${key}: ${_stale_resources[$key]:-}"
+        done
+    done
+}
+
+purge_resources() {
+    local project="$1"
+    local label_select="$2"
+    local days="$3"
+    local type resource
+
+    declare -A stale_resources=()
+    log "Purging days=${days:-unset} label_select=${label_select}"
+    collect_stale_resources_for_ref "${project}" "${label_select}" "${days}" \
+        stale_resources
+
+    log "stale_resources: ${stale_resources[@]}"
+
+    for type in "${PULP_TYPES[@]}"; do
+        destroy_pulp_resources "${type}" "distribution" \
+            "${stale_resources[${type}_distribution]:-}"
+    done
+    for type in "${PULP_TYPES[@]}"; do
+        destroy_pulp_resources "${type}" "repository" \
+            "${stale_resources[${type}_repository]:-}"
+    done
+}
+
+run_package_cleanup() {
+    local project="$1"
+    local label="$2"
+    local days
+
+    mapfile -t policy_refs < <(get_policy_refs "${project}" "${label}")
+    _label_select=$(
+        printf "${label}!=%s\n" "${policy_refs[@]}" | paste -sd, -
+    )
+    log "label_select: ${_label_select}"
+
+    read_default_purge_policy "${project}" days
+    log "days: ${days}"
+
+    purge_resources "${project}" "${_label_select}" "${days}"
+}
+
+run_orphan_cleanup() {
+    log "Applying orphan content purge policy"
+    if ! pulp orphan cleanup \
+            --protection-time "${PROTECTION_TIME}"; then
+        log "ERROR: Failed to run orphan cleanup"
+        return 1
+    fi
+    log "Orphan cleanup completed"
+}
+
+log "Cleaning up project: ${PROJECT}"
+
+log "Running package cleanup"
+run_package_cleanup "${PROJECT}" "${LABEL}"
+
+log "Running orphan cleanup"
+run_orphan_cleanup
--- a/pulp-cleanup/build/cleanup.sh
+++ b/pulp-cleanup/build/cleanup.sh
@@ -1,0 +1,573 @@
+#!/bin/bash
+
+# Purge orphaned Pulp distributions/publications and stale unlisted repos
+# per purge-policy.yaml.
+
+set -euo pipefail
+
+export PATH="$HOME/.local/bin:$PATH"
+
+log() {
+    echo "[pulp_cleanup] $*" >&2
+}
+
+if [ $# -lt 1 ] || [ -z "${1}" ]; then
+    log "ERROR: project name required"
+    exit 1
+fi
+
+if [ -z "${WORKSPACE:-}" ] || [ ! -d "${WORKSPACE}" ]; then
+    log "ERROR: WORKSPACE is not set or not a directory"
+    exit 1
+fi
+
+readonly PROJECT="$1"
+readonly PURGE_POLICY_FILE="${WORKSPACE}/scripts/purge-policy.yaml"
+readonly PURGE_POLICY_PY="${WORKSPACE}/scripts/purge_policy.py"
+readonly DRY_RUN="${DRY_RUN:-false}"
+
+readonly LABEL="ref"
+readonly PULP_LIST_LIMIT=1000
+readonly PULP_TYPES=(rpm deb)
+
+# Default protection time in minutes
+readonly PROTECTION_TIME=1440
+
+DEFAULT_POLICY_JSON=""
+REF_POLICY_JSON=""
+PULP_DESTROY_FAILURES=0
+
+load_purge_policy_section() {
+    local project="$1"
+    local label="$2"
+    local output
+
+    if [ ! -f "${PURGE_POLICY_FILE}" ]; then
+        log "ERROR: Purge policy file not found: ${PURGE_POLICY_FILE}"
+        return 1
+    fi
+
+    if ! output=$(
+        python3 "${PURGE_POLICY_PY}" \
+            --project "${project}" --label "${label}" \
+            --file "${PURGE_POLICY_FILE}" 2>&1
+    ); then
+        log "ERROR: Failed to read ${label} policy for ${project}: ${output}"
+        return 1
+    fi
+
+    printf '%s\n' "${output}"
+}
+
+load_purge_policies() {
+    local project="$1"
+
+    if ! DEFAULT_POLICY_JSON=$(
+        load_purge_policy_section "${project}" default
+    ); then
+        return 1
+    fi
+    if ! REF_POLICY_JSON=$(load_purge_policy_section "${project}" ref); then
+        return 1
+    fi
+}
+
+read_default_purge_policy() {
+    local -n _days="$1"
+
+    _days=$(echo "${DEFAULT_POLICY_JSON}" | jq -r '.days')
+}
+
+fetch_pulp_resource_json() {
+    local type="$1"
+    local resource="$2"
+    local project="$3"
+    local label_select="$4"
+    local pulp_output
+
+    if ! pulp_output=$(
+        pulp "${type}" "${resource}" list \
+            --limit "${PULP_LIST_LIMIT}" \
+            --ordering 'pulp_last_updated' \
+            --label-select project="${project}","${label_select}" \
+            2>&1
+    ); then
+        log "ERROR: Failed to list ${type} ${resource}: ${pulp_output}"
+        return 1
+    fi
+
+    sed -n '/^\[/,$p' <<< "${pulp_output}"
+}
+
+fetch_pulp_list_json() {
+    local type="$1"
+    local resource="$2"
+    local pulp_output
+
+    if ! pulp_output=$(
+        pulp "${type}" "${resource}" list \
+            --limit "${PULP_LIST_LIMIT}" \
+            "${@:3}" \
+            2>&1
+    ); then
+        log "ERROR: Failed to list ${type} ${resource}: ${pulp_output}"
+        return 1
+    fi
+
+    sed -n '/^\[/,$p' <<< "${pulp_output}"
+}
+
+list_repositories() {
+    local type="$1"
+    local project="$2"
+    local label_select="$3"
+
+    fetch_pulp_resource_json "${type}" "repository" "${project}" \
+        "${label_select}"
+}
+
+build_unlisted_label_select() {
+    local label="$1"
+    local -a policy_refs=("${@:2}")
+
+    printf "${label}!=%s\n" "${policy_refs[@]}" | paste -sd, -
+}
+
+list_distributions_for_label_select() {
+    local type="$1"
+    local project="$2"
+    local label_select="$3"
+    local dists_json
+
+    if ! dists_json=$(
+        fetch_pulp_resource_json "${type}" "distribution" "${project}" \
+            "${label_select}"
+    ); then
+        printf '[]\n'
+        return 0
+    fi
+
+    if ! echo "${dists_json}" | jq -e 'type == "array"' >/dev/null 2>&1; then
+        log "WARNING: Invalid ${type} distribution list" \
+            "for label_select=${label_select}"
+        printf '[]\n'
+        return 0
+    fi
+
+    echo "${dists_json}" | jq -c 'sort_by(.pulp_last_updated)'
+}
+
+fetch_pulp_show_json() {
+    local type="$1"
+    local resource="$2"
+    local href="$3"
+    local pulp_output
+
+    if ! pulp_output=$(
+        pulp "${type}" "${resource}" show --href "${href}" 2>&1
+    ); then
+        log "WARNING: Failed to read ${type} ${resource}" \
+            "${href}: ${pulp_output}"
+        return 1
+    fi
+
+    sed -n '/^{/,$p' <<< "${pulp_output}"
+}
+
+repository_version_exists() {
+    local type="$1"
+    local version_href="$2"
+    local repository_href version_num
+
+    if [[ ! "${version_href}" =~ ^(.+)/versions/([0-9]+)/?$ ]]; then
+        log "WARNING: Unrecognized repository version href: ${version_href}"
+        return 1
+    fi
+
+    repository_href="${BASH_REMATCH[1]}/"
+    version_num="${BASH_REMATCH[2]}"
+
+    if pulp "${type}" repository version show \
+            --repository-href "${repository_href}" \
+            --version "${version_num}" >/dev/null 2>&1; then
+        return 0
+    fi
+
+    return 1
+}
+
+get_publication_orphan_status() {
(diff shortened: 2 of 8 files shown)


djgalloway:
I think we need something more like [chacra's retention map](https://github.com/ceph/chacra/blob/main/deploy/playbooks/roles/common/templates/prod.py.j2#L88-L162).

Or at least leave this PR as-is but add `KEEP_MINIMUM` or `RETAIN_MINIMUM` on top of `RETAIN_DAYS`.

vamahaja:
> I think we need something more like [chacra's retention map](https://github.com/ceph/chacra/blob/main/deploy/playbooks/roles/common/templates/prod.py.j2#L88-L162).
> 
> Or at least leave this PR as-is but add `KEEP_MINIMUM` or `RETAIN_MINIMUM` on top of `RETAIN_DAYS`.

Updated logic PTAL.

scripts/purge-policy.yaml
djgalloway, line 1: I think it makes more sense to have this in `pulp-cleanup/config`

scripts/purge-policy.yaml
```diff
@@ -0,0 +1,58 @@
+ceph:
+  default:
+    days: 14
+    keep_minimum: 3
+  ref:
```
djgalloway, line 5: Add `main` please.  Give it the same retention as a release branch

pulp-cleanup/build/cleanup.sh
djgalloway, line 1: `keep_minimum` is never read from anywhere in here

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,190 @@
...
+readonly PROJECT="$1"
+readonly PURGE_POLICY_FILE="${WORKSPACE}/scripts/purge-policy.yaml"
+
+readonly LABEL="ref"
+readonly PULP_LIST_LIMIT=1000
+readonly PULP_TYPES=(rpm deb)
+readonly PULP_RESOURCES=(distribution repository)
+
+# Default protection time in hours
+readonly PROTECTION_TIME=24
```
djgalloway, line 16: Minutes.  Do 1440.

```
Options:
  --content-hrefs TEXT       List of specific Contents to delete if they are
                             orphans
  --protection-time INTEGER  How long in minutes Pulp should hold orphan
                             Content and Artifacts before becoming candidates
                             for cleanup task
```

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,190 @@
...
+
+fetch_pulp_resource_json() {
+    local type="$1"
+    local resource="$2"
+    local project="$3"
+    local label_select="$4"
+
+    pulp "${type}" "${resource}" list \
+        --limit "${PULP_LIST_LIMIT}" \
+        --ordering '-pulp_created' \
```
djgalloway, line 50: `-pulp_created` fetches the newest N.  We want to delete the oldest N.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,190 @@
...
+        --label-select project="${project}","${label_select}" \
+        2>&1 | sed -n '/^\[/,$p'
```
djgalloway, line 52: The job/script will fail silently with this.  Make a failed `pulp` call fatal.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,190 @@
...
+    fi
+
+    echo "${pulp_json}" | jq -r --arg days "${days}" '
+        .[]
+        | select(
+            (.pulp_created | split(".")[0] + "Z" | fromdateiso8601)
+            < (now - ($days | tonumber) * 86400)
+          )
+        | .name
+    '
```
djgalloway, line 71: We should filter on `pulp_last_updated` instead.

Also, we'll have a `ceph-tentacle-rocky-10-x86_64` _repository_ forever.  We want to clean up all of the _distrubutions_ inside that are older than 14 days but keep a minimum of 10.

So if the _repository_ matches any of the named branches in `purge-policy.yaml`, we should proceed with cleaning up the _distributions_ inside by following the policy.

If the _repository_ doesn't match a named branch in `purge_policy.yaml` and its `pulp_last_updated` is over 14 days ago, delete the repository.  If newer than 14 days, we should check inside for distributions and clean up any older than 14 days.

pulp-cleanup/build/Jenkinsfile
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }
+
+        stage('Setup yq') {
+            steps {
+                script {
+                    echo "Setting up yq..."
+                    sh """#!/bin/bash -ex
+                        source ./scripts/setup_yq.sh
```
djgalloway, line 47: Sourcing the script just defines the function.  Just run it directly.

pulp-cleanup/config/definitions/pulp-cleanup.yaml
```diff
@@ -0,0 +1,37 @@
+- job:
+    name: pulp-cleanup-nightly
+    disabled: true
+    display-name: 'Pulp: Cleanup Nightly'
+    project-type: pipeline
+    description: >
+      Nightly parameterized cleanup of stale Teuthology Pulp resources.
```
djgalloway, line 7: Suggest `Nightly parameterized cleanup of stale ceph builds in Pulp`

pulp-cleanup/config/definitions/pulp-cleanup.yaml
```diff
@@ -0,0 +1,37 @@
+- job:
+    name: pulp-cleanup-nightly
+    disabled: true
```
djgalloway, line 3: Is this intentional?

pulp-cleanup/config/definitions/pulp-cleanup.yaml
djgalloway, line 1: The other job definitions are `.yml`

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,190 @@
...
+
+run_package_cleanup() {
+    local project="$1"
+    local label="$2"
+    local days
+
+    mapfile -t policy_refs < <(get_policy_refs "${project}" "${label}")
+    _label_select=$(
+        printf "${label}!=%s\n" "${policy_refs[@]}" | paste -sd, -
+    )
```
djgalloway, line 165: If anything goes wrong here, this falls back to "delete everything"

Gate on a yq failure, missing file, etc.  Any error.

djgalloway:
Maybe implement a `DRY_RUN`.  "Would delete repository ceph-wip-nbala-test-1-centos-9-default-x86_64"

"Leaving distributions X Y Z in repository ceph-tentacle-rocky-10-x86_64 to satisfy keep_minimum: 10"

Something like those.  I'd like to see some dry-run output before we let this loose even though the artifacts in pulp aren't terribly important yet.

Finished with my review for now.

pulp-cleanup/build/cleanup.sh
vamahaja: `keep_minimum` is used in `pulp_upload.sh` to set `REPO_VERSIONS_TO_RETAIN`

https://github.com/ceph/ceph-build/pull/2644/changes#diff-e43616de828404d15446d65b52a9351b6042eefe915702c04074e6b674e46d71R58

scripts/purge-policy.yaml
vamahaja: this is referred in `pulp_upload.sh` and `cleanup.sh`, hence this path.

scripts/purge-policy.yaml
vamahaja: done

pulp-cleanup/config/definitions/pulp-cleanup.yml
```diff
@@ -0,0 +1,37 @@
+- job:
+    name: pulp-cleanup-nightly
+    disabled: true
```
vamahaja: yes, right now job is disabled. Once all comments are addressed, enable the job.

pulp-cleanup/config/definitions/pulp-cleanup.yml
vamahaja: done

pulp-cleanup/config/definitions/pulp-cleanup.yaml
vamahaja: thanks, updated description.

pulp-cleanup/build/cleanup.sh
vamahaja: this is intentional

`list_stale_pulp_resources` method works in two steps, at first step `fetch_pulp_resource_json` fetches first `1000` resources in json format and then `filter_resource_names_by_age` filters resources by days set by purge policy.

pulp-cleanup/build/cleanup.sh
vamahaja: Agree on filter, updated `fetch_pulp_resource_json` to use `pulp_last_updated`.

All repositories are getting created with parameter `--retain-repo-versions`, which are based on purge policy. Repositories like `ceph-tentacle-rocky-10-x86_64` will delete older versions before creating new one. As we are triggering orphan cleanup as part of this job, it will take care of cleaning all stale references.

vamahaja: Please find my reply inline below

> We should filter on `pulp_last_updated` instead.
> --> **Agree on filter, updated fetch_pulp_resource_json to use pulp_last_updated.**

> Also, we'll have a `ceph-tentacle-rocky-10-x86_64` _repository_ forever. We want to clean up all of the _distrubutions_ inside that are older than 14 days but keep a minimum of 10.
> --> **All repositories are getting created with parameter `--retain-repo-versions`, which are based on purge policy. Repositories like ceph-tentacle-rocky-10-x86_64 will delete older versions before creating new one. As we are triggering orphan cleanup as part of this job, it will take care of cleaning all stale references.**

> So if the _repository_ matches any of the named branches in `purge-policy.yaml`, we should proceed with cleaning up the _distributions_ inside by following the policy.
> **--> These are all release branch repos and never get deleted, however the versions, distribution and publication are managed by pulp via parameter `--retain-repo-versions` set during repo creation.**

> If the _repository_ doesn't match a named branch in `purge_policy.yaml` and its `pulp_last_updated` is over 14 days ago, delete the repository. If newer than 14 days, we should check inside for distributions and clean up any older than 14 days.
> **--> In this case if repo is not updated within `default.days` get deleted and distribution and publication are managed by pulp via parameter `--retain-repo-versions` set during repo creation.**

pulp-cleanup/build/Jenkinsfile
vamahaja: fixed

pulp-cleanup/build/cleanup.sh
vamahaja: separated `sed` command and added check for pulp call

pulp-cleanup/build/cleanup.sh
vamahaja: Done !

Added multiple checks to validate policy parameters

pulp-cleanup/build/cleanup.sh
vamahaja: Updated `PROTECTION_TIME` to `1440`.

References can be anything from repo, distribution or publication, identifying resource type based on `hrefs` can be difficult. I though of having separate parameters for mentioned resources but with names, but this will add unnecessary parameters to job.

We can take this as improvement and add changes as required.

vamahaja:
> Maybe implement a `DRY_RUN`. "Would delete repository ceph-wip-nbala-test-1-centos-9-default-x86_64"
> 
> "Leaving distributions X Y Z in repository ceph-tentacle-rocky-10-x86_64 to satisfy keep_minimum: 10"
> 
> Something like those. I'd like to see some dry-run output before we let this loose even though the artifacts in pulp aren't terribly important yet.
> 
> Finished with my review for now.

Good idea !

Added `bool` parameter `DRY_RUN` to job which will list repositories marked for deletion.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,288 @@
+#!/bin/bash
+
+# Purge stale Pulp repositories and distributions per purge-policy.yaml.
+
+set -euo pipefail
+
```
djgalloway, line 7: Add `export PATH="$HOME/.local/bin:$PATH"`

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,288 @@
...
+    refs=$(
+        yq -r ".${project}.${label} | keys[]" "${PURGE_POLICY_FILE}" 2>&1
+    ) || true
+    yq_status=$?
+
+    if [ "${yq_status}" -ne 0 ]; then
+        log "ERROR: Failed to read ${label} policy for ${project}: ${refs}"
+        return 1
+    fi
```
djgalloway:
```suggestion
    local refs

    if [ ! -f "${PURGE_POLICY_FILE}" ]; then
        log "ERROR: Purge policy file not found: ${PURGE_POLICY_FILE}"
        return 1
    fi

    if ! refs=$(
        yq -r ".${project}.${label} | keys[]" "${PURGE_POLICY_FILE}" 2>&1
    ); then
        log "ERROR: Failed to read ${label} policy for ${project}: ${refs}"
        return 1
    fi
```
`$?` after `cmd || true` is always 0 (it's the status of `true`), so this check could never fire — and with `2>&1`, a yq error message would flow through as ref names and end up in the label selector.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,288 @@
...
+        yq -r \
+            ".${project}.default | [(.days // \"\")] | @tsv" \
+            "${PURGE_POLICY_FILE}" 2>&1
+    ) || true
+    yq_status=$?
+
+    if [ "${yq_status}" -ne 0 ]; then
+        log "ERROR: Failed to read default policy for ${project}: ${policy_line}"
+        return 1
+    fi
```
djgalloway:
```suggestion
    local policy_line

    if ! policy_line=$(
        yq -r \
            ".${project}.default | [(.days // \"\")] | @tsv" \
            "${PURGE_POLICY_FILE}" 2>&1
    ); then
        log "ERROR: Failed to read default policy for ${project}: ${policy_line}"
        return 1
    fi
```
Same dead-check pattern: `yq_status` is always 0 after `|| true`.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,288 @@
...
+            --ordering '-pulp_last_updated' \
+            --label-select project="${project}","${label_select}" \
+            2>&1
+    ) || true
+    pulp_status=$?
+
+    if [ "${pulp_status}" -ne 0 ]; then
+        log "ERROR: Failed to list ${type} ${resource}: ${pulp_output}"
+        return 1
+    fi
```
djgalloway:
```suggestion
    local pulp_output

    if ! pulp_output=$(
        pulp "${type}" "${resource}" list \
            --limit "${PULP_LIST_LIMIT}" \
            --ordering '-pulp_last_updated' \
            --label-select project="${project}","${label_select}" \
            2>&1
    ); then
        log "ERROR: Failed to list ${type} ${resource}: ${pulp_output}"
        return 1
    fi
```
Same dead-check pattern: `pulp_status` is always 0 after `|| true`, so a failed `pulp` list (auth error, bad selector, missing binary) is never reported here.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,288 @@
...
+
+    while IFS= read -r name; do
+        [ -z "${name}" ] && continue
+        if [ "${DRY_RUN}" = "true" ]; then
+            log "[DRY RUN] Would destroy ${type} ${resource}: ${name}"
+            continue
+        fi
+        log "Destroying ${type} ${resource}: ${name}"
+        if ! pulp "${type}" "${resource}" destroy \
+                --name "${name}"; then
```
djgalloway:
```suggestion
    local type="$1"
    local resource="$2"
    local resource_list="$3"
    local name lookup_flag

    # pulp-cli looks up rpm distributions with --distribution; everything
    # else (deb distributions, rpm/deb repositories) uses --name.
    lookup_flag="--name"
    if [ "${type}" = "rpm" ] && [ "${resource}" = "distribution" ]; then
        lookup_flag="--distribution"
    fi

    if [ -z "${resource_list}" ]; then
        return 0
    fi

    while IFS= read -r name; do
        [ -z "${name}" ] && continue
        if [ "${DRY_RUN}" = "true" ]; then
            log "[DRY RUN] Would destroy ${type} ${resource}: ${name}"
            continue
        fi
        log "Destroying ${type} ${resource}: ${name}"
        if ! pulp "${type}" "${resource}" destroy \
                "${lookup_flag}" "${name}"; then
```
The pulp-cli version we run doesn't accept `--name` for rpm distributions — it wants `--distribution` (see the `lookup_flag` handling in `scripts/pulp_upload.sh` and the dev pipeline's Jenkinsfile), so purging stale rpm distributions would fail on the first one.

scripts/pulp_upload.sh
```diff
@@ -45,13 +50,19 @@ resolve_os_version_for_repo() {
...
-    printf '%s\n' "$versions"
+    local keep_minimum
+
+    keep_minimum=$(
+        yq -r --arg ref "${BRANCH}" \
+            --arg project "${PULP_PROJECT}" \
+            '.[$project].ref[$ref].keep_minimum
+             // .[$project].default.keep_minimum' \
+            "${PURGE_POLICY_FILE}"
+    )
```
djgalloway, line 81: This adds a hard `yq` dependency to every job that runs `pulp_upload.sh`, but nothing provisions it there — `setup_yq.sh` is only invoked by the new pulp-cleanup Jenkinsfile, and ceph-dev-pipeline never runs it. Since this script runs under `set -ex` and `get_repo_versions_to_retain` is called at top level, an agent without `yq` will fail every dev-build upload the moment this merges.

Two related hazards:

1. **Dialect**: `--arg` (and this jq-style expression) only works with the Python/kislyuk `yq`. If any agent has the Go (mikefarah) `yq` first on `PATH` (common via snap or a manual install), `setup_yq.sh`'s `command -v yq` short-circuit accepts it and this call then fails with an unknown-flag error. Worth verifying the dialect, not just presence — e.g. `yq --help 2>&1 | grep -q jq`.
2. **Availability**: Ubuntu 22.04 (jammy) has no `yq` apt package at all, so `setup_yq.sh` can't install it there even where it is run.

Suggest either wiring `setup_yq.sh` into ceph-dev-pipeline before the upload stage (with a dialect check), or dropping the yq dependency here — e.g. pre-rendering the policy to JSON in the cleanup job only, or parsing with `python3 -c` + PyYAML, which is already on the build agents.

djgalloway:
`cleanup.sh` needs logic to notify shaman of a deleted repo.  See chacra's code: https://github.com/ceph/chacra/blob/main/chacra/asynch/__init__.py#L98-L148.

pulp-cleanup/build/cleanup.sh
djgalloway: Sure, the upload side reads it but the purge side is where chacra enforced `keep_minimum`, and cleanup.sh has no equivalent safeguardfloor. Either the cleanup should skip the newest N per ref before age-filtering, or the field should be renamed so it doesn't imply chacra's guarantee."

pulp-cleanup/build/cleanup.sh
djgalloway: > fetch_pulp_resource_json fetches first 1000 resources

This is my point.  You want it to fetch the _oldest_ 1000 resources.

pulp-cleanup/build/cleanup.sh
vamahaja: done

pulp-cleanup/build/cleanup.sh
vamahaja: fixed, in all places.

pulp-cleanup/build/cleanup.sh
vamahaja: fixed

pulp-cleanup/build/cleanup.sh
vamahaja: fixed

pulp-cleanup/build/cleanup.sh
vamahaja: thanks, updated script.

scripts/pulp_upload.sh
vamahaja: Removed `setup_yq.sh`, added python script to process purge policy yaml.

pulp-cleanup/build/cleanup.sh
vamahaja: I am not expecting 1000 repositories at single point :smile: but if that's the case then better to get last `1000` resources.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,291 @@
...
+    local type="$1"
+    local resource="$2"
+    local project="$3"
+    local label_select="$4"
+    local pulp_output
+
+    if ! pulp_output=$(
+        pulp "${type}" "${resource}" list \
+            --limit "${PULP_LIST_LIMIT}" \
+            --ordering '-pulp_last_updated' \
```
djgalloway: Sorry, I meant to comment [this](https://github.com/ceph/ceph-build/pull/2644#pullrequestreview-4878159591) here.

Drop the `-` from `-pulp_last_updated` to get the _oldest_ 1000.

pulp-cleanup/build/cleanup.sh
djgalloway: This is still not addressed.  Would you like me to add a commit?

djgalloway:
> shaman

This is still not addressed.  I can add a commit if you'd like.

scripts/purge-policy.yaml
```diff
@@ -0,0 +1,61 @@
...
+    reef-nvmeof:
+      keep_minimum: 2
+    squid-nvmeof:
+      keep_minimum: 2
+    squid-nvmeof-8.0:
+      keep_minimum: 2
+    wip-rbd-cgsm-base:
+      keep_minimum: 2
+    hammer:
+      days: 30
```
djgalloway: `ref.days` isn't actually read anywhere in this logic.  Would you like me to add a commit to do so?

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,291 @@
...
+readonly PURGE_POLICY_FILE="${WORKSPACE}/scripts/purge-policy.yaml"
+readonly PURGE_POLICY_PY="${WORKSPACE}/scripts/purge_policy.py"
+readonly DRY_RUN="${DRY_RUN:-false}"
+
+readonly LABEL="ref"
+readonly PULP_LIST_LIMIT=1000
+readonly PULP_TYPES=(rpm deb)
+readonly PULP_RESOURCES=(distribution repository)
+
+# Default protection time in hours
```
djgalloway: You fixed this to be minutes but the comment still says hours

vamahaja: fixed

pulp-cleanup/build/cleanup.sh
vamahaja: changed to descending order.

scripts/purge-policy.yaml
vamahaja: changed approach, it will be used for deleting versions and dist.

pulp-cleanup/build/cleanup.sh
vamahaja: Changed approach to delete dist. and pub based on versions available in repositories.

vamahaja:
@djgalloway updated cleanup script with new approach to cleanup dist, pub and unlisted repos -

**Listed repositories**

1. List distributions labeled `project=ceph,ref=<ref>`.
2. For each distribution, show its publication and read` repository_version`.
3. If the publication is missing, `repository_version` is empty, or that version no longer exists, delete the distribution and its publication.
4. `keep_minimum` is maintained by pulp through `--retain-repo-version` parameter set at the time of repo creation.

**Unlisted repositories**

1. List distributions labeled `project=ceph,ref!=<listed refs...>`.
2. Delete orphaned distributions/publications using the same rules as step 2.
3. List repositories with the same unlisted label selector.
4. For each repository older than `default.days`, delete all remaining distributions (and publications) for that repo, then delete the repo.
5. `keep_minimum` is maintained same as listed repos.

pulp-cleanup/build/cleanup.sh
```diff
@@ -0,0 +1,561 @@
...
+    local type="$1"
+    local project="$2"
+    local repo_name="$3"
+    local dists_json
+
+    if [ "${type}" = "rpm" ]; then
+        if ! dists_json=$(
+            fetch_pulp_list_json "${type}" distribution \
+                --name-contains "${repo_name}" \
+                --ordering 'pulp_last_updated'
```
djgalloway, line 273: nit: I just noticed the deb code below includes `--label-select "project=${project}"` but this check does not.

scripts/pulp_upload.sh
```diff
@@ -45,13 +50,36 @@ resolve_os_version_for_repo() {
...
+        echo "${ref_json}" \
+            | jq -r --arg ref "${BRANCH}" \
+                --argjson default "${default_json}" '
+            .[$ref].keep_minimum // $default.keep_minimum
+        '
+    )
+    printf '%s\n' "${keep_minimum}"
 }
 
 # Create a Pulp repository if it doesn't exist.
```
djgalloway, line 85: "if it doesn't exist" is the problem. `--retain-repo-versions` is only honored by `repository create`, so every repo that already exists (`ceph-tentacle-*`, `ceph-main-*`, `ceph-umbrella-*`) keeps whatever it was created with, and edits to `keep_minimum` in `purge-policy.yaml` never reach them.

That's not cosmetic. Version pruning is the only cleanup a listed ref ever gets — `cleanup.sh` just sweeps the distributions Pulp leaves behind when it drops a version. If `retain_repo_versions` is stale, nothing prunes and the sweep has nothing to find.

Re-apply it on the "already exists" path, but only write when the value actually differs. An unconditional update bumps the repository's `pulp_last_updated`, which is the field `cleanup.sh` uses to decide an unlisted repo is stale, and a build that bails at one of the `skipping content modification and publication` returns would buy a dead repo another 14 days.

GitHub won't let me leave an applyable suggestion down there since it's outside the diff, so: this replaces lines 100-107 (the `local labels_json ...` line through the closing `fi` of the existence check). It widens the `local` declaration too, since the check now keeps the JSON it already fetched.

```bash
    local labels_json key value repo_json current_retain

    log "Checking if Pulp repository ${repo_name} already exists"
    if repo_json=$(
        pulp "${os_pkg_type}" repository show --name "${repo_name}" 2>/dev/null
    ); then
        log "WARNING: Pulp repository ${repo_name} already exists"

        # --retain-repo-versions is only honored by "repository create", so
        # an existing repo keeps whatever it was created with and edits to
        # keep_minimum in purge-policy.yaml never reach it. Version pruning
        # is the only cleanup a listed ref ever gets, so re-apply it here.
        #
        # Only write when the value actually differs. An unconditional
        # update bumps the repository's pulp_last_updated, which is the
        # field pulp-cleanup uses to decide an unlisted repo is stale.
        current_retain=$(
            sed -n '/^{/,$p' <<< "${repo_json}" \
                | jq -r '.retain_repo_versions // empty'
        )
        if [ "${current_retain}" != "${retain_repo_versions}" ]; then
            log "Updating ${repo_name} retain_repo_versions:" \
                "${current_retain:-unset} -> ${retain_repo_versions}"
            if ! pulp "${os_pkg_type}" repository update \
                    --name "${repo_name}" \
                    --retain-repo-versions "${retain_repo_versions}"; then
                log "ERROR: Failed to update retain_repo_versions for" \
                    "${repo_name}"
            fi
        fi
        return 0
    fi
```

Left the failed-update case as a logged ERROR rather than `exit 1` — packages still upload fine when only the retention tweak fails. Happy to make it fatal if you'd rather it be loud.

djgalloway:
1. The PR is still lacking a shaman notification.  You can use https://github.com/ceph/ceph-build/blob/main/scripts/notify_shaman_pulp_repo.sh and https://github.com/ceph/ceph-build/blob/main/ceph-dev-pipeline/build/Jenkinsfile#L596 as examples.
2. `destroy_pulp_resources` returns 1 if any destroy fails.  I'd suggest logging a warning and continuing the same way `destroy_pulp_publication` does.

Other than those and the other 2 comments I added today, I think this is good to go.  Thanks for your work on it.

pulp-cleanup/build/cleanup.sh
vamahaja: this is because `deb` plugin doesn't support `--name-contains`

djgalloway: But what about `--label-select "project=${project}"`

scripts/pulp_upload.sh
vamahaja: @djgalloway it's possible to set `retain_repo_versions` to existing repos, please check below -

```
ceph-build on  pulp-cleanup 
❯ pulp rpm repository list --limit 1 --label-select branch=tentacle
Not all 8 entries were shown.
[
  {
    "pulp_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/",
    "prn": "prn:rpm.rpmrepository:019e69b1-5103-7cee-9d50-27bde8e66574",
    "pulp_created": "2026-05-27T13:48:08.835889Z",
    "pulp_last_updated": "2026-06-17T14:18:37.012726Z",
    "versions_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/versions/",
    "pulp_labels": {
      "ref": "tentacle",
      "arch": "source",
      "branch": "tentacle",
      "distro": "rocky",
      "project": "ceph",
      "distro_version": "10"
    },
    "latest_version_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/versions/5/",
    "name": "ceph-tentacle-rocky-10-source",
    "description": null,
    "retain_repo_versions": null,
    "retain_checkpoints": null,
    "remote": null,
    "autopublish": false,
    "metadata_signing_service": null,
    "package_signing_service": null,
    "package_signing_fingerprint": null,
    "retain_package_versions": 0,
    "checksum_type": null,
    "metadata_checksum_type": null,
    "package_checksum_type": null,
    "sqlite_metadata": false,
    "repo_config": {},
    "compression_type": null,
    "layout": null
  }
]


ceph-build on  pulp-cleanup 
❯ pulp rpm repository update --name ceph-tentacle-rocky-10-source --retain-repo-versions 10
Started background task /pulp/api/v3/tasks/01a090d1-89ce-7d5a-8af1-949cf228034c/
Done.

ceph-build on  pulp-cleanup took 2s 
❯ pulp rpm repository list --limit 1 --label-select branch=tentacle
Not all 8 entries were shown.
[
  {
    "pulp_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/",
    "prn": "prn:rpm.rpmrepository:019e69b1-5103-7cee-9d50-27bde8e66574",
    "pulp_created": "2026-05-27T13:48:08.835889Z",
    "pulp_last_updated": "2026-09-11T14:14:06.572897Z",
    "versions_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/versions/",
    "pulp_labels": {
      "ref": "tentacle",
      "arch": "source",
      "branch": "tentacle",
      "distro": "rocky",
      "project": "ceph",
      "distro_version": "10"
    },
    "latest_version_href": "/pulp/api/v3/repositories/rpm/rpm/019e69b1-5103-7cee-9d50-27bde8e66574/versions/5/",
    "name": "ceph-tentacle-rocky-10-source",
    "description": null,
    "retain_repo_versions": 10,
    "retain_checkpoints": null,
    "remote": null,
    "autopublish": false,
    "metadata_signing_service": null,
    "package_signing_service": null,
    "package_signing_fingerprint": null,
    "retain_package_versions": 0,
    "checksum_type": null,
    "metadata_checksum_type": null,
    "package_checksum_type": null,
    "sqlite_metadata": false,
    "repo_config": {},
    "compression_type": null,
    "layout": null
  }
]
```
I will update all existing repo which are missing `retain_repo_versions` based on `purge-policy.yaml`

vamahaja: @djgalloway I have updated all repos with `retain_repo_versions`, pulp will enforce the retention policy when a new repository version is created for these repos.

vamahaja:
Updated `destroy_pulp_publication` to follow same path as `destroy_pulp_resources`, also added resource count to log number of repos deleted.

PR #2623, opened 2026-06-22
Add optional Pulp container registry upload to ceph-dev-pipeline.

deepssin:
Mirror dev container images from Quay to Pulp when PULP_REGISTRY_UPLOAD is enabled

log - https://jenkins.ceph.com/view/all/job/ceph-dev-pipeline/6447/console

PR #2618, opened 2026-06-18
enhancement: create Pulp repositories with labels before upload

vamahaja:
Add repository creation to the Pulp upload flow so CI jobs ensure the target repo exists and carries labels discovery metadata before packages are uploaded.
- Add `create_pulp_repository()` to idempotently create `rpm`/`deb` repositories, apply metadata labels as JSON and honor `retain-repo-versions`.
- Extract `get_pulp_distribution_labels()` and reuse it for both repository and distribution label application.
- Refactor `publish_pulp_distribution()` to use the shared label helper.
- Require `BRANCH` from the environment instead of deriving it from `CEPH_VERSION` via `release_from_version()`

PR #2645, opened 2026-07-08
CVE pipeline

djgalloway:
## Summary

This adds a `cve-pipeline` (and `cve-source-dist`) job for building packages and containers from ceph-private.git entirely privately — Jenkins build logs, artifacts, packages, and containers. Packages go to the internal-only Pulp instance and containers to the internal-only Quay instance.

To avoid job duplication, the CVE jobs reuse the existing Jenkinsfiles (`cve-pipeline` → `ceph-dev-pipeline`, `cve-source-dist` → `ceph-source-dist`) and differ only in the parameters passed: `CEPH_REPO` points at ceph-private.git, uploads to public chacra are disabled, and the internal Quay credential is selected when `JOB_NAME == 'cve-pipeline'`.

Along the way, ceph-dev-pipeline gains the pieces needed for Pulp-published builds to be fully usable:

- **`CHACRA_UPLOAD` parameter** — chacra upload can now be toggled independently of Pulp upload. `THROWAWAY` still overrides both. cve-pipeline defaults it to `false` so nothing leaks to a public chacra node.
- **Shaman notification for Pulp repos** (`scripts/notify_shaman_pulp_repo.sh`) — mirrors chacra's repo records: `url` is the browsable package repo (the Pulp content distribution for the build's arch, e.g. `.../pulp/content/repos/ceph/<branch>/<sha1>/<distro>/<version>/flavors/<flavor>/<arch>/`), and `chacra_url` is the repository's API endpoint (`/pulp/api/v3/repositories/...`). `pulp_upload.sh` hands off the repository href and package manager version via `$WORKSPACE/pulp_repo_info`.
- **`ceph.repo` in the ceph-release RPM points at the right place** — when `PULP_UPLOAD=true`, the baseurls written into the ceph-release RPM's `/etc/yum.repos.d/ceph.repo` (and the spec's project URL) use the Pulp content location instead of chacra, matching the per-arch distribution layout from #2618 (`$basearch`, `noarch`, `SRPMS`).

## Notes

- Rebased on top of #2618 (Pulp repository creation with labels) and #2623 (Pulp container registry upload); the earlier commits in this branch that created repositories and fixed `BRANCH` handling in `pulp_upload.sh` were dropped as superseded.
- Review feedback addressed: `spec_project_url` is now derived from `repo_base_url` (`"${repo_base_url}/"`), fixing the unintentional `${chacra_url}r/` vs `${chacra_url}/r/` mismatch, and the shaman notify call reuses `spec_project_url` instead of repeating the hardcoded Pulp URL.
- `PULP_SERVER_URL` is hardcoded in both `setup_pulp.sh` and `pulp_upload.sh` (comment added noting they must match), and the Pulp content URL is hardcoded in the Jenkinsfile. If/when the internal CVE Pulp instance lives at a different hostname, these should become a parameter — follow-up material.
- `cve-pipeline.yml` intentionally does not define `PULP_REGISTRY_UPLOAD`; the registry-upload stage skips when unset. It can be added later if CVE builds should mirror containers into the Pulp registry.

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -475,6 +475,7 @@ def doUploadPackagesStage() {
       export FLAVOR="${env.FLAVOR}"
 
       ./scripts/pulp_upload.sh
+      ./scripts/notify_shaman_pulp_repo.sh ready ceph ${os.name} ${os.version_name} ${env.ARCH} https://pulp.front.sepia.ceph.com/pulp/content/repos/ceph/${env.BRANCH}/${env.SHA1}/${os.name}/${os.version_name}/flavors/${env.FLAVOR}/${env.ARCH}/
     else
       echo "Skipping pulp upload because PULP_UPLOAD=${env.PULP_UPLOAD}"
     fi
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -460,7 +460,7 @@ def doUploadPackagesStage() {
     export OS_VERSION="${os.version}"
     export OS_VERSION_NAME="${os.version_name}"
     export OS_PKG_TYPE="${os.pkg_type}"
-    if [ "${env.THROWAWAY}" != "true" ]; then ./scripts/chacra_upload.sh; fi
+    if [ "${env.THROWAWAY}" != "true" ] && [ "${env.CHACRA_UPLOAD}" == "true" ]; then ./scripts/chacra_upload.sh; fi
   """
 
   sh """#!/bin/bash
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -624,7 +624,8 @@ pipeline {
           }
           stage("builder container") {
             environment {
-              CONTAINER_REPO_CREDS = credentials('quay-ceph-io-ceph-ci')
+              // Use quay-int.front cred if cve-pipeline, otherwise default to quay.ceph.io cred
+              CONTAINER_REPO_CREDS = credentials("${env.JOB_NAME == 'cve-pipeline' ? 'quay-int-login' : 'quay-ceph-io-ceph-ci'}")
               DOCKER_HUB_CREDS = credentials('dgalloway-docker-hub')
             }
             when {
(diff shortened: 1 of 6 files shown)


zmc (approved):
Looks good to me. I might consider using names like 'ceph-cve-pipeline' and 'ceph-cve-source-dist', but very much not blocking feedback

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -426,21 +426,34 @@ def doUploadPackagesStage() {
...
+  // Push packages to chacra.ceph.com under the 'test' ref if ceph-release-pipeline's TEST=true
+  if ( os.pkg_type == "rpm" ) {
+    env.SHA1 = env.TEST?.toBoolean() ? 'test' : env.SHA1
+  }
+  // The ceph-release RPM's /etc/yum.repos.d/ceph.repo baseurls and the repo
+  // URL reported to shaman must match where packages were actually
+  // published: pulp when PULP_UPLOAD=true, otherwise chacra.
+  // chacra_url already ends with a trailing slash.
+  def repo_base_url
+  if ( env.PULP_UPLOAD == "true" ) {
```
vamahaja, line 438: This will create ceph-release package pointing to pulp repo and upload to chacra when `CHACRA_UPLOAD` is set to `true`.

Is this expected ?

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -688,7 +703,7 @@ pipeline {
               expression { env.CI_CONTAINER == 'true' && container_distros.contains(env.DIST) }
             }
             environment {
-              CONTAINER_REPO_CREDS = credentials('quay-ceph-io-ceph-ci')
+              CONTAINER_REPO_CREDS = credentials("${env.JOB_NAME == 'cve-pipeline' ? 'quay-int-login' : 'quay-ceph-io-ceph-ci'}")
```
vamahaja, line 706: instead of this better to have separate option for these two repos like `CEPH_CONTAINER_REPO` & `CEPH_CVE_CONTAINER_REPO`.

I would suggest to use pulp container registry with option `PUSH_TO_QUAY_REGISTRY` & `PUSH_TO_PULP_REGISTRY`

ceph-dev-pipeline/build/Jenkinsfile
djgalloway: CHACRA_UPLOAD will never be true for cve-pipeline

ceph-dev-pipeline/build/Jenkinsfile
djgalloway: I'm trying to avoid adding a ton of variables to ceph-dev-pipeline.  cve-pipeline will only be used maybe once a year.  We need to use this today along with https://github.com/ceph/ceph-build/pull/2628.  I've already had to rebase and refactor this PR once.  I need to move on.

PR #2651, opened 2026-07-09
enhancement: split `Chacra` and `Pulp` upload into independent pipeline stages

djgalloway:
Refactor the ceph-dev-pipeline build job so Chacra and Pulp uploads are independently configurable and run as separate Jenkins stages.

Summary:

### Fixed in place (squashed into this PR's commit)

1. **Pulp existence check** — `pulp ... repository list` prints `[]` when nothing matches, which is a non-empty string, so `PULP_REPO_EXISTS` was always `"true"` and every matrix cell skipped compilation. The label query also could never match what pulp_upload.sh stores (`flavor=` vs the stored `flavors` key, `distro_version=noble` vs stored `24.04`, and repositories are branch-scoped so their `sha1` label belongs to whichever build created them). The check now looks up the per-sha1 **distribution by its deterministic name** and uses the exit code, mirroring pulp_upload.sh naming: ubuntu uses the codename, rpm arches use `aarch64` (not `arm64`), and rpm lookups use `--distribution` while deb uses `--name`.

2. **Parallel matrix races** — `env.FOO = ...` assignments are global to the build and shared by every parallel matrix cell (last write wins); only the axis vars themselves get stage-scoped injection. So `CHACRACTL_EXISTS_RC`, `PULP_REPO_EXISTS`, and the `OS_*`/`CEPH_VERSION` values set in "prepare upload" could be overwritten by another cell mid-flight — e.g. the noble cell sets `OS_PKG_TYPE=deb` while centos9 is between "prepare upload" and its Chacra upload. Check results now live in `@Field` maps keyed per cell (like `build_matrix`), and the upload scripts receive `OS_*` via branch-scoped `withEnv`.

3. **Shaman credentials and build status** — the old upload stage's `environment{}` block (which also covered `post{}`) was lost in the split: `update_shaman.sh` in the Chacra stage's `post{}` and `notify_shaman_pulp_repo.sh` in the Pulp stage ran with no `SHAMAN_API_KEY`. Both credentials are now bound on the parent "upload packages" stage, and the completed/failed status update moved there too — otherwise pulp-only builds (cve-pipeline runs `CHACRA_UPLOAD=false`) stayed "started" in shaman forever.

4. **Repo URL bugs** — the URL passed to notify_shaman_pulp_repo.sh was missing a `/` before the arch segment (`.../flavors/defaultx86_64/`), and used `os.version_name` where pulp publishes debian under the numeric version (`debian/bookworm` vs actual `debian/12`).

5. **`PULP_UPLOAD` default stays `false`** in ceph-dev-pipeline.yml. Defaulting it `true` makes pulp.front.sepia reachability a hard requirement of every dev *and* release build before compilation, and posts shaman repo records pointing at Sepia-internal URLs (ceph-release-pipeline doesn't pass `PULP_UPLOAD`, so it inherits this default). cve-pipeline keeps its own default of `true`. If pulp-by-default everywhere is intended, that should be its own deliberate change.

### Follow-up commits on the branch

- **`ce478f5e` key build_matrix by flavor and fix FLAVOR axis** — `build_matrix` was keyed `DIST_ARCH` but the matrix has a FLAVOR axis, so default/debug cells of the same dist/arch shared a key and one flavor's existence check cancelled the other's build. Also fixes the nested `axes {` → `axis {` declaration (both pre-date this PR).
- **`4dfe0d98` cve-pipeline: add FORCE parameter** — the Jenkinsfile consults `FORCE`; without it a CVE rebuild of an already-uploaded sha1 can never be forced.
- **`6f8da0bf` pulp_upload.sh: rpm arch for the repository handoff lookup** — the final `repository show --name ${REPO_NAME}-${ARCH}` used `arm64`, but rpm repos are created as `-aarch64`, so arm64 rpm uploads died under `set -e` after all uploads succeeded.
- **`5feb069d` ceph-release-pipeline: pass boolean parameters as booleans** — `THROWAWAY`/`FORCE`/`FORCE_VERSION`/`TAG` now go through `.toBoolean()`, since the Jenkinsfile compares `params.THROWAWAY == false` strictly.
- **`cc561756` include the flavor in pulp repository and distribution names** — previously default and debug builds of the same branch/sha1 shared repositories and produced identically named distributions, so the stale-distribution delete would clobber one flavor with the other. `FLAVOR` is now part of `REPO_NAME` (e.g. `ceph-main-centos-9-default-x86_64`, `dist-ceph-main-centos-9-default-x86_64-<sha8>`), and the Jenkinsfile existence check matches. Since base_paths are unique in pulp, a distribution created under the old naming that still owns the base_path is deleted before the new one is created. Old unflavored repositories are left in place and can be cleaned up manually; existing sha1s won't be detected by the new check and will rebuild once.
- **`c55882cc` remove the dead TEST='test' chacra ref logic** — nothing defines or forwards `TEST` to ceph-dev-pipeline, and if it were ever enabled, `env.SHA1[0..6]` on the 4-character `'test'` string throws. Removed from the Jenkinsfile and the unused `TEST` parameter dropped from ceph-release-pipeline.

### Verification

The distribution-name existence check was validated against real objects on pulp.front (rc=0 for an existing build, rc=1 for a bogus short sha1) under the *old* naming; the flavored naming shifts everything one rebuild forward but is the same mechanism. None of this has been through a live Jenkins run — suggest a `THROWAWAY=true` smoke test with `CEPH_BUILD_BRANCH=wip-pulp-upload-fixes`, then one real `PULP_UPLOAD=true` run to confirm the flavored repos/distributions land where the check expects.

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -241,15 +241,6 @@ def doCopyArtifactsStage() {
 
 // Stage 5.1 (check for chacra packages): ask shaman whether a chacra repo is ready for this build (and skip compilation if so, unless FORCE).
 def doCheckForChacraPackagesStage() {
-  withCredentials([
-    string(credentialsId: "chacractl-key", variable: "CHACRACTL_KEY"),
-    string(credentialsId: "shaman-api-key", variable: "SHAMAN_API_KEY")
-  ]) {
-    // Configures ~/.chacractl for the upload stage later in this cell.
-    sh """#!/bin/bash -ex
-      ./scripts/setup_chacractl.sh
-    """
-  }
   def os = get_os_info(env.DIST)
   // Uploads go to whichever chacra node shaman's /api/nodes/next/ hands out,
   // so `chacractl exists` against a freshly allocated node only succeeds when
@@ -538,6 +529,14 @@ def withUploadEnv(Closure body) {
 // Stage 8.1 (upload packages to Chacra): build ceph-release RPM (rpm distros) and upload packages to Chacra.
 def doUploadChacraStage() {
   def os = get_os_info(env.DIST)
+  // Point ~/.chacractl at the chacra node shaman allocates for uploads.
+  // Runs here rather than in the check stage (which asks shaman, not
+  // chacra) so that shaman having no healthy chacra nodes only fails
+  // builds that actually need to upload to chacra. CHACRACTL_KEY and
+  // SHAMAN_API_KEY come from the "upload packages" stage environment{}.
+  sh """#!/bin/bash -ex
+    ./scripts/setup_chacractl.sh
+  """
   def chacra_url = sh(
     script: "grep '^url' ~/.chacractl", returnStdout: true,
   ).trim().split('"')[1].trim().replaceAll(/\/+$/, '')
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -563,9 +563,13 @@ def doUploadPulpStage() {
     if (os.pkg_type == "rpm") {
       buildCephReleaseRpm(repo_url)
     }
+    // The shaman record's url must be the flavor-level repo root with the
+    // arch dirs (x86_64/, noarch/, SRPMS/) beneath it, matching chacra's
+    // convention: consumers like ceph.git's Containerfile append
+    // noarch/Packages/... to it to fetch the ceph-release RPM.
     sh """#!/bin/bash -ex
       ./scripts/pulp_upload.sh
-      ./scripts/notify_shaman_pulp_repo.sh ready ceph ${os.name} ${os.version_name} ${env.ARCH} ${repo_url}/${env.ARCH}/
+      ./scripts/notify_shaman_pulp_repo.sh ready ceph ${os.name} ${os.version_name} ${env.ARCH} ${repo_url}/
     """
   }
 }
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -316,40 +316,37 @@ def doArtifactsChecksStage() {
   }
 
   def cell = "${DIST}_${ARCH}_${FLAVOR}"
-  // -1 means "not checked" when that backend's upload is disabled.
-  def chacractl_rc = params.CHACRA_UPLOAD ? chacra_exists_rc[cell] : -1
-  def pulp_repo_exists = params.PULP_UPLOAD && pulp_dist_exists[cell] == true
-  def skip_reasons = []
+  def chacra_exists = chacra_exists_rc[cell] == 0
+  def pulp_exists = pulp_dist_exists[cell] == true
 
-  if (params.CHACRA_UPLOAD && chacractl_rc == 0) {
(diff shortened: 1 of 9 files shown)


vamahaja (approved):

djgalloway:
https://jenkins.ceph.com/job/preserve-ceph-dev-pipeline-dgalloway/8/

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -239,24 +239,34 @@ def doCopyArtifactsStage() {
...
+  // Uploads go to whichever chacra node shaman's /api/nodes/next/ hands out,
+  // so `chacractl exists` against a freshly allocated node only succeeds when
+  // the rotation happens to repeat (e.g. binaries on 2.chacra, check asks
+  // 1.chacra -> false 404). Shaman's repo records are the source of truth for
+  // which node holds a build; pulp records live in the same collection, so
+  // filter on chacra_url.
+  def search_url = "https://shaman.ceph.com/api/search/?project=ceph" +
+      "&distros=${os.name}/${os.version_name}/${env.ARCH}" +
+      "&flavor=${env.FLAVOR}&ref=${env.BRANCH}&sha1=${env.SHA1}"
+  chacra_exists_rc["${DIST}_${ARCH}_${FLAVOR}"] = sh(
```
dmick, line 254: Two concerns/questions:
1) I am not certain if it is possible for there to be more than one response from the shaman query, or if there are, if they would all also have status==ready.  That may be a thing that something prevents from happening.  But if it could happen, we probably want the first (most-recent, [0]) record to match/be ready.

2) it would not hurt to add a test for the chacra URL, to verify that chacra actually claims to have the packages.  chacra/shaman desynchronization happens, because of bugs, asymmetric service outages, etc. and it'll save a failed job attempting to use built repos that end up not actually being present.

scripts/build_container
```diff
@@ -12,27 +12,31 @@ if [[ $CI_CONTAINER == "true" && $DISTRO =~ centos|rocky|alma && "$RELEASE" =~ 8
...
-    if [[ "$ready" == "false" ]] ; then
+    if [[ "$ready" != "true" ]] ; then
       chacra_endpoint="ceph/${BRANCH}/${SHA1}/${DISTRO}/${RELEASE}"
       echo "FAIL: timed out waiting for shaman repo to be built:  https://shaman.ceph.com/api/repos/${chacra_endpoint}/flavors/${FLAVOR}/"
-      # don't fail the build here on purpose
-      # update_build_status "failed" "ceph" $NORMAL_DISTRO $NORMAL_DISTRO_VERSION $NORMAL_ARCH
-      # exit 1
+      # Without a ready repo the container build would install stale or
+      # missing packages, so fail instead of building a wrong image.
+      exit 1
```
dmick, line 39: Yeah, the choice to succeed the build even if the container build didn't happen dates back to Sage, and is probably long-obsolete in intent

dmick (approved):

ceph-dev-pipeline/build/Jenkinsfile
djgalloway: 1. It is possible because of the way we are tricking it to know about pulp repos.  You can get two ArchFlavorBranchRef repos with two separate `chacra_url`s.
2. Good idea: https://github.com/ceph/ceph-build/pull/2651/changes/14d624eaf1b267661cbe977d1a59d5d325695036

PR #2642, opened 2026-07-02
bwc: pin podman auth to a persistent authfile

djgalloway:
podman login was writing credentials to the default location under XDG_RUNTIME_DIR (/run/user/<uid>), which is logind-managed tmpfs. The Jenkins agent runs without XDG_RUNTIME_DIR set, so the runtime dir can be torn down or re-resolved between bwc_login and the podman build it is meant to authenticate. When that happens, the base image pull inside build-with-container.py silently falls back to anonymous and hits Docker Hub's per-IP unauthenticated rate limit (toomanyrequests), e.g.:

  https://jenkins.ceph.com/job/ceph-pull-requests-arm64/99076/

Set REGISTRY_AUTH_FILE to a persistent path in $HOME and pass --authfile to podman login so login and all subsequent podman/buildah invocations in the job resolve the same credential file regardless of session state.

PR #2656, opened 2026-07-15
ceph-dev-pipeline: pin podman auth to a persistent authfile

djgalloway:
## What

Applies the same fix as aae7f2fd (#2653-era bwc.sh change) to the ceph-dev-pipeline builder container stage: `podman login` now writes credentials to a persistent authfile at `$HOME/.config/containers/auth.json` instead of the default location under `XDG_RUNTIME_DIR`.

## Why

`XDG_RUNTIME_DIR` is logind-managed tmpfs that can be torn down or re-resolved between steps. When that happens, credentials silently vanish and the base image pull inside `build-with-container.py` falls back to anonymous, hitting Docker Hub's per-IP unauthenticated rate limit (`toomanyrequests`).

Refs: https://tracker.ceph.com/issues/77920

## Reviewer notes

Unlike the `bwc.sh` fix, an `export` inside the login `sh` block wouldn't work here — each Jenkins `sh` step runs a fresh shell. So `REGISTRY_AUTH_FILE` is set at the Groovy `env` level, which Jenkins exports into every subsequent `sh` step in the matrix cell. That means the credential file is also resolved by the later `podman pull`/`push` steps, the build stage's `build-with-container.py` runs (the actual rate-limited path), and the container stage's `./scripts/build_container` — podman honors the `REGISTRY_AUTH_FILE` env var even without an explicit `--authfile` flag.

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -346,10 +346,22 @@ def doArtifactsChecksStage() {
 // Stage 6 (builder container): log into container registries and pull/build/push the ceph-build container image for this matrix cell.
 def doBuilderContainerStage() {
   env.CEPH_BUILDER_IMAGE = "${env.CONTAINER_REPO_HOSTNAME}/${env.CONTAINER_REPO_ORGANIZATION}/ceph-build"
+  // Pin podman auth to a persistent authfile. The agent runs as a systemd
+  // service with no login session and no XDG_RUNTIME_DIR, so podman derives
+  // the default auth.json location from ambient node state at each
+  // invocation, landing in runtime tmpfs the job doesn't control.
+  // Credentials written by podman login here have been observed unavailable
+  // to podman build in the same job minutes later, so the base image pull
+  // goes anonymous and hits Docker Hub's rate limit (toomanyrequests,
+  // https://tracker.ceph.com/issues/77920). Set via env so every subsequent
+  // sh step (pull/build/push here, and build-with-container.py in the build
+  // stage) resolves the same file.
+  env.REGISTRY_AUTH_FILE = "${env.HOME}/.config/containers/auth.json"
   sh '''#!/bin/bash
     set -ex
-    podman login -u ${CONTAINER_REPO_CREDS_USR} -p ${CONTAINER_REPO_CREDS_PSW} ${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}
-    podman login -u ${DOCKER_HUB_CREDS_USR} -p ${DOCKER_HUB_CREDS_PSW} docker.io
+    mkdir -p "${REGISTRY_AUTH_FILE%/*}"
+    podman login --authfile ${REGISTRY_AUTH_FILE} -u ${CONTAINER_REPO_CREDS_USR} -p ${CONTAINER_REPO_CREDS_PSW} ${CONTAINER_REPO_HOSTNAME}/${CONTAINER_REPO_ORGANIZATION}
+    podman login --authfile ${REGISTRY_AUTH_FILE} -u ${DOCKER_HUB_CREDS_USR} -p ${DOCKER_HUB_CREDS_PSW} docker.io
   '''
   def ceph_builder_tag_short = "${env.BRANCH}.${env.DIST}.${env.ARCH}.${env.FLAVOR}"
   def ceph_builder_tag = "${env.SHA1[0..6]}.${ceph_builder_tag_short}"


ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -346,10 +346,18 @@ def doArtifactsChecksStage() {
 // Stage 6 (builder container): log into container registries and pull/build/push the ceph-build container image for this matrix cell.
 def doBuilderContainerStage() {
   env.CEPH_BUILDER_IMAGE = "${env.CONTAINER_REPO_HOSTNAME}/${env.CONTAINER_REPO_ORGANIZATION}/ceph-build"
+  // Pin podman auth to a persistent authfile. The default location is under
+  // XDG_RUNTIME_DIR (logind-managed tmpfs), which can be torn down or
+  // re-resolved between sh steps, silently dropping credentials and hitting
```
idryomov: Just curious, what exactly is meant by "torn down" and "re-resolved"?  For `$XDG_RUNTIME_DIR` https://specifications.freedesktop.org/basedir/latest/ says:
> The lifetime of the directory MUST be bound to the user being logged in. It MUST be created when the user first logs in and if the user fully logs out the directory MUST be removed. If the user logs in more than once they should get pointed to the same directory, and it is mandatory that the directory continues to exist from their first login to their last logout on the system, and not removed in between. Files in the directory MUST not survive reboot or a full logout/login cycle.

Is full logout/login possible between sh steps in question?

> Files in this directory MAY be subjected to periodic clean-up. To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.

If not, do you see files getting cleaned up sooner than after 6 hours?

djgalloway: I dug into this on the node that hit the failure (toko06), and I have to walk back the "torn down" language from my commit message as the evidence doesn't support it there. `journalctl -u systemd-logind` shows zero sessions since boot (the Jenkins agent runs as a systemd service, no PAM session at all), we run `loginctl enable-linger` on these nodes, and `/run/user/<uid>` has existed untouched since boot. So a logout/login cycle did not happen between the sh steps, and the 6-hour cleanup clause isn't implicated either.  systemd-tmpfiles-clean ran at 10:49 that day, nowhere near the failure window.

What I can say from the same investigation: `XDG_RUNTIME_DIR` is unset in the agent's env, so podman derives the auth.json location itself on each invocation.  On this node it lands on `/run/user/<uid>/containers/auth.json` (verified with `podman --log-level=debug`), which only exists because of linger. The failure was: podman login succeeded, and the FROM docker.io/ubuntu:22.04 pull inside podman build went anonymous 60 seconds later, same node, same env. This rules out logind teardown, tmpfiles cleanup, and concurrent jobs (one executor per node); I haven't yet caught what actually made the credentials unavailable, and I've got an audit watch on the file to catch it if it recurs.

The reason I still think this is the right thing to do: every default location podman can pick here is runtime tmpfs with a lifecycle owned by something else (logind sessions, tmpfiles aging, podman's own /tmp fallback heuristics when the runtime dir is unusable), and which one gets picked depends on ambient node state. Pinning REGISTRY_AUTH_FILE to $HOME takes all of that out of the picture — one persistent path, same for login and every later podman invocation, no dependency on session or tmpfs state.

This was literally all within the same ~5 minute span.

```
+ podman login -u **** -p **** quay.ceph.io/ceph-ci
Login Succeeded!
+ podman login -u **** -p **** docker.io
Login Succeeded!
+ podman pull quay.ceph.io/ceph-ci/ceph-build:8ef46bd.wip-dis-testing.jammy.x86_64.default
Trying to pull quay.ceph.io/ceph-ci/ceph-build:8ef46bd.wip-dis-testing.jammy.x86_64.default...
Error: initializing source docker://quay.ceph.io/ceph-ci/ceph-build:8ef46bd.wip-dis-testing.jammy.x86_64.default: reading manifest 8ef46bd.wip-dis-testing.jammy.x86_64.default in quay.ceph.io/ceph-ci/ceph-build: manifest unknown
+ podman pull quay.ceph.io/ceph-ci/ceph-build:wip-dis-testing.jammy.x86_64.default
Trying to pull quay.ceph.io/ceph-ci/ceph-build:wip-dis-testing.jammy.x86_64.default...
Getting image source signatures
Copying blob sha256:8fb8c08b4f5aba58d73fc34364d88fed7b6932734f31e842ceb163e7ac3916ad
Copying blob sha256:ef2564f56659ba036af01ca29ac087afc98fce7ecec3212662529dbfa1f36c26
Copying blob sha256:e27fa87580538c0f02be7bb071a751015ab530006085e6a816cb5fdb370b636e
Copying blob sha256:b71a4b3372d639be2da2d9e0749e7f2d35cd7dd9577f8988ba4cffd4f7de54a6
Copying config sha256:3a679e9f9eb0cf72c0c7f1f86a69681539e9b0a7e2818b13c1e3dbfe161c363d
Writing manifest to image destination
3a679e9f9eb0cf72c0c7f1f86a69681539e9b0a7e2818b13c1e3dbfe161c363d
+ echo
+ echo WITH_CRIMSON=true
+ cd dist/ceph
+ python3 src/script/build-with-container.py --env-file=/home/jenkins-build/build/workspace/ceph-dev-pipeline/.env --image-repo=quay.ceph.io/ceph-ci/ceph-build --tag=8ef46bd.wip-dis-testing.jammy.x86_64.default --image-variant=packages -d jammy -e build-container
2026-07-13 19:49:50,044: INFO: want to execute build step: build-container
2026-07-13 19:49:50,044: INFO: want to execute build step: dnfcache
2026-07-13 19:49:50,045: INFO: step done: dnfcache completed in 00:00:00
2026-07-13 19:49:50,045: INFO: Executing command: podman build --pull -t quay.ceph.io/ceph-ci/ceph-build:8ef46bd.wip-dis-testing.jammy.x86_64.default --label=io.ceph.build-with-container.src=sha256:5150fa786406c9d74bc021d4de5706465b7795c0dffb03f7d6c4166d826d8c76 --label=io.ceph.build-with-container.image-variant=packages.crimson --build-arg=CEPH_BASE_BRANCH=main --build-arg=DISTRO=docker.io/ubuntu:22.04 --build-arg=FOR_MAKE_CHECK=false --build-arg=WITH_CRIMSON=1 -f Dockerfile.build .
[1/2] STEP 1/5: FROM scratch AS bootstrap
[1/2] STEP 2/5: ARG CEPH_CTR_SRC=/usr/local/src/ceph
--> 51b0fca7f438
[1/2] STEP 3/5: COPY     src/script/lib-build.sh     src/script/run-make.sh     ${CEPH_CTR_SRC}/src/script/
--> f4bac092c67b
[1/2] STEP 4/5: COPY debian ${CEPH_CTR_SRC}/debian
--> 841b420dad18
[1/2] STEP 5/5: COPY     ceph.spec.in     do_cmake.sh     install-deps.sh     run-make-check.sh     src/script/buildcontainer-setup.sh     ${CEPH_CTR_SRC}/
--> bd247104990f
[2/2] STEP 1/13: FROM docker.io/ubuntu:22.04
Trying to pull docker.io/library/ubuntu:22.04...
time="2026-07-13T19:50:22Z" level=warning msg="Failed, retrying in 2s ... (1/3). Error: initializing source docker://ubuntu:22.04: reading manifest 22.04 in docker.io/library/ubuntu: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
```

idryomov: > I haven't yet caught what actually made the credentials unavailable

I don't like this part, but your reasoning for still making the change seems sensible to me.  If nothing else, being consistent with https://github.com/ceph/ceph-build/pull/2642 shouldn't hurt.

idryomov (approved):

PR #2660, opened 2026-07-17
ceph-release-containers: build umbrella+ release containers from Rocky Linux 10

djgalloway:
The release container job never set `FROM_IMAGE`, so `container/build.sh` in ceph.git fell back to its default (`quay.io/centos/centos:stream9`) for every release.

Starting with umbrella (v21), release containers should be based on Rocky Linux 10, matching what the dev/CI pipeline already does via `scripts/build_container`.

This sets `FROM_IMAGE` explicitly from the `VERSION` parameter: Rocky 10 for major version >= 21, CentOS Stream 9 for older releases.

A companion ceph.git PR flips the `FROM_IMAGE` defaults in `container/` to Rocky 10 on main so umbrella and later default correctly even when `FROM_IMAGE` isn't passed.

### DIFF
--- a/ceph-release-containers/build/Jenkinsfile
+++ b/ceph-release-containers/build/Jenkinsfile
@@ -97,6 +97,7 @@ pipeline {
           )
           script {
             sh '''#!/bin/bash -ex
+              command -v skopeo || sudo dnf install -y skopeo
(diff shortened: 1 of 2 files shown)


djgalloway:
Ran here https://jenkins.ceph.com/view/all/job/ceph-release-containers/40/stages/

dmick (approved):

PR #2670, opened 2026-07-29
Fix for ppc-make-check job

anshuman-agarwala:


hasan4791:
Lets update the label to "ppc64le" in the yml and also if possible we can rename the file too!!

zmc (approved):
Look

anshuman-agarwala:
> Lets update the label to "ppc64le" in the yml and also if possible we can rename the file too!!

Done

hasan4791:
> > Lets update the label to "ppc64le" in the yml and also if possible we can rename the file too!!
> 
> Done

We are good to merge.

PR #2674, opened 2026-08-06
ceph-pr-pipeline: combine the ceph.git PR checks into one pipeline

djgalloway:
Replaces the GitHub Pull Request Builder plugin and six freestyle jobs
(ceph-pr-commits, ceph-pr-submodules, ceph-pull-requests,
ceph-pull-requests-arm64, ceph-pr-api, ceph-windows-pull-requests) with a
webhook trigger job and one pipeline. The GitHub status contexts are
unchanged, so branch protection rules don't change.

## What it does

- **One clone, one build.** ceph.git is checked out once and compiled once
  (x86_64, `bwc -e buildtests`); make check and the API tests run on separate
  builders from the shared tree, moved via a zstd tarball in S3. Signed-off-by
  and Unmodified Submodules are reimplemented against the GitHub API and never
  clone at all. Windows and arm64 run as parallel legs with their own builds
  (their binaries were never shareable); arm64 restores-or-builds on a single
  node.
- **Binary cache.** Tarballs are keyed by `pr-builds/<PR>/<head sha>.<arch>`,
  so any re-run for an unchanged head skips compilation (`FORCE_BUILD`
  overrides).
- **Gating.** ceph-org authors build automatically; everyone else needs the
  `ci-approved` label. Every push removes the label, cancels running builds
  for the PR, and resets statuses to pending — it must be re-added for the new
  code.
- **Phrases.** `jenkins retest ...` re-runs everything (cheap for an unchanged
  head); `jenkins test <check>` runs one named leg; bare `jenkins test` runs
  nothing.
- **Statuses.** Each leg posts its own context. Pending is posted seconds into
  the run, before legs wait for executors; canceled/failed runs finalize any
  still-pending contexts so nothing hangs at "queued" on GitHub.

## Rollout

Designed to run side by side with the old jobs (`STATUS_PREFIX` keeps test
statuses away from the required contexts). See
[ceph-pr-pipeline/README.md](ceph-pr-pipeline/README.md) for the checklist:
credentials, one-time sandbox signature approvals, webhook, flip.

## Validated so far

- Quick checks tested against real PRs (including the multi-file JSON
  pagination fix in 70c3695a)
- Manual pipeline runs on jenkins.ceph.com through prepare/quick-checks/build
- Still to validate before the flip: API tests inside the build container,
  windows leg on a known-good libvirt node, arm64 leg, cobertura publishing
  (not wired yet)

### DIFF
--- a/ceph-pr-pipeline/README.md
+++ b/ceph-pr-pipeline/README.md
@@ -129,8 +129,15 @@ for the same PR (covers manual runs and webhook races).  A build whose
      'method hudson.model.Job getBuilds',
      'method hudson.model.Run getNumber',
      'method hudson.model.Run isBuilding',
+     // On newer cores these resolve to the HistoricalBuild interface instead
+     'method jenkins.model.HistoricalBuild getNumber',
+     'method jenkins.model.HistoricalBuild isBuilding',
      'method hudson.model.Actionable getAction java.lang.Class',
      'method hudson.model.ParametersAction getParameter java.lang.String',
+     // StringParameterValue is the narrow variant; keep only whichever one
+     // the sandbox actually asks for (the broad ParameterValue one can read
+     // password parameters, the narrow one cannot)
+     'method hudson.model.StringParameterValue getValue',
      'method hudson.model.ParameterValue getValue',
      'method hudson.model.Run getExecutor',
      'method hudson.model.Executor interrupt hudson.model.Result',
--- a/ceph-pr-pipeline/README.md
+++ b/ceph-pr-pipeline/README.md
@@ -11,10 +11,12 @@ clone and (for three of them) its own multi-hour build:
 | `make check` | [ceph-pull-requests](../ceph-pull-requests/) | shared build → ctest on its own builder |
 | `ceph API tests` | [ceph-pr-api](../ceph-pr-api/) | shared build → API tests on its own builder |
 | `ceph windows tests` | [ceph-windows-pull-requests](../ceph-windows-pull-requests/) | parallel leg (different binaries, nothing to share) |
+| `make check (arm64)` | [ceph-pull-requests-arm64](../ceph-pull-requests-arm64/) | parallel leg: own build+test on one arm64 node, own cache key |
 
-`make check (arm64)` ([ceph-pull-requests-arm64](../ceph-pull-requests-arm64/))
-deliberately stays a separate job: it can never reuse the x86_64 binaries, so
-folding it in would add matrix complexity for zero savings.
+arm64 can never reuse the x86_64 binaries, so its leg does its own build --
+but it shares the pipeline's gating, cancellation, early pending statuses,
+and the S3 cache (keyed per arch), so re-runs skip the arm64 compile too.
+It builds and tests on a single node because the arm64 pool is small.
 
 The GitHub status contexts are identical to the old jobs, so **branch
 protection rules do not change**.
@@ -36,6 +38,7 @@ webhook (pull_request / issue_comment)
             ├─ Signed-off-by         (small)  GitHub API, no clone
             ├─ Unmodified Submodules (small)  GitHub API, no clone
             ├─ ceph windows tests    (libvirt) own clone + win32 build + tests
+            ├─ build and test (arm64)(arm64)  own clone, restore-or-build + tests
             └─ build and test
                  build ceph          (huge)   ONE clone, bwc -e buildtests,
                  │                            tree → S3 (skipped if cached)
@@ -83,12 +86,11 @@ Implemented in the trigger job:
   the contexts to pending.  The label must be re-added after each push —
   including force-pushes — so approval always refers to reviewed code.
 - Comment phrases (org members always; others only while the label is
-  present):
-  `jenkins retest` (all checks), `jenkins test make check`, `jenkins test api`, `jenkins test windows`,
-  `jenkins test signed`, `jenkins test submodules`, `jenkins test all`.
-  `jenkins test make check arm64` is left to the arm64 job's own GHPRB
-  trigger.  `jenkins do not test` in the PR description still skips
-  auto-builds.
+  present): `jenkins retest ...` re-runs **everything** (the S3 cache makes
+  that cheap for an unchanged head); `jenkins test <check>` runs one leg and
+  requires naming it -- `make check`, `make check arm64`, `api`, `windows`,
+  `signed`, `submodules`.  A bare `jenkins test` runs nothing.  `jenkins do
+  not test` in the PR description still skips auto-builds.
 
 ## Force-push / cancellation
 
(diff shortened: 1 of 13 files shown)


djgalloway:
Screenshot 2026-08-07 at 7 09 24 PM
Screenshot 2026-08-07 at 7 10 10 PM

ceph-pr-pipeline/build/Jenkinsfile
```diff
@@ -0,0 +1,489 @@
...
+  }
+  if (wantCheck("windows") && windows_skip &&
+      windows_branches.contains(env.ghprbTargetBranch)) {
+    postStatus("windows", "success", "no need to run windows tests")
+  }
+}
+
+def doSignedCheck() {
+  checkoutCephBuild()
+  withStatus("signed", "commit signature check") {
```
phlogistonjohn, line 266: Sorry if this sounds extremely nitpicky but when I see "signed" I think  cryptographically signed as opposed to has a signed-off-by trailer (DCO).

ceph-pr-pipeline/build/Jenkinsfile
```diff
@@ -0,0 +1,489 @@
...
+
+def restoreBuildTree() {
+  checkoutCephBuild()
+  sh "./scripts/setup_container_runtime.sh"
+  sh(
+    label: "download build tree from cache",
+    script: bwcPreamble() + """
+      rm -rf ceph
+      s3_setup
+      s3_cache_get ceph
```
phlogistonjohn, line 379: There are a few things in the build that end up writing into the source tree. Does this capture and sync those too?

ceph-pr-pipeline/build/Jenkinsfile
```diff
@@ -0,0 +1,489 @@
...
+    sh(
+      label: "run make check",
+      script: bwcPreamble() + """
+        cd ceph
+        ${writeBwcEnvFile()}
+        bwc 4 -e tests
+        sleep 5
+        ps -ef | grep -v jnlp | grep ceph || true
+
+        # "bwc -e tests" exits 0 even when individual tests fail so the CTest
```
phlogistonjohn: Would you like this behavior to change?

ceph-pr-pipeline/README.md
```diff
@@ -0,0 +1,154 @@
...
+  Daemon-heavy make check tests already run in that container, so this is
+  expected to work.
+- Squid-targeted PRs: the old `ceph-api-squid` job ran on jammy *hosts*; here
+  everything is containerized so the host OS no longer matters, but squid PRs
+  should be spot-checked.
+- The GHPRB "rebuild" button semantics are replaced by re-running the
+  pipeline build (Jenkins "Rebuild" with the same parameters works and will
+  hit the binary cache).
+- `build-with-container.py` names its container `ceph_build`, so two legs of
+  the same (or different) PRs must not share a builder concurrently; with
```
phlogistonjohn: If this becomes an issue we can give an option to set this if needed.

ceph-pr-pipeline/README.md
```diff
@@ -0,0 +1,154 @@
...
+webhook (pull_request / issue_comment)
+  └─ ceph-pr-pipeline-trigger        gating, cancellation, label handling
+       └─ ceph-pr-pipeline
+            prepare                  (small)  resolve PR, classify changed files
+            ├─ Signed-off-by         (small)  GitHub API, no clone
+            ├─ Unmodified Submodules (small)  GitHub API, no clone
+            ├─ ceph windows tests    (libvirt) own clone + win32 build + tests
+            └─ build and test
+                 build ceph          (huge)   ONE clone, bwc -e buildtests,
+                 │                            tree → S3 (skipped if cached)
```
phlogistonjohn: Would you also want to reuse the builder container image? There's a small chance that if the job lands on a different host, just by timing the containers could end up slightly different. saving the image to s3 or whatnot and loading it on the other side would guarantee it's using the same image that was used to build the binaries.

phlogistonjohn (commented):
Some basic feedback (from just the 1st commit)

ceph-pr-pipeline/build/Jenkinsfile
djgalloway: I'm just duplicating the existing keyword https://github.com/ceph/ceph-build/blob/main/ceph-pr-commits/config/definitions/ceph-pr-commits.yml#L57

ceph-pr-pipeline/build/Jenkinsfile
djgalloway: Good question.  We tar up the whole `ceph/` dir https://github.com/ceph/ceph-build/pull/2674/changes/BASE..ce63f7b3939429ff6c5cfe9d5ea38b677953269b#diff-8d67e39c66fdcf4032b8400667fb009bb5fda37895946903576d64880768a88bR37-R48 but _before_ we run any tests.  So there shouldn't be any test artifacts included.

https://github.com/ceph/ceph-build/pull/2674/changes/BASE..ce63f7b3939429ff6c5cfe9d5ea38b677953269b#diff-8d67e39c66fdcf4032b8400667fb009bb5fda37895946903576d64880768a88bR37-R48

ceph-pr-pipeline/build/Jenkinsfile
djgalloway: Sure.  Can it be done for `tests` only?  One less `if` statement in the ceph-build tree.

ceph-pr-pipeline/README.md
djgalloway: The win here would be determinism (i.e., the identical built environment), not speed, at the cost of 1.5-2GB per PR.

We run https://github.com/ceph/ceph/blob/main/Dockerfile.build#L37 so many times already, the local host cache is usually populated.  A job I just checked took 3 seconds.

ceph-pr-pipeline/build/Jenkinsfile
phlogistonjohn: ok

ceph-pr-pipeline/README.md
phlogistonjohn: ok. this is not set in stone and could be revisited if needed

ceph-pr-pipeline/build/Jenkinsfile
phlogistonjohn: I think this is a general issue in `run-make-check.sh`:
```
    if in_jenkins; then
        if ! ctest $CHECK_MAKEOPTS --no-compress-output --output-on-failure --test-output-size-failed 1024000 -T Test; then
            # do not return failure, as the jenkins publisher will take care of this
            rm -fr ${TMPDIR:-/tmp}/ceph-asok.* ${CTEST_RESOURCE_FILE}
        fi
    else
        if ! $DRY_RUN ctest $CHECK_MAKEOPTS --output-on-failure; then
            rm -fr ${TMPDIR:-/tmp}/ceph-asok.* ${CTEST_RESOURCE_FILE}
            return 1
        fi
    fi
```

We're (almost certainly) running the 1st branch, where `in_jenkins` returns true. Note that the ctest command fails it'll run the `rm -fr` command and that exit status (0) will be used.
We can change it to match the behavior of the other branch easily, but I don't know if this is what we want - if this difference is intentional it could have other side-effects.

Making it only apply to the `-e tests` would be difficult because it's not bwc "suppressing" the error.  I suppose we could do env var shenanigans and add an new env var that says, "give me the real exit code of ctests in this case"... WDYT?

phlogistonjohn: Somehow I missed the comment... this is clearly intentional.

dmick (approved):

epuertat:
🤯 this is superb! I see this is bringing the `Jenkinsfile` syntax. How about moving this to the ceph/ceph `main` (and optionally last 2 releases)? That way CI improvements in `main` don't cause regressions in other branches CI, and we don't need 2-3 PRs in 2 different repos whenever we want to improve CI.

PR #2675, opened 2026-08-07
tcmu-runner: remove centos7 and centos8

djgalloway:
## What

Removes centos7 and centos8 from the tcmu-runner matrix job, leaving centos9 and rocky10:

- `DISTROS` parameter default and description now list only `centos9 rocky10`
- `AVAILABLE_DIST` axis drops `centos7` and `centos8` (rocky10 has no entry because it builds in a container on centos9 nodes)
- The combination-filter keeps the rocky10-onto-centos9 routing but drops the arm64 clause — with centos7 gone, every remaining distro builds on both arches

## Notes for reviewers

Rebased on main after the rocky10 support from #2627 landed. The build scripts (`build/setup`, `build/build_rpm`) detect the distro at runtime and needed no changes. `tcmu-runner-trigger` has no distro references.

### DIFF
--- a/tcmu-runner/config/definitions/tcmu-runner.yml
+++ b/tcmu-runner/config/definitions/tcmu-runner.yml
@@ -24,8 +24,8 @@
 
       - string:
           name: DISTROS
-          description: "A list of distros to build for. Available options are: centos7, centos8, centos9, rocky10"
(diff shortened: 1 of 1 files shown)


ChristinaMeno (approved):
Looks good. Merge away

PR #2677, opened 2026-08-11
pulp_upload: label deb distributions with the distro codename

djgalloway:
teuthology's `PulpProject` searches Pulp distributions via `pulp_label_select` using the distro **codename** for deb-based distros (`distro_version=jammy`), matching the chacra/shaman convention. `pulp_upload.sh` was labeling distributions with the numeric version (`distro_version=22.04`) even though the base_path already uses the codename, so every Ubuntu job died in `check_packages`:

```
teuthology.exceptions.VersionNotFoundError: Failed to fetch package version from No results found for https://pulp.front.sepia.ceph.com/pulp/api/v3/distributions/deb/apt
```

(e.g. `yuriw-2026-08-10_23:39:44-smoke-wip-yuriw-cve-dryrun_tentacle-distro-default-trial/453537` — the packages exist at `.../repos/ceph/wip-yuriw-cve-dryrun_tentacle/<sha1>/ubuntu/jammy/flavors/default/x86_64/`, but the search matched zero distributions.)

This uses `OS_VERSION_NAME` for the `distro_version` label on deb builds (the Jenkinsfile passes it for both Ubuntu and Debian) so the label matches what teuthology queries for. RPM distros are unaffected; both sides already use the major version there.

Existing distributions for in-flight branches can be repaired without a rebuild:

```
pulp deb distribution label set --name <dist-name> --key distro_version --value jammy
```

### DIFF
--- a/scripts/pulp_upload.sh
+++ b/scripts/pulp_upload.sh
@@ -13,7 +13,7 @@
 #   SHA1            - Git commit SHA for this build
 #   OS_NAME         - Target OS (e.g. centos, ubuntu)
 #   OS_VERSION      - Target OS version
-#   OS_VERSION_NAME - Ubuntu codename (used when OS_NAME=ubuntu)
+#   OS_VERSION_NAME - Distro codename, e.g. jammy, bookworm (used when OS_PKG_TYPE=deb)
 #   OS_PKG_TYPE     - Package format: "rpm" or "deb"
 #   FLAVOR          - Build flavor label (e.g. default, crimson)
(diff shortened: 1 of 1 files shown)


djgalloway:
Closing: we want to keep the shaman/chacra convention of numeric distro versions (22.04) rather than codenames, so the labels pulp_upload.sh sets are correct as-is. The fix belongs on the teuthology side instead: ceph/teuthology#2241 will make PulpProject search with the numeric version like ShamanProject does.

PR #2679, opened 2026-08-17
Fixed werror flag for ppc64le

anshuman-agarwala:


djgalloway:
The other two architectures do not do this.  Is there a particular warning that has to be ignored on ppc64?

anshuman-agarwala:
Yes there are certain warnings that pop up only on ppc64le architecture. 
From what I can tell this is due to us using gcc instead of clang.
We've tried running the build on clang before but there are certain issues we hit on power architecture.
Eventually we want to get rid of this flag but for now it's a stopgap.

For reference, this job has failed due to warnings being treated as errors: https://jenkins.ceph.com/job/ceph-make-check-periodic-ppc64le/2/console

djgalloway:
Ok, as long as it's temporary.

djgalloway (approved):

PR #2680, opened 2026-08-17
ceph-dev-pipeline: fix SHA1 check race between parallel matrix cells

djgalloway:
Fixes the intermittent failure seen in https://jenkins.ceph.com/job/ceph-dev-pipeline/8303/ (and #8292):

```
SHA1 from artifact (50202b62fe9071e2bd335e0cfff5d2f37c9ef12a) does not match parameter value ()
```

When the `SHA1` parameter is empty (ceph-dev-cron leaves it blank), each matrix cell's "copy artifacts" stage fills `env.SHA1` in from the `dist/sha1` artifact. `env` is shared across the parallel cells, so a sibling cell can set `env.SHA1` between one cell's read of `sha1_trimmed` (`""`) and its `if (env.SHA1 && ...)` check, which then spuriously fails.

Compare against the immutable job parameter (`params.SHA1`) instead, and always publish the artifact's SHA1 into `env.SHA1` (identical to the parameter when one was given).

### DIFF
--- a/ceph-dev-pipeline/build/Jenkinsfile
+++ b/ceph-dev-pipeline/build/Jenkinsfile
@@ -186,15 +186,19 @@ def doCopyArtifactsStage() {
     filter: artifact_filter,
   )
   sh 'sudo journalctl --show-cursor -n 0 --no-pager | tail -n1 | cut -d\" \" -f3 > $WORKSPACE/cursor'
-  def sha1_trimmed = env.SHA1.trim().toLowerCase()
+  // env is shared across the parallel matrix cells, and when the SHA1 parameter
+  // is empty each cell fills env.SHA1 in from the artifact below. Compare against
+  // the immutable job parameter (params.SHA1) rather than re-reading env.SHA1,
+  // otherwise a sibling cell can set env.SHA1 between our read of sha1_trimmed
+  // and the check, and we'd spuriously fail with "does not match parameter value ()".
+  def sha1_trimmed = (params.SHA1 ?: "").trim().toLowerCase()
   def sha1_props = readProperties file: "${env.WORKSPACE}/dist/sha1"
   def sha1_from_artifact = sha1_props.SHA1.trim().toLowerCase()
-  if ( env.SHA1 && sha1_from_artifact != sha1_trimmed ) {
+  if ( sha1_trimmed && sha1_from_artifact != sha1_trimmed ) {
     error message: "SHA1 from artifact (${sha1_from_artifact}) does not match parameter value (${sha1_trimmed})"
-  } else if ( ! env.SHA1 ) {
-    env.SHA1 = sha1_from_artifact
   }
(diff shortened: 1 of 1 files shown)


djgalloway:
Note for reviewers: this isn't new — sampling cron-tick (14:45/20:08 UTC) ceph-dev-pipeline failures, the same `does not match parameter value ()` error shows up in at least one build per tick going back to mid-July (e.g. #7125, #7170, #7272, #7288, #7327, #7372, #7495, #7590, #7928, #7954, #8013, #8014, #8036, #8092, #8098, #8100, #8214, #8250-8254, #8264, #8278, #8292, #8303). It only bites at cron ticks because ~5 pipeline builds start simultaneously and controller step latency balloons — `readProperties` on the 40-byte dist/sha1 takes 5-18 s vs 0.1-1 s off-peak — which turns a millisecond race window into a many-second one.

ceph-dev-pipeline/build/Jenkinsfile
```diff
@@ -186,15 +186,19 @@ def doCopyArtifactsStage() {
...
   def sha1_props = readProperties file: "${env.WORKSPACE}/dist/sha1"
   def sha1_from_artifact = sha1_props.SHA1.trim().toLowerCase()
-  if ( env.SHA1 && sha1_from_artifact != sha1_trimmed ) {
+  if ( sha1_trimmed && sha1_from_artifact != sha1_trimmed ) {
     error message: "SHA1 from artifact (${sha1_from_artifact}) does not match parameter value (${sha1_trimmed})"
-  } else if ( ! env.SHA1 ) {
-    env.SHA1 = sha1_from_artifact
   }
-  println "SHA1=${sha1_trimmed}"
+  env.SHA1 = sha1_from_artifact
```
dmick, line 200: if env is shared, this seems worrisome (like any updates to env, really; multiple writers is uncomfortable).  Should we be using env at all for results of code evaluation inside parallel paths?  Does it only make invoking the subcommands slightly easier, and should it maybe be invoking the commands with a copy of the local sha1_from_artifact rather than trying to pipeline it through env?

djgalloway: SHA1 as a parameter is optional so it has to be an env var.  All the cells read the same `dist/sha1` from the same ceph-source-dist job, so they all write the same value; the bug was the check using `env.SHA1` as "did the user pass one?", which stops being true once the first cell fills it in. Comparing against params.SHA1 fixes that.

dmick: yeah, it has to be in the executed shell env, but not necessarily in the global var 'env', I thought maybe.  and yeah I get how this fixes it; just concerned in general by multiple-parallel-writer things if they don't have to be.

but I don't know if you can easily set the child env for an sh command, so ... call it a thing to wonder about for the future.

dmick (approved):

PR #2688, opened 2026-08-20
sepia-fog-images: really claim testnodes, and pause the queue first

djgalloway:
[Build #6](https://jenkins.ceph.com/job/sepia-fog-images/6/console) failed in the ansible stage with `trial059` UNREACHABLE. The node was fine — the pipeline and a scheduled teuthology job were both driving it.

> [!NOTE]
> Needs `teuthology-lock --no-reimage`, merged in ceph/teuthology#2250. The job's prepare phase clones teuthology `main` on every run, so no `TEUTHOLOGYBRANCH` override is needed.

## What happened

| time (UTC) | event |
| --- | --- |
| 20:54:34 | teuthology job 499501 finishes on trial059, releases it |
| 20:54:5x | the lock phase picks trial059 out of `--status up --locked false` |
| **20:54:54** | **teuthology job 499559 locks trial059** — a real, atomic paddles lock |
| 20:54:5x | our `--update --status down --desc` lands *after*, clobbering 499559's description |
| 20:55:24 | FOG reports **3 active tasks** for the 2 hosts we queued 2 deploys for |
| 21:02 | trial059 comes up as ubuntu 22.04; `funCheckHostOs` passes, nothing looks wrong |
| 21:03:15–17 | 1.9s window, 61 consecutive `authorized_keys` items fail with `mkdir /home/ubuntu/.ansible/tmp ... exited with result 1`; items resume `ok` right after, but ansible still marks the host UNREACHABLE → rc 4 → `set -e` kills the phase |
| 21:05:39 | cleanup unpauses and releases **only trial075** |
| 21:06:32 | job 499581 locks trial075, 53s later |

## Why marking a node down never claimed anything

`lock_many` filters `Node.up.is_(True) AND Node.locked.is_(False)`, so a node genuinely marked down can't be selected — which fixes the ordering above: 499559's lock landed first, ours second.

And nothing stopped us. `teuthology-lock --update` → `PUT /nodes/<name>/` → `index_put()` → `node.update(...)`, with no ownership check anywhere on that path. `Node._check_for_update()` bails out immediately unless the payload carries a `locked` key:

```python
locking = values.get('locked')
if locking not in (True, False):
    return
```

`--status down --desc` sends only `up` and `description`. So paddles overwrote a node another job owned and returned success — the claim operation had no failure mode, and losing the race was silent by construction. Any reimage then sets the node back `up`, which is why cleanup released only trial075 and left trial059 wearing this build's lock description while a teuthology job owned it. It still reads that way in paddles.

## Really claim them

`funClaim` now takes a real lock and *then* marks the node down:

```bash
teuthology-lock --lock --no-reimage --desc "$lockdesc" "$1" || return 1
teuthology-lock --update --status down --desc "$lockdesc" "$1"
```

`--lock` alone was unusable here — for a FOG-backed machine type `lock/cli.py` reimages the node on the way in, the opposite of what the deploy phase wants — hence the `--no-reimage` dependency above. A lock payload carries a `locked` key, so `_check_for_update` actually runs and a contended claim fails loudly. Locking before touching the description also means we can no longer scribble on another job's node.

- `funRelease` undoes both via `teuthology-lock --unlock` (`-f` so one failed host doesn't strand the rest). On FOG-type nodes that powers the node off on its way out, which is fine — the queue's next reimage power-cycles it anyway.
- `funClaimedHosts` finds this build's hosts by lock owner + description instead of by status, so a foreign reimage can't make the pipeline forget a host it's holding. `--brief` with neither `-a` nor `--owner` already filters to the invoking user, so there's no owner plumbing.
- The claim loop walks the whole free list (`funFreeHosts`) instead of retrying the head of it, since losing a race is now a real outcome.

## Pause the queue first

Moved from the capture phase to the lock phase. The lock is atomic, but a live dispatcher still competes for every node the queue frees, and lock/deploy/ansible were all running against one. Paused, nodes only ever come free towards the job. The capture phase re-arms the same 2h expiry, and the four inline pause loops collapse into `funPauseQueue`.

## Testing

Every `teuthology-lock` invocation was checked against the `--no-reimage` branch's own `parse_args` and the assertions `cli.main()` runs — including that `--brief` with no `-a`/`--owner` defaults to the invoking user, and that `--update --no-reimage` is correctly rejected.

Phase logic was exercised against a stubbed lock server, running the real phases:

- lock → queue paused first, two nodes locked as `jenkins-build@soko04...` and marked down, loop exits at `numlocked=2`
- **the build #6 case** → flipped a held node back to `up` behind the script's back; unlock still found and released both
- lost lock race (`Cannot lock an already-locked node`) → falls through to the next free node instead of spinning
- another build's node (down, different description) → correctly ignored
- `DEFINEDHOSTS` set → pauses, skips locking, leaves the lock server untouched
- `PAUSEQUEUE=false` → no queue calls, and no misleading "LEAVING the queue paused" warning in cleanup

The stub can't prove the paddles round-trip, so a live build off this branch — with `TEUTHOLOGYBRANCH` pointed at the teuthology branch — is still the real check.

---

# Also in this PR

## `sepia-fog-images: say why a major-tracking seed found nothing`

Seeding `trial_rocky_10` from the newest captured point release already worked (`funNewestMinorImage`), and `rocky_upgrade_scope=major` already makes prep-fog-capture upgrade the seed against the major tree before capture. Only the diagnostics were missing: when no point release exists either, the job exited 1 saying `No captured FOG image named trial_rocky_10 exists`, which reads as if nothing was tried. Now it names the seed and the upgrade when one is found, and names the empty search when one isn't. No behaviour change.

## `sepia-fog-images: only drain deploys for the types we are capturing`

`phase_capture` waited on `.count` from `/task/active` — every active deploy on the FOG server. soko03 serves the whole lab, so a busy trial queue blocked a smithi capture, and with a 1hr retry budget could fail the job over deploys that can't touch the images being rewritten.

Now it counts only tasks whose host belongs to `$pausetypes`, via `funTypeHostIds` + `funCountTasksFor`. Host names are matched anchored, `^<type>[0-9]+$`, so `trial` doesn't also drag in `trial-perf`. An empty ID list is an error, not "nothing to wait for" — `phase_deploy` already resolved a host for every one of these types.

Against synthetic `/host` and `/task/active` payloads with six active deploys: capturing `smithi` waits on 1 instead of 6; `trial` matches trial059/trial075 but not trial-perf004 or trialrunner; `trial trial-perf` picks up all three.

## `sepia-fog-images: IMAGETYPE names the source image, not the capture`

The old semantics (`b8014107`/`669e54e5`) had `IMAGETYPE` rename the *capture*: `MACHINETYPES=trial IMAGETYPE=trial-perf` built trial-perf images on trial hardware, and verify booted them on a trial node too — the image for one machine type never touched that machine type.

Flipped: the job always captures `<MACHINETYPES>_<distro>` from nodes of that machine type (hardware-matched capture and verify), and `IMAGETYPE` names a different type's existing image to deploy as the **starting point**. `MACHINETYPES=trial-perf IMAGETYPE=trial` locks a trial-perf node, deploys the existing `trial_X`, runs ansible + prep-fog-capture, and captures the result as `trial-perf_X` (created in FOG on first capture). The source image is only ever read, so reruns re-derive from the same baseline and nothing can overwrite the source from the wrong hardware. `pausetypes` no longer includes the source type — its queue keeps running. The rocky major-only seeding follows the source type. Parameter descriptions on the job rewritten to spell all of this out.

### DIFF
--- a/sepia-fog-images/build/fog-images.sh
+++ b/sepia-fog-images/build/fog-images.sh
@@ -3,7 +3,8 @@
 # this script once per stage with a phase argument:
 #
 #   fog-images.sh prepare   Clone/bootstrap teuthology and ceph-cm-ansible
-#   fog-images.sh lock      Lock testnodes (skipped when DEFINEDHOSTS is set)
+#   fog-images.sh lock      Pause the teuthology queue and lock testnodes
+#                           (locking skipped when DEFINEDHOSTS is set)
 #   fog-images.sh deploy    FOG-deploy the existing image for each distro and
 #                           wait for the network sentinel file
 #   fog-images.sh ansible   Run cephlab.yml and prep-fog-capture.yml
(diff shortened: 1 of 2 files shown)


dmick:
eeeeehhhhhh....

I think probably teuthology-lock ought to grow a "--no-reimage" added option, or maybe a "--lock-only" option, either one to avoid the reimage step.  maybe hidden option.  but lock should have all the lock-related algorithms inside it; glomming on from outside is a recipe for desynchronization.

djgalloway:
> eeeeehhhhhh....
> 
> I think probably teuthology-lock ought to grow a "--no-reimage" added option, or maybe a "--lock-only" option, either one to avoid the reimage step. maybe hidden option. but lock should have all the lock-related algorithms inside it; glomming on from outside is a recipe for desynchronization.

Fixed with https://github.com/ceph/teuthology/pull/2250

dmick (approved):

PR #2690, opened 2026-08-27
ceph-pr-pipeline-trigger: readable build descriptions; only abort superseded builds

djgalloway:
Three follow-ups to the ceph-pr-pipeline trigger from watching today's first day of traffic:

**Say who did what, with links.** Trigger build descriptions now name the event and actor — `PR 71305 force push by anthonyeleven: triggering [signed submodules …]` — with the PR as a clickable link, instead of the raw webhook action and PR author. Force-pushes are distinguished from plain pushes by comparing the synchronize payload's `before`/`after` shas with the compare API (`diverged`/`behind` ⇒ force-push). The cause string gains the sender too.

**Link the pipeline run it schedules.** `build(waitForStart: true)` hands back the downstream run as soon as it exists (still without waiting for completion), so the description also links to the `ceph-pr-pipeline` build it started.

**Only abort builds the new run supersedes.** Cancellation matched on the PR number alone, so with one command allowed per comment, asking for two checks meant the second comment's build killed the first (`jenkins test make check arm64` on ceph/ceph#71356 aborted the still-running make-check build for the same sha). Both cancellation sites — the trigger's `cancelRunningPrBuilds` and the pipeline's `abortOlderBuilds` backstop — now abort only builds on an older head or with overlapping `CHECKS` on the same head. Pushes still cancel everything because the head changed.

Deployment notes:
- The JJB config adds `sender`/`push_before`/`push_after` webhook variables, so run `jenkins-jobs update` for `ceph-pr-pipeline-trigger` when this merges (before the next webhook, or descriptions read "push by null" until then).
- The new `ParametersAction` getters in the cancellation functions may want a one-time in-process script approval, same as the originals did.

Both Jenkinsfiles pass jenkins.ceph.com's declarative linter; the JJB definition renders with `jenkins-jobs test`.

### DIFF
--- a/ceph-pr-pipeline-trigger/build/Jenkinsfile
+++ b/ceph-pr-pipeline-trigger/build/Jenkinsfile
@@ -10,7 +10,9 @@
 //     developer re-adds the label.
 //   - "jenkins test <check>" comments re-run individual checks (org members
 //     always; others only while the PR carries the ci-approved label).
-//   - Any newly triggered run aborts still-running runs for the same PR.
+//   - A newly triggered run aborts still-running runs it supersedes: same
+//     PR on an older head, or overlapping checks on the same head.  Runs
+//     of disjoint checks on the same head are left alone.
 
 import groovy.transform.Field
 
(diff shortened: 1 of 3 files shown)


JoshuaGabriel:
thanks for spotting that, I was getting confused on why things were not finishing

falcocer-ibm (approved):

PR #2692, opened 2026-08-27
ceph-pr-pipeline: move the build-tree cache to the doli LRC's RGW

djgalloway:
## What

Moves the ceph-pr-pipeline build-tree cache (the zstd tarballs that hand the
compiled tree from the build leg to the make check / API test builders, and
serve as the re-run cache) from IBM COS to the RGW already running on the
Sepia LRC (the doli cluster), and makes the arm64 leg's cache upload
best-effort.

- `CACHE_ENDPOINTS` lists all six `doli0N.front.sepia.ceph.com:8080` RGWs;
  `s3_setup` picks the first one that answers, so a single doli host being
  down doesn't fail a leg.
- The `ibm-cloud-sccache-bucket` credential is replaced by
  `doli-lrc-pr-builds` (keys of the dedicated `ceph-pr-builds` RGW user).
  ceph-dev-pipeline is untouched and keeps using the IBM credential.
- The arm64 leg's upload no longer fails the leg: unlike x86_64 (where the
  tarball is the hand-off to other builders and must stay fatal), arm64
  builds and tests on the same node, so a cache outage shouldn't throw away
  a multi-hour build before the tests have even run.

## Why

Every builder the pipeline can land on lives in the Sepia lab, so the cache
traffic (a few huge, short-lived objects per PR) never needs to leave the
lab or ride on the IBM COS quota.

## Server-side setup (already done on the LRC)

- Dedicated `ceph-pr-builds` RGW user and bucket.
- Bucket-wide 21-day expiry lifecycle rule (same policy the COS bucket had)
  and a 20T user quota as a runaway safety valve.
- `doli-lrc-pr-builds` Jenkins credential created.

## Validation

- End-to-end from braggi01 with awscli (the client the legs use): upload,
  head-object (confirms objects get stamped with the `expire-21d` rule),
  download, byte-compare, delete.
- All six RGW endpoints answer by hostname from every builder pool the
  pipeline uses (braggi, soko, toko, bath, adami).
- Jenkinsfile passes the declarative linter.

Suggested post-merge check: one `FORCE_BUILD=true` run to exercise the new
cache through Jenkins' own credential, then the old COS `ceph-pr-builds`
bucket can be deleted.

### DIFF
--- a/ceph-pr-pipeline/README.md
+++ b/ceph-pr-pipeline/README.md
@@ -32,9 +32,9 @@ webhook (pull_request / issue_comment)
                  └─ ceph API tests    tree ← S3, run-backend-api-tests (in container)
 ```
 
-- **Cache:** build trees are zstd-tarred to S3 as
-  `pr-builds/<PR>/<head sha>.<arch>.tar.zst`.  Re-runs for an unchanged head
-  sha skip compilation (`FORCE_BUILD=true` overrides).
+- **Cache:** build trees are zstd-tarred to the Sepia LRC's RGW (the doli
+  cluster) as `pr-builds/<PR>/<head sha>.<arch>.tar.zst`.  Re-runs for an
+  unchanged head sha skip compilation (`FORCE_BUILD=true` overrides).
 - **Statuses:** each leg posts its own context via the GitHub API.  Pending
   is posted from `prepare` before legs wait for executors; canceled/failed
   runs finalize any still-pending contexts.  Docs/container/gha-only PRs
@@ -59,7 +59,8 @@ webhook (pull_request / issue_comment)
 
 1. Credentials: `pr-check-trigger-token`, `github-status-check-token`
    (repo:status + issues write + org read), `github-readonly-token`,
-   `dgalloway-docker-hub`, `ibm-cloud-sccache-bucket`, `ceph_win_ci_private_key`.
+   `dgalloway-docker-hub`, `doli-lrc-pr-builds` (RGW keys for the
+   `ceph-pr-builds` user on the doli LRC), `ceph_win_ci_private_key`.
 2. Deploy both jobs with JJB, then pre-approve the sandbox signatures in
    Manage Jenkins → Script Console:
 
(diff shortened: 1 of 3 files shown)


djgalloway:
Measured before/after with two same-day runs moving the same-size artifact (9.2G zstd x86_64 build-tree tarball):

- **before:** [build 51](https://jenkins.ceph.com/job/ceph-pr-pipeline/51/) (ceph/ceph#71356, IBM COS `s3.us-south`)
- **after:** [build 59](https://jenkins.ceph.com/job/ceph-pr-pipeline/59/) (ceph/ceph#71394, doli RGW, run with `CEPH_BUILD_BRANCH=` this branch, `FORCE_BUILD=true`, `STATUS_PREFIX=pipeline-test/`)

| step (9.2G tarball, x86_64) | IBM COS (51) | doli RGW (59) |
|---|---|---|
| upload (`aws s3 cp`) | 58s (~160 MB/s) | 48s (~195 MB/s) |
| download + untar (test leg) | ~137s and ~199s (two legs) | ~98s |
| tar creation (unchanged) | 33–39s | 33s |

Notes on the numbers:

- COS upload throughput was also erratic across legs the same day: the arm64 upload of a 9.3G tarball took 2m22s (~67 MB/s) vs 58s for x86_64. The doli path stays inside the lab, so it shouldn't see that WAN variance. (No doli arm64 number yet — build 59 ran `make-check` only.)
- Download durations include untar on the destination builder; COS numbers are Blue Ocean step durations minus ~15s of awscli venv setup.
- Build 59 also validates the new `doli-lrc-pr-builds` Jenkins credential end-to-end: head-object, upload and download all went through it.

Storage estimate at steady state: ~9.2G per x86_64 entry, ~9.3G per arm64 entry (main-target PRs only), keyed by PR head sha with a server-side 21-day expiry. At roughly 30 x86_64 + 25 arm64 new entries/day that's ~500G/day, so ~10–11T steady state — about half the 20T quota set on the `ceph-pr-builds` RGW user (a hard stop against runaways, not an eviction policy; the lifecycle is the only reaper). Objects come back from head-object stamped `expire-21d`, so the rule is confirmed active.

batrick (approved):

PR #2694, opened 2026-08-27
ceph-windows: cross-compile in a podman container instead of a VM

djgalloway:
The windows CI cross-compiles Ceph inside a freshly provisioned libvirt Ubuntu VM on every run: download the cloud image, boot, cloud-init, rsync the tree in, build everything (including all the mingw deps) from scratch, rsync ceph.zip back out, destroy the VM. That was ~50 minutes of a recent successful run (ceph-windows-pull-requests #86724) before the cluster-side work even started.

`win32_build.sh` is a plain Linux mingw cross-compile, so this runs it in a podman container on the builder host instead:

- a static toolchain image (`scripts/ceph-windows/win32-build.containerfile`) with the apt packages from `win32_deps_build.sh`'s ubuntu branch
- `scripts/ceph-windows/win32_build_container` bind-mounts the workspace and produces the same `$WORKSPACE/ceph.zip` that `setup_libvirt_windows_vm`/`run_tests` consume
- the compiled mingw deps (mingw-llvm, boost, openssl, ...) persist on the builder, keyed by the sha of the branch's `win32_deps_build.sh` + `mingw_conf.sh` + the containerfile, so any deps change simply builds a new cache entry; `win32_build.sh` already skips the deps build when it finds the cache's completed marker (a flock guards the first build per key)
- ccache for the ceph tree itself, with fixed container mount paths so object paths stay stable across builds

The new `ceph-windows-container-build` job validates this on the same builder pool the windows test jobs use, without touching ceph-pr-pipeline or the existing windows jobs; it doubles as a one-off manual windows build job for a branch or PR.

Validation on braggi07 (ceph main):
- [build 2](https://jenkins.ceph.com/job/ceph-windows-container-build/2/) (fully cold): **17.7 min** total — 6 min deps build, ~20 s configure, ~7 min ninja (32 workers), 222M ceph.zip
- [build 3](https://jenkins.ceph.com/job/ceph-windows-container-build/3/) (warm caches): **5.5 min** total — deps build skipped, ninja ~25 s at an 89% ccache hit rate, identical ceph.zip

vs ~50 minutes for the VM-based build phase it replaces. Not yet validated: running the produced ceph.zip through the actual windows tests — that's the next step before switching ceph-pr-pipeline's windows leg over.

Assisted-by: Claude Fable 5 (claude-fable-5)

### DIFF
--- a/ceph-pr-pipeline/README.md
+++ b/ceph-pr-pipeline/README.md
@@ -12,7 +12,7 @@ rules don't change.
 | `Unmodified Submodules` | [ceph-pr-submodules](../ceph-pr-submodules/) | GitHub API only, no clone, seconds |
 | `make check` | [ceph-pull-requests](../ceph-pull-requests/) | shared build → ctest on its own builder |
 | `ceph API tests` | [ceph-pr-api](../ceph-pr-api/) | shared build → API tests on its own builder |
-| `ceph windows tests` | [ceph-windows-pull-requests](../ceph-windows-pull-requests/) | parallel leg, own (win32) build |
+| `ceph windows tests` | [ceph-windows-pull-requests](../ceph-windows-pull-requests/) | parallel leg, containerized builds + one Windows VM (see below) |
 | `make check (arm64)` | [ceph-pull-requests-arm64](../ceph-pull-requests-arm64/) | parallel leg, build+test on one arm64 node |
 
 ## Flow
@@ -24,7 +24,7 @@ webhook (pull_request / issue_comment)
             prepare                   resolve PR, classify files, post pending statuses
             ├─ Signed-off-by          GitHub API
             ├─ Unmodified Submodules  GitHub API
-            ├─ ceph windows tests     (libvirt) own clone + build + tests
+            ├─ ceph windows tests     (noble+libvirt) container builds + Windows VM
             ├─ build and test (arm64) (arm64)  restore-or-build + tests
             └─ build and test
                  build ceph           (huge)  ONE clone, bwc -e buildtests, tree → S3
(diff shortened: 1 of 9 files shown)


djgalloway:
Now integrated into ceph-pr-pipeline's windows leg and validated end to end: [pipeline run 89](https://jenkins.ceph.com/job/ceph-pr-pipeline/89/) ran the full leg for ceph PR 71382 (windows check only, `container-test/` status prefix) on soko13 with completely cold caches — **95 min total, SUCCESS**, with the container-built ceph.zip passing the real windows tests:

- containerized cross-compile: **8 min cold** (4 min deps + 4 min ninja; ~2 min once the node's caches are warm) — was ~50 min in the build VM
- vstart build in the cluster VM: 32 min (unchanged; next candidate — restore the shared S3 build instead)
- windows tests: 46 min; VM image downloads + boots: ~3 min

Old ceph-windows-pull-requests successes ran 114–177 min. Also in this branch: the windows leg now requires `installed-os-noble` builders like the other container legs (run 88 landed on jammy braggi12, where setup_container_runtime.sh only provides docker, and failed on `podman: command not found` — only braggi10-13 are jammy in the libvirt pool).

djgalloway:
vstart-in-container is now validated end to end as well: [pipeline run 99](https://jenkins.ceph.com/job/ceph-pr-pipeline/99/) ran the full windows leg on soko12 with **completely cold caches** — **79 min, SUCCESS, all tests passed** (ceph_test_dokan ~15 min, matching the VM-cluster baseline). Cold breakdown: ~12.5 min cross-compile (image + deps + build, fresh node), ~13 min windows VM + vstart build + cluster up, ~48 min tests, ~4 min artifacts. Warm nodes should land around 60-65 min.

Two vstart behaviors surfaced during validation (runs 92/97/98), for the record:

- vstart writes `mds root ino uid/gid = $(id -u/-g)` into the MDS config, and the windows client mounts with `client_mount_uid/gid = 1000` + `client_permissions = true` — so the cluster must be started by uid 1000 (as the VM's ubuntu user always did, silently). Started by root, every Dokan write gets EACCES.
- vstart writes outside VSTART_DEST: `logrotate.conf`/`logrotate.state` in the build dir and `vstart_environment.sh` next to itself in src. The vstart user now owns the build dir wholesale plus that one src file (the full write surface, per audit of vstart.sh).

The READY wait also fails fast now when the container exits, instead of sitting out its timeout.

Whole leg, measured: **114–177 min (old VM flow) → 79 min cold / ~60 min warm**, with only the Windows client VM remaining as a VM.

djgalloway:
The rework at a glance — old VM flow vs this branch vs the goal, on one time scale, plus how the containers, caches, and the Windows VM fit together:

Windows leg rework: timelines, architecture, and status

batrick (approved):

PR #2698, opened 2026-08-28
ceph-pr-pipeline: run make check (arm64) on all active release branches

djgalloway:
The arm64 leg was restricted to `main` while we validated the pipeline. Open it up to the same active branches as the other legs: squid, tentacle, umbrella, and main.

Jenkinsfile passes the declarative linter on jenkins.ceph.com.


djgalloway:
Overall arm64 builder utilization (busy time) is very low so we have capacity.

image

batrick (approved):

PR #2699, opened 2026-08-31
scripts/build_utils: classify scripts changes as QA only

batrick:


djgalloway:
I don't see a `scripts/` in ceph.git?

batrick:
> I don't see a `scripts/` in ceph.git?

ooops, src/script is what i meant.

djgalloway:
Needs to go here too https://github.com/ceph/ceph-build/blob/main/ceph-pr-pipeline/build/pr_checks.sh#L100

batrick:
> Needs to go here too https://github.com/ceph/ceph-build/blob/main/ceph-pr-pipeline/build/pr_checks.sh#L100

done, ty

djgalloway (approved):
For posterity, `run-make-check.sh` is in `src/script` but it'll still get run on x86 when changes are made to it.  Breakage on arm64 won't get detected because that check is skipped when QA_ONLY=true.

PR #2705, opened 2026-09-03
sepia-fog-images: STARTWITHMAAS installs the OS from MAAS as the starting point

djgalloway:
## What

Adds a `STARTWITHMAAS` checkbox (default **off**) to the sepia-fog-images pipeline. When ticked, the starting point for each distro is a **fresh MAAS install** instead of an existing FOG image:

1. The host's PXE entry on the dnsmasq server (soko01) is pointed at `maas` (`funSetPxe`, which now fails loudly if the host has no `dhcp-host` line at all).
2. MAAS (soko02) installs the distro (allocate + deploy; MAAS drives the power cycle). A machine MAAS has never used (state `New`) is commissioned automatically first — PXE already points at maas by then. Stale Deployed/Allocated/rescue records are cleared automatically.
3. A host that only exists in MAAS is **registered in FOG automatically** using the boot MAC MAAS reports (deploy and verify phases both). The FOG image record is created on first capture, as before.
4. Once MAAS reports Deployed, PXE goes back to `fog`; if the installed OS's default user isn't `ubuntu` (Rocky/CentOS cloud images), the `ubuntu` user is created from it.
5. The normal ansible → prep → fsck → capture → verify flow runs unchanged.

When unchecked, behavior is exactly as before: FOG-deploy an existing image, and **fail** if no usable one exists.

Distro → boot resource mapping: Ubuntu versions map to series codenames; everything else is matched against the resources MAAS actually has (verified against live soko02: `rocky10`/`rocky9` bare custom names deploy as `osystem=custom`, `centos/centos9-stream`, `rhel/*`). A pinned minor (`rocky_10.1`) only accepts an exact-version resource; only major-tracking distros (`rocky_10`, `centos_9.stream`) may fall back to a major-level one.

## Why

A distro with no captured FOG image was a dead end ("seed the first image manually"). The first target is the gibba fleet: all 45 hosts are FOG-registered on soko03 and MAAS-enrolled on soko02, but there are zero `gibba_*` images.

## Test prerequisites (infra, pending)

- maas CLI on soko04 (`snap install maas`)
- jenkins-build's pubkey in `~ubuntu/.ssh/authorized_keys` on soko01 (funSetPxe)
- jenkins-build's pubkey on the MAAS user the `maas-api-key` credential belongs to (so seeded hosts are ssh-able)

Then: `MACHINETYPES=gibba DISTROS=ubuntu_22.04 STARTWITHMAAS=true CEPH_BUILD_BRANCH=wip-sepia-fog-images-maas-seed`

### DIFF
--- a/sepia-fog-images/README.rst
+++ b/sepia-fog-images/README.rst
@@ -30,15 +30,17 @@ These steps should only have to be performed when a new teuthology host is being
 
 #. Define ``FOG_API_TOKEN`` and ``FOG_USER_TOKEN`` as **Global name/password pairs** in Jenkins.  These are only a fallback: when the agent's ``/etc/teuthology.yaml`` has a ``fog:`` block, the job reads the FOG endpoint and tokens from there (currently ``http://soko03.front.sepia.ceph.com/fog``).
 
-#. The ``maas-api-key`` **Secret text** credential must exist in Jenkins (shared with the builder-reimage job).  Only needed for ``FSCKMETHOD=maas-rescue``.
+#. The ``maas-api-key`` **Secret text** credential must exist in Jenkins (shared with the builder-reimage job).  Needed for ``FSCKMETHOD=maas-rescue`` and for ``MAASFALLBACK``.
+
+#. For ``MAASFALLBACK``, the ``jenkins-build`` user's public SSH key must be added to the MAAS user the API key belongs to (MAAS injects that user's keys into the deployed OS's default account, which is how the job gets onto a freshly-seeded host).
 
 **NOTE:** This job also relies on:
 
 - ceph-sepia-secrets_ -- If the job is being run on a teuthology host, ``/etc/ansible`` should already be symlinked to a ceph-sepia-secrets checkout.
 - ceph-cm-ansible/tools_ -- ``prep-fog-capture.yml`` preps a host for capturing.
 - The fsck postinit hook on the FOG server (``/images/dev/postinitscripts/fsck_before_capture.sh``, sourced by ``fog.postinit``).  The job refreshes it over ssh when it can, but installing it once by hand is enough for ``FSCKMETHOD=fog-postinit``.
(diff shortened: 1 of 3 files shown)


djgalloway:
Pushed four commits onto this branch while seeding `trial_ubuntu_26.04` with it (run #31 captured + verified):

- `8ef9e887` map `ubuntu_26.04` → `ubuntu resolute` (MAAS already had the boot resource; the hardcoded codename map stopped at noble)
- `ddd51ac7` re-commission machines stuck in `Failed commissioning` (bounded, 2 attempts) instead of bailing; widened the Ready wait for the extra PXE cycle
- `89b499d7` `funMaasEnsureNetwork`: only a *dhcp* link counts as configured — a fresh commission leaves an `auto` link and MAAS then deploys with a static IP of its own choosing (trial001 came up at .192.89 while DNS said .193.1 and looked unreachable)
- `b702b9d9` `funMaasEnsureBootDisk`: with no rotary disk, boot from the smallest physical disk (trial = 2×1920GB payload NVMe + 960GB boot NVMe); MAAS's default put the OS on a payload disk and the testnode disk-matching then found only one free ~1700GB disk

Companion fixes on the cm-ansible side are in ceph/ceph-cm-ansible#881 (sudo-rs, ntpsec, timezone, package renames, netplan wait, and keeping the firmware EFI shell last in BootOrder — the 'reboot hang' root cause).

djgalloway:
Added 0821b855: `phase_prepare` now also writes `StrictHostKeyChecking no` + `UserKnownHostsFile /dev/null` for `*.front.sepia.ceph.com` into the agent's `~/.ssh/config` (same as the other lab accounts). Without it, the ansible runs from this job record each capture node's host key in jenkins-build's known_hosts, and the next teuthology reimage of that node from that account hangs for hours on `BadHostKeyException` (bit us on trial002/trial065/trial150 today). Teuthology-side fix so a stale entry can never hang a reimage anyway: ceph/teuthology#2255.

PR #2709, opened 2026-09-09
ceph-pull-requests-ppc64le: make check for PRs on ppc64le, on request

djgalloway:
https://tracker.ceph.com/issues/80185: let developers run make check for a PR on ppc64le by commenting `jenkins test make check ppc64le`, alongside the existing weekly `ceph-make-check-periodic-ppc64le` job.

- New freestyle job modeled on `ceph-pull-requests-arm64`, with the weekly ppc64le job's build configuration (centos9 container, no dashboard, no breakpad/werror/fio).  No S3 build cache: the ppc64le builders can't reach the doli cluster, so every run builds fresh, like the weekly job.
- Triggered by the generic webhook trigger on the same ceph/ceph webhook that feeds `ceph-pr-pipeline-trigger` (matched via the shared `pr-check-trigger-token` credential), filtered to new PR comments containing the phrase.  Since nothing sits in front of the job, the build script enforces the same policy as the pipeline trigger: the commenter needs write/admin on ceph/ceph, or the PR must carry the `ci-approved` label.
- Statuses are posted as `make check (ppc64le)` by the job itself for every build cause, via a new `FORCE_GITHUB_PR_STATUS` override of the manual-triggers-only guard in `update_github_pr_status`.
- `ceph-pr-pipeline-trigger` gets a wider negative lookahead so the ppc64le phrase doesn't also queue the x86_64 make-check leg.

JJB definition renders clean with `jenkins-jobs test`; the trigger Jenkinsfile passes the declarative linter.

### DIFF
--- a/ceph-pr-pipeline-trigger/build/Jenkinsfile
+++ b/ceph-pr-pipeline-trigger/build/Jenkinsfile
@@ -64,9 +64,10 @@ def phraseToChecks(String body) {
   def hits = []
   if (body =~ /(?i)jenkins\s+test\s+signed/)      { hits << "signed" }
   if (body =~ /(?i)jenkins\s+test\s+submodules/)  { hits << "submodules" }
-  // negative lookahead so "make check arm64" doesn't also queue the
-  // x86_64 leg; both can still be named in one comment
-  if (body =~ /(?i)jenkins\s+test\s+make\s+check(?!\s+arm64)/) { hits << "make-check" }
+  // negative lookahead so "make check arm64" (and "make check ppc64le",
+  // which triggers the standalone ceph-pull-requests-ppc64le job) doesn't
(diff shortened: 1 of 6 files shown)


djgalloway:
Why a standalone job instead of a comment-only leg in ceph-pr-pipeline: this keeps the on-request ppc64le check entirely off the production trigger/pipeline path while the scarce ppc64le builders and the phrase trigger prove themselves — if anything here is flaky, nothing else is at risk. The leg also couldn't share any of the pipeline's build machinery anyway (the ppc64le builders can't reach the doli S3 cache, so every run builds fresh like the weekly job). If it sees real use, it can fold into the pipeline as a comment-only leg later, which would replace this job's own auth check and status posting with the trigger's.

batrick (approved):

PR #2713, opened 2026-09-10
ceph-pr-pipeline: retry GitHub API calls, and never let a status POST decide a leg's outcome

djgalloway:
Two commits; together they replace #2710.

### 1. Retry the GitHub API calls instead of dying on one bad response

2026-09-10 13:42 UTC, builds [1126](https://jenkins.ceph.com/job/ceph-pr-pipeline/1126/) and [1127](https://jenkins.ceph.com/job/ceph-pr-pipeline/1127/) (both ceph/ceph#71659) died 25 seconds in, in the same second, on the first line of `doPrepare`:

```
+ curl -sf -u **** -H Accept: application/vnd.github+json https://api.github.com/repos/ceph/ceph/pulls/71659
ERROR: script returned exit code 22
```

Builds started just before (1125) and just after (1128) went through fine — a momentary GitHub hiccup, and `curl -sf` threw away the status code.

Both Jenkinsfile call sites and `pr_checks.sh`'s `gh_api` now go through a helper that retries 429 / 5xx / connection failures / the **rate-limit flavour of 403** with a 5s, 10s, 15s, 20s backoff. 403 is also what a revoked token gets and that never clears on its own, so it's only retried when GitHub's body says "rate limit". Anything else fails immediately, with the URL, status code and response body in the log.

`curl --retry` alone wouldn't cover this: it ignores 403, and `--retry-all-errors` (what #2710 used) is too new for the curl on some builders.

### 2. A status POST never fails a leg, and the sweep always runs

Three holes in the status lifecycle, all seen this week:

- **`postStatus()` threw on a failed POST**, so a status update could change a leg's outcome. In the 2026-09-09 brownout that turned dozens of green legs red — [985](https://jenkins.ceph.com/job/ceph-pr-pipeline/985/)'s 78-minute arm64 make check passed, then died posting `-> success`. Now best-effort: warn and carry on; `leg_finalized` only flips once a terminal POST actually lands, so the `post{}` sweep can repair what slipped. *(from #2710)*
- **No `success{}` branch in `post{}`**, so a green build whose `-> success` was lost left it pending forever. *(from #2710)*
- **`finalizePendingStatuses()` bailed when `HEAD_SHA` was unset** — but the trigger has already set every `CHECKS` context to "queued" on the sha it passes in, so a build dying before prepare resolves the PR (1126/1127, and the "superseded" path) leaves them all stuck pending: no failure to look at, PR silently blocked. That's what 71659 looked like. Now falls back to `params.HEAD_SHA` and sweeps the `CHECKS` tokens. A manual run without `HEAD_SHA` has nothing trigger-posted to clean up, so it still returns.

### Testing

- Declarative linter: **Jenkinsfile successfully validated** (both commits); `bash -n` clean on `pr_checks.sh`.
- Helper exercised standalone: live 200 (body on stdout, rc=0); live 404 (fails fast, prints GitHub's message, rc=22); local server 503×2→200 (two retries, 15s, rc=0); dead port (retries as "connection failure" — curl writes `000` for `%{http_code}` on a refused connection, so the non-zero exit is what's checked); local 403 with "secondary rate limit" body ×2→200 (retried, rc=0); local 403 "Bad credentials" (fails fast in 0s, rc=22).

Found while running `make check` on Ubuntu 26.04 via `DISTRO_BASE=resolute`; unrelated to that work otherwise.

### DIFF
--- a/ceph-pr-pipeline/build/Jenkinsfile
+++ b/ceph-pr-pipeline/build/Jenkinsfile
@@ -59,6 +59,8 @@ import groovy.transform.Field
 @Field boolean run_api = false
 @Field boolean run_windows = false
 @Field boolean build_cached = false
+// true once prepare has resolved the PR and decided which legs run
+@Field boolean prepared = false
 // context key -> true once a terminal (non-pending) status was posted, so the
 // abort/failure sweep in post{} only finalizes contexts still left pending.
 @Field Map leg_finalized = [:]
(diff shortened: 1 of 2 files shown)


guits (approved):
it matches what killed builds 1126/1127, i think the helper is doing the right things, looks good to me

PR #2714, opened 2026-09-10
ceph-pr-pipeline: merge PRs locally and key the cache by head+base

djgalloway:
Reworked per @batrick's review: no more reliance on GitHub's test merge, in either the checkout or the cache key.

**Before:** legs checked out `refs/pull/N/merge` (refreshed lazily by GitHub — it can silently point at the PR merged onto a stale base), and the S3 build cache was keyed by the PR head sha alone (so a PR that sat still while its base moved kept restoring a tree built against old main — [run 1137](https://jenkins.ceph.com/job/ceph-pr-pipeline/1137/) and [run 1152](https://jenkins.ceph.com/job/ceph-pr-pipeline/1152/) died with `FindBoost "0.0.0"` when such trees could no longer re-run cmake in today's build container).

**Now:**
- `prepare` pins two shas once: the PR head (as before) and the **current target-branch head**.
- Every leg checks out the PR head and **merges it locally** onto exactly that pinned base sha. Merge conflicts fail cleanly (and the existing `mergeable == false` fast-path in prepare still reports them plainly before any leg runs). All legs of a run build one identical tree even if the base moves mid-run — which is actually stronger than the old behavior, where GitHub could move the merge ref between legs.
- The cache key is exactly the merge's two inputs: `pr-builds/<PR>/<head sha>-<base sha>.<arch>.tar.zst`. It invalidates itself when either side moves; old head-keyed objects age out via the bucket's 21-day expiry.

The local merge needs real ancestry, so the shallow clone is replaced by a full-history clone backed by a **per-builder heads-only mirror** (auto-created on first use, `git fetch`-refreshed each run under flock, `gc.auto=0` so reference clones never lose borrowed objects). The mirror is purely an optimization — the clone fetches anything it lacks straight from GitHub, so a missing/stale mirror costs minutes, never correctness. This also retires the hand-seeded `CEPH_REFERENCE_REPO` TODO: the param now just overrides the mirror location.

### DIFF
--- a/ceph-pr-pipeline/build/Jenkinsfile
+++ b/ceph-pr-pipeline/build/Jenkinsfile
@@ -266,24 +266,46 @@ def checkoutCephBuild() {
   """
 }
 
-// One shallow checkout of the PR merge ref into ./ceph.  Point
-// CEPH_REFERENCE_REPO at a local mirror to cut the 4-5 minute clone down to
-// seconds (the git plugin silently skips a reference path that doesn't exist).
+// One checkout of the PR merged onto its target branch, with the merge done
+// HERE rather than by GitHub: refs/pull/N/merge (and its merge_commit_sha)
+// is only refreshed when something pokes the PR's mergeability, so checking
+// it out can silently test a stale base.  prepare pinned HEAD_SHA and
+// BASE_SHA once, every leg merges those same two shas, and the S3 cache is
+// keyed by exactly those inputs -- nothing depends on GitHub's test merge.
+//
+// The full history needed for the merge base comes cheaply from a
+// per-builder mirror (heads only: all of refs/pull/* would be immense),
+// auto-created on first use and refreshed per run.  The mirror is purely an
+// optimization -- the clone below fetches whatever it lacks straight from
+// GitHub, so a missing or stale mirror costs minutes, never correctness.
+// CEPH_REFERENCE_REPO overrides its location.
 def checkoutCephPr() {
-  sh "rm -rf ceph"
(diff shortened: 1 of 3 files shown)


ceph-pr-pipeline/build/Jenkinsfile
```diff
@@ -314,6 +314,18 @@ def doPrepare() {
...
+  // branch head -- the very commit refs/pull/N/merge points at, which is
+  // what checkoutCephPr() checks out and every leg builds.  The S3 cache
+  // must be keyed by it, not by the head sha: the head doesn't change when
+  // the target branch moves, but the built tree does, and a tree cached
+  // against an old target can no longer even re-run cmake in today's build
+  // container (runs 1137/1152: FindBoost "0.0.0" because /opt/ceph boost
+  // changed under a stale cached tree).  The field is populated here -- the
+  // fetch loop above waited for mergeability, and conflicted PRs (which
+  // have no test merge at all) error out below.
+  env.CACHE_SHA = pr_data.merge_commit_sha ?: env.HEAD_SHA
```
batrick: Oh. Interesting. This is actually not good. Github is famously bad at updating PR merge commits after the base branch advances. We should probably not be relying on that at all. Can we not just have jenkins do a trivial merge itself?

batrick (dismissed):

ceph-pr-pipeline/build/Jenkinsfile
djgalloway: Yes, implemented

batrick (approved):

PR #2719, opened 2026-09-11
builder.yml: install vagrant based on the libvirt Jenkins label

djgalloway:
Builders are selected by the `libvirt` Jenkins label, but builder.yml only installed vagrant when `-e libvirt=true` was passed. Hosts registered advertising `libvirt` with no vagrant, and `cephadm-ansible-prs` jobs landing on them died with `vagrant: command not found`.

- Derive `libvirt` from the host's Jenkins label instead of a CLI flag.
- Assert when an override would contradict the label.
- Verify vagrant runs as the jenkins user before the play registers the node, so a host that can't run vagrant never comes online claiming it can.
- Let an inventory supply `secrets_path` via `default_secrets_path` (play vars outrank inventory vars, hence the indirection). Pairs with ceph/ceph-sepia-secrets#1236.
- Fail fast in the cephadm-ansible-prs functional scenarios when vagrant is missing.

Ran against the three builders this found broken (braggi01, adami06, adami07); all 35 `libvirt`-labelled hosts now pass.

### DIFF
--- a/ansible/examples/builder.yml
+++ b/ansible/examples/builder.yml
@@ -1,12 +1,20 @@
 ---
 ## Instead of trying to keep 4 separate playbooks up to date, let's do it all here.
-## The only difference from using multiple playbooks is we need to specify `-e libvirt=true` and/or `-e permanent=true` if the builder will be permanent/static.
+## The only difference from using multiple playbooks is we need to specify `-e permanent=true` for a
+## permanent/static builder outside Sepia.  Sepia builders are detected by IP and set permanent themselves.
+## Vagrant/libvirt installation is driven by the host's `jenkins_labels` entry: a label containing `libvirt` installs them.  Override with `-e libvirt=true|false`.
 ## Tested on: CentOS 7, CentOS 8, Xenial, Bionic, Focal, Leap 15.1 using ansible 2.8.5
 ##
 ## Example:
 ## define labels in inventory "jenkins_labels" dict, keyed by fqdn
 ##
-## ansible-playbook -vvv -M ./library/ builder.yml, -e '{"api_uri": "https://jenkins.ceph.com"}' --limit braggi01*
+## From a ceph-sepia-secrets checkout (its ansible.cfg supplies the inventory and vault password,
+## and its group_vars supply default_secrets_path), a Sepia builder needs no extra vars at all:
+##
+##   ansible-playbook -M ~/src/ceph-build/ansible/library/ ~/src/ceph-build/ansible/examples/builder.yml --limit braggi01*
+##
+## Elsewhere, pass what the inventory does not provide, e.g.
+##   ansible-playbook -vvv -M ./library/ builder.yml -e '{"api_uri": "https://jenkins.ceph.com"}' --limit braggi01*
 #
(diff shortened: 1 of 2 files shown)


djgalloway:
Pushed 7803e86 to close the gap rather than document it.

Deriving `libvirt` from the label fixed the default case, but two paths could still register a builder advertising `libvirt` with no vagrant on it:

1. **`-e libvirt=false` on a libvirt-labelled host** — extra-vars outrank the derived default, so the override could still produce a lying label. Now an `assert` at the top of the play refuses to proceed.
2. **vagrant or the plugin silently not landing** — the `jenkins_node` registration at the end of the play didn't care whether any of the install tasks had achieved anything. Now the play verifies `vagrant` and `vagrant plugin list | grep vagrant-libvirt` as the jenkins user, inside the existing `libvirt|bool` block.

Because registration is the last thing the play does, a host that can't run vagrant now fails out before it ever comes online — so reaching the Jenkins node list at all means vagrant ran there. That's what makes the label trustworthy.

The verification runs `vagrant plugin list` with no special environment on purpose: `build_utils.sh` invokes vagrant from Jenkins the same way, so checking it under the install-time `GEM_HOME`/`CONFIGURE_ARGS` would test something the jobs never do.

The guard in `cephadm-ansible-prs/build/build` stays, but its comment no longer claims the invariant is broken — it's now only a backstop for a label hand-edited in the Jenkins UI or vagrant removed from a builder after provisioning.

Unchanged from before: this is all provision-time enforcement, so braggi01 and adami07 still need a `builder.yml` run, and the rest of the `libvirt`-labelled pool is still unaudited.

djgalloway:
Pushed bf1fd6a, which pairs with ceph/ceph-sepia-secrets#1236 to let an inventory supply what currently has to go on the CLI.

The interesting constraint: **play vars outrank inventory `group_vars`**, so `secrets_path` cannot just be overridden from the sepia inventory — it would be silently ignored. Verified rather than assumed:

```
group_vars: permanent: true   +   play vars: permanent: false   ->   "permanent=False"
```

So `secrets_path` gains a `default_secrets_path` indirection instead:

```yaml
secrets_path: "{{ lookup('env', 'ANSIBLE_SECRETS_PATH') | default(default_secrets_path | default('/etc/ansible/secrets'), true) }}"
```

`ANSIBLE_SECRETS_PATH` still wins, and the `/etc/ansible/secrets` fallback keeps the ephemeral mita/prado builders — which run this playbook locally as root with no inventory — working unchanged.

Also in this commit:

- `examples/library` symlinked at `../library`, so Ansible finds the `jenkins_node` module by itself and `-M ./library/` is no longer needed.
- Header comment corrected: `-e permanent=true` is only needed **outside** Sepia. The Sepia-IP `set_fact` at line 458 already forces it for anything on `172.21.`/`8.43.`/`10.20.`, and `set_fact` does outrank play vars.

Together with the secrets-repo side, a Sepia builder reprovisions with:

```
ansible-playbook ~/src/ceph-build/ansible/examples/builder.yml --limit braggi01*
```

guits (approved):

PR #2729, opened 2026-09-22
ceph-dev-pipeline: require a sepia builder for Pulp uploads

djgalloway:
## What

In the `parallel build` matrix agent label, add `&&sepia` when `PULP_UPLOAD` or `PULP_REGISTRY_UPLOAD` is set.

## Why

cve-pipeline [#38](https://jenkins.ceph.com/job/cve-pipeline/38/) and #36 failed only in the centos9/arm64 cell. The failing step was "configure Pulp":

```
Error: HTTPSConnectionPool(host='pulp.front.sepia.ceph.com', port=443): ... [Errno 110] Connection timed out
```

In both builds that cell ran on `9.11.121.69+ceph129`. `pulp.front.sepia.ceph.com` is a CNAME to the pok OpenShift route at private address 172.16.53.7, and ceph129 (on IBM's 9.x network) can't reach it. ceph129 is currently the only agent labeled `gigantic` without `sepia`, and all 43 other `gigantic` agents have both labels.

## Notes for reviewers

- The requirement only applies when Pulp is involved. Both Pulp params default to false in ceph-dev-pipeline, so ceph129 still takes chacra-only builds.
- The alternative is to remove the `gigantic` label from ceph129.
- The declarative linter (`/pipeline-model-converter/validate`) accepts the change. It hasn't run in a real build yet. A cve-pipeline rebuild after merge will test it.


dmick (approved):
Was not even aware we had a 'sepia' label.  Cool.

